PR #14, opened 2023-04-25
add tqdm progress bar

maurerle:
- this does not have a big impact on performance when refresh=False
- fixes is not None check for DataFrame


nick-harder (approved):
nice!

PR #24, opened 2023-04-26
make db connection optional

maurerle:
- fix test case
- one should be able to create world with the default args (without a given database)


maurerle:
Now the checks are running through and linting works too :)

nick-harder (approved):

PR #26, opened 2023-05-02
Added EOM bidding strategy from flexABLE

nick-harder:
-added strategy for bidding on EOM from flexABLE
-see Example_01_b and Example_02
-new strategy in strategies folder


nick-harder:
ran black and isort

maurerle (approved):

PR #31, opened 2023-05-10
Feature/control reserve market

nick-harder:
-preparing to add CRM
-several smalled changes and fixes
-black and isort performed
-examples work

### DIFF
--- a/assume/strategies/flexable.py
+++ b/assume/strategies/flexable.py
@@ -20,7 +20,7 @@ class flexableEOM(BaseStrategy):
         bid_quantity_flex, bid_price_flex = 0, 0
 
         if operational_window is not None:
-            self.current_time = operational_window['window']['start']
+            self.current_time = operational_window["window"]["start"]
             # =============================================================================
             # Powerplant is either on, or is able to turn on
             # Calculating possible bid amount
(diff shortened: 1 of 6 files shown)


assume/strategies/flexable.py
```diff
@@ -19,6 +20,7 @@ def calculate_bids(
         bid_quantity_flex, bid_price_flex = 0, 0
 
         if operational_window is not None:
+            self.current_time = operational_window["window"]["start"]
```
maurerle, line 23: This is wrong - the start of the bidding window is not the current time

assume/units/powerplant.py
```diff
@@ -168,57 +106,64 @@ def calculate_operational_window(
...
+
         operational_window = {
+            "window": {"start": start, "end": end},
             "current_power": {
                 "power": current_power,
                 "marginal_cost": self.calc_marginal_cost(
                     power_output=current_power,
-                    current_time=start,
-                    partial_load_eff=True,
+                    timestep=start,
```
maurerle, line 152: what is the timestep for - if we have window containing start + end above?

maurerle (commented):
Having the current_time in the strategies seems weird to me.
Especially as it is not the current simulation time.

The timestep in the operational_window seems to be unneeded
I don't think it makes sense to delay this PR because of those two reasons

assume/strategies/flexable.py
nick-harder: yes, you are correct. This is still partially work in progress, this will change to better naming.

PR #36, opened 2023-05-12
revert del statement, because leads to db locked error

kim-mskw:
Undo del line in outputs for main, which I accidentially merged


maurerle:
I would rather close this in favor of the other PR which handles the opened database session correctly through an access handler.
I hope that's fine ;)

PR #41, opened 2023-05-16
Expand grafana dashboard

kim-mskw:
Just cleaned the dashboard a bit and made minor changes to the output role as well as added proper documentation

### DIFF
--- a/assume/common/units_operator.py
+++ b/assume/common/units_operator.py
@@ -125,7 +125,7 @@ class UnitsOperator(Role):
         # await self.send_dispatch_plan()
 
     def send_dispatch_plan(self):
-        # todo group by unit_id
+        #TODO group by unit_id
         current_time = pd.to_datetime(self.context.current_timestamp, unit="s")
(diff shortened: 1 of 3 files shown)


assume/common/units_operator.py
```diff
@@ -135,6 +135,8 @@ def send_dispatch_plan(self):
             dispatch_plan = {"total_capacity": total_capacity}
             self.units[unit_id].get_dispatch_plan(dispatch_plan, current_time)
 
+            print(dispatch_plan)
+
```
maurerle, line 139: This should be logger.debug("Got Dispatch Plan: %s", dispatch_plan)

maurerle, line 139: This should be 

```suggestion
            logger.debug("Got Dispatch Plan: %s", dispatch_plan)
```

maurerle:
@kim-mskw you need to run black so that the linter is happy ;)

maurerle (approved):
did not imagine how the json looks like but everything else looks good :D

PR #43, opened 2023-05-16
Heatpump and electrolyser

Manish-Khanra:


maurerle:
Can you run black and remove .idea and output/local_db files?

You can add the unit type to self.unit_types in world too :blush:

PR #51, opened 2023-05-22
dispatch is now written exactly when it is time

### LINKED ISSUE #49: Units info in DB
The units info is written again and again during each run ov the same scenario, which can be seen in the total capacity plot, which gets larger with each iteration. please fix

maurerle:
- dispatch was written after each market feedback, this does not work with multiple markets
- fix for deletion of previous data using a DBMS with transactions
- calculate the cumulated timeseries from given orderbook
- dispatch table contains market_id and bid_id
- closes #49

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,6 +130,7 @@ dmypy.json
 
 .DS_Store
 *.orig
+.idea
 .vscode
 examples/outputs
 examples/local_db/
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -65,15 +65,17 @@ class WriteOutput(Role):
         # Get list of table names in database
         if self.db is not None:
             table_names = inspect(self.db.bind).get_table_names()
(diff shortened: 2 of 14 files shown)


kim-mskw (commented):
KPI capacity factor throws error:
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation "unit_dispatch" does not exist
LINE 1: ..._id, avg(power/max_power) as capacity_factor from unit_dispa...
                                                             ^

[SQL: select bid_id as name, market_id, avg(power/max_power) as capacity_factor from unit_dispatch ud join unit_meta um on ud.bid_id = um."index" and ud.simulation=um.simulation where um.simulation = 'example_01' group by name, market_id]
(Background on this error at: https://sqlalche.me/e/20/f405)

kim-mskw (approved):

PR #57, opened 2023-05-24
Crm bidding strategies

nick-harder:


assume/common/utils.py
```diff
@@ -291,6 +289,8 @@ def aggregate_step_amount(orderbook: Orderbook, begin=None, end=None, groupby=[]
 
     If called without groupby, this returns the aggregated orderbook timeseries
     """
+    if groupby is None:
+        groupby = []
```
maurerle, line 293: What is the benefit of this?

assume/common/utils.py
```diff
@@ -316,11 +320,16 @@ def aggregate_step_amount(orderbook: Orderbook, begin=None, end=None, groupby=[]
...
-                    (date + timedelta(hours=duration_hours), -bid["volume"]) + add
+                deltas.extend(
+                    (
+                        (date, bid["volume"]) + add,
+                        (
+                            date + timedelta(hours=duration_hours),
+                            -bid["volume"],
+                        )
+                        + add,
+                    )
```
maurerle, line 331: if you want to reduce appends you should also list comprehend the for loop `for date in starts:`.

Well I don't know - thats probably not our main bottleneck is it?

assume/common/utils.py
nick-harder: before the groupby was  mutable argument, which is not desired. so I replaced it but this is the work around

assume/common/utils.py
nick-harder: yeah, could be helpful as well. Maybe not a bottleneck, but some optimization and nices looking code is still nice.

PR #59, opened 2023-05-25
Crm bidding strategies

nick-harder:
-added differentiation of product type in demand units
-added the option to set bidding strategies to none
-bidding is off if a bidding strategy is not defined for a product type
-tested with demand units and crm markets


maurerle:
sorry, did only see the review request

PR #67, opened 2023-06-07
Implement storages

adamsjohanna:
storages are implemented with the strategy from flexible. Since there is no price forecast to compare with, the storage units discharge once and then do nothing. This will change when the forecast is integrated.


maurerle:
Hi Johanna, can you run:
`black assume tests examples`
`isort assume tests examples`
and merge main into your branch, to resolve the conflicting world.py?

assume/strategies/flexable_storage.py
```diff
@@ -0,0 +1,104 @@
...
+
+        Return: volume, price
+        Strategy analogue to flexABLE
+
+        """
+        if operational_window is not None:
+            # =============================================================================
+            # Storage Unit is either charging, discharging, or off
+            # =============================================================================
+            self.current_time = operational_window["window"]["start"]
```
maurerle, line 32: That is not correct.
The current time is not equal to the start of the current bidding window for which the bid calculation was requested.
I think you can fix this by copying the changes from the powerplant, in which this has een changed

assume/strategies/flexable_storage.py
```diff
@@ -0,0 +1,104 @@
...
+
+                bids = [{"price": average_price, "volume": -bid_quantity}]
+
+            else:
+                bids = []
+
+        return bids
+
+    def calculate_price_average(self, unit):
+        t = self.current_time
```
maurerle, line 80: I would add the time_delta as a parameter to this function and abandon self.current_time completely..?

Also make sure that the functions are still working if the time horizon for the bids are not exactly 1h but 4h - then you need to work with an aggregation of the time slice - can be seen in powerplant.py now :)

maurerle (changes requested):
This is based on a main state where the powerplant did not support bids for more than 1 hour and did have some weird behaviour with the concept of "current_time".
Would be cool if you can adjust this for the storages too..?

assume/strategies/flexable_storage.py
adamsjohanna: This is then also wrong in flexable.py. I'll try to correct both then

PR #69, opened 2023-06-09
add market config to calculate_bids

maurerle:
* add scaling factor to bidding strategy
* allow params when creating bidding strategies
* add long term case

### DIFF
--- a/assume/units/powerplant.py
+++ b/assume/units/powerplant.py
@@ -329,7 +329,7 @@ class PowerPlant(BaseUnit):
         base_load = self.total_power_output[start:end]
         # check if min_power is a series or a float
         min_power = (
-            self.min_power.at[start]
+            self.min_power[start:end].max()
             if type(self.min_power) is pd.Series
             else self.min_power
         )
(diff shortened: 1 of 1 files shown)


kim-mskw (commented):
Can you sketch out why you need the policy agent again?

maurerle:
That was a WIP file which should not be part of this commit - I removed it

assume/units/powerplant.py
```diff
@@ -147,9 +147,9 @@ def calculate_energy_operational_window(
         min_power, max_power = self.calculate_min_max_power(start, end)
 
         # adjust for ramp down speed
-        min_power = max(current_power - self.ramp_down, min_power)
+        min_power = max(-self.ramp_down, min_power)
```
nick-harder, line 150: why is this like that? Why is current power not taken into account now?

assume/units/powerplant.py
```diff
@@ -169,14 +169,14 @@ def calculate_energy_operational_window(
             "min_power": {
                 "power": min_power,
                 "marginal_cost": self.calc_marginal_cost(
-                    power_output=min_power,
+                    power_output=current_power + min_power,
```
nick-harder, line 172: and this construct is also weird. why?

nick-harder (changes requested):
I am not sure about new formulation for max and min power in the power plant. Please adress in the comments.

assume/units/powerplant.py
maurerle: because min_power/max_power from calculate_min_max_power is also relative.
This is needed so that we are not bidding twice our total power with two markets

nick-harder (approved):

PR #71, opened 2023-06-16
include reserve capacity in dispatch plan

HendrikWulfert:


maurerle (approved):
Looks good, did you run the crm example to make sure that this change has the expected result?
- so the agents should not bid the total power if they need to preserve capacity
- the agents should bid at least the `minimum power + neg_capacity reserved` so that they can actually reduce that much

If both is the case you can click the merge button :)

PR #74, opened 2023-06-21
Modified fuel prices handling and forecast agent

nick-harder:
-allow for fuel prices to be single values
-modified forecasting agent
-reimplemented merit order calculation in forecast agent -modified data handlng for forecast agent
-modified example 01 to simplify
-modified config in example 02 to remove llm cases as not relevant


assume/units/powerplant.py
```diff
@@ -136,7 +161,7 @@ def calculate_operational_window(
 
         if product_type == "energy":
             return self.calculate_energy_operational_window(start, end)
-        elif product_type in {"capacity_pos", "capacity_neg"}:
+        elif product_type in ["capacity_pos", "capacity_neg"]:
```
maurerle, line 164: I don't mind, but isn't a set more correct here (unordered, unique values)?
In-Set checks are also faster than lists, but this does not matter at all with less than 10 values (as a rule of thumb)

assume/world.py
```diff
@@ -264,28 +274,32 @@ def creator(container):
...
+            )
+            forecast_agent = RoleAgent(
+                self.container, suggested_aid=f"forecast_agent_{market_id}"
+            )
+            forecast_agent.add_role(forecast_provider)
+            self.forecast_providers[market_id] = forecast_provider
+
+            temp[market_id] = forecast_provider.price_forecast_df
+
+        temp.to_csv(f"{path}/price_forecasts.csv", index=True)
```
maurerle, line 302: why are we keeping this additional file?

assume/world.py
```diff
@@ -593,7 +607,7 @@ async def run_simulation(self):
 
         start_ts = calendar.timegm(self.start.utctimetuple())
         end_ts = calendar.timegm(self.end.utctimetuple())
-        pbar = tqdm(total=end_ts - start_ts)
+        pbar = tqdm(total=end_ts - start_ts, leave=True)
```
maurerle, line 610: https://tqdm.github.io/docs/tqdm/
> leave: bool, optional
> If [default: True], keeps all traces of the progressbar upon termination of iteration. If None, will leave only if position is 0.

So this does not add anything?

assume/world.py
```diff
@@ -609,9 +623,11 @@ async def run_simulation(self):
                 self.clock.set_time(end_ts)
 
             await tasks_complete_or_sleeping(self.container)
-        pbar.close()
+        # pbar.close()
```
maurerle, line 626: This is actually correct to close the context here (just like you would close a open file handle) - but as the program terminates anymore, the context is closed there too - still I would revert this line?

maurerle (changes requested):
Looks good in general, but I left same remarks for improvement.

I always feel like this is slowing things down a lot, on the other hand, it is also a needed review process somehow..

assume/world.py
nick-harder: so it won't waste time calculating it again and again each time you run a simulation

assume/world.py
nick-harder: oh yeah true, in my build in docs it showed default value as false.

nick-harder:
@maurerle fixed the suggestions

maurerle (approved):

PR #75, opened 2023-06-27
Added CRM bidding strategies from flexible

nick-harder:


nick-harder:
Hi @kim-mskw, könntest du es dir bitte anschauen?

assume/common/outputs.py
kim-mskw, line 1: When I want to run the same study case with a different scenario it will also be overwirtten if it is programmed like this?

nick-harder: right now how it was done is that it only takes the case study as simulation id and writes it, so it means we cannot have same case study from two different scenarios, as one will overwrite another. I didn't finish adding it since I wanted to discuss it first, but basically the idea was, that we also add a tag of scenario and a tag of case study, so in grafana we have two lists. Right now I am only passing the case study and scenario as extra compared to only case study as before. If you @kim-mskw could finilize it, it would be great

maurerle: Why do we need to have scenario and study_case separated?
I am not quite sure we do need it at all, but instead of the current change to separete it - i would rather just use f"{scenario}_{study_case}" as simulation_id?

nick-harder: what would be the technical challenge to have both? I had it like this during my work on 2 seperate papers, and it was really nice when you have like 15 different cases pro paper to not scroll through all 30 (especially if they are nor ordered the exact right way), but only seeing those relevant to you.

maurerle: Well there are multiple locations which just use a unique simulation_id - having a pair of two strings everywhere changes a lot here (when you filter the database you need to remember to filter across both fields and not just `scenario == ...`).
I am not in favor of this added complexity because I am sure I will forget.
But if thats what we want to have, its fine for me too.

I see the benefit is to group simulations which belong together for research - this could also be done based on a substring in grafana and in queries, which is what I did for my latest study.

nick-harder: if this would significantly increase the complexity, I would be also against it. What do you think, @kim-mskw @Manish-Khanra ?

nick-harder: @Manish-Khanra sorry, but do you mean? I believe you might have got a bit confused. The discussion here is more about the how we define scenario and case study in the database. But maybe we should really take it to a different issue.

kim-mskw:
> if this would significantly increase the complexity, I would be also against it. What do you think, @kim-mskw @Manish-Khanra ? 

Well I can think of a couple of places where we would have to change this to enable considering different simulation. The simulation id can be whatever string you define, so it could be "scenario_case" if you want it to be. Seems easier to me, to do it like that.

assume/common/utils.py
```diff
@@ -116,23 +116,25 @@ def make_market_config(
...
             until=end,
             cache=True,
         ),
-        additional_fields=market_params.get("additional_fields", []),
-        product_type=market_params.get("product_type", "energy"),
         opening_duration=pd.Timedelta(market_params["opening_duration"]),
+        market_mechanism=market_params["market_mechanism"],
+        maximum_bid_volume=market_params["maximum_bid_volume"],
+        maximum_bid_price=market_params["maximum_bid_price"],
+        minimum_bid_price=market_params["minimum_bid_price"],
```
maurerle, line 131: this should be `.get(...)` calls to be optional

assume/strategies/flexable.py
```diff
@@ -127,27 +128,74 @@ def get_starting_costs(self, time, unit):
...
+class flexablePosCRM(BaseStrategy):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # check if kwargs contains crm_foresight argument
+        if "crm_foresight" in kwargs:
+            self.foresight = pd.Timedelta(kwargs["crm_foresight"])
         else:
-            price_forecast = unit.price_forecast.loc[t : t + self.foresight]
+            self.foresight = pd.Timedelta("4h")
```
maurerle, line 140: This can also be solved with:
`self.foresight = pd.Timedelta(kwargs.get("crm_foresight", "4h"))`
which does exactly this

assume/strategies/flexable.py
```diff
@@ -127,27 +128,74 @@ def get_starting_costs(self, time, unit):
...
+class flexableNegCRM(BaseStrategy):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.foresight = pd.Timedelta("12h")
+        # check if kwargs contains crm_foresight argument
+        if "crm_foresight" in kwargs:
+            self.foresight = pd.Timedelta(kwargs["crm_foresight"])
+        else:
+            self.foresight = pd.Timedelta("4h")
```
maurerle, line 197: see above comment

maurerle (changes requested):
I would address the comments regarding using `.get` on dictionaries with a default and revert the change to the outputs file

assume/common/utils.py
nick-harder: If I use .get and the value is not available, it return the value as None. And so since the value exists but is None, it overwrites the default values we defined in the class. Do you have any idea for a nicer solution to this instead of if else statements in the market class?

nick-harder:
@maurerle changes implemented, please review

maurerle (approved):

PR #85, opened 2023-07-04
Implement storages and their bidding strategies

nick-harder:


assume/common/outputs.py
```diff
@@ -58,6 +58,9 @@ def __init__(
         self.write_dfs: dict = defaultdict(list)
 
         if self.db is not None:
+            # TODO: #84 check if simulation_id is already in db
```
maurerle, line 61: I would not want to documente this twice - as you already created an issue for this

assume/units/storage.py
```diff
@@ -408,51 +349,53 @@ def set_dispatch_plan(
...
+                dispatch_plan["total_power"] * duration * self.efficiency_charge * (-1)
             )
+            self.current_SOC = self.current_SOC + charged_energy
 
-        elif self.outputs["energy"][start:end_excl].min() < self.min_power_discharge:
+        else:
             self.current_status = 0
             self.current_down_time += 1
-            self.outputs["energy"].loc[start:end_excl] = 0
+            self.outputs["energy"].loc[start:end] = 0
```
maurerle, line 378: I am quite sure that end_excl was needed in the section above..?

power
10:00 - 40
11:00 - 50
12:00 - 60

now start is 10:00 and end is 11:00 for our bid
if you set  self.outputs["energy"].loc[start:end] = 0 here - you would set the dispatch for the time from 11-12 to 0 too, which is wrong - same for other occurances of this behavior.

The end has to be exclusive. So the previous behavior was more correct?


Additionally, this does not work with multiple markets here - the checks should go into `execute_current_dispatch`

maurerle (changes requested):
I think the checks should go to `execute_current_dispatch`

assume/units/storage.py
nick-harder: I believe the assumption to have the same value for the next time step can still be valid as it helps the bidding strategies. Also, the correct and final value will be set after the respective market clearing. 
So I sold 10MW form 10 to 11, so for next bid calculation I don't want to assume my currrent power as 0, since this is not true, but 10, as this was my last state. So I use this 10 as current power and maybe sell now 15. But I see also some other points where this leads to errors. We should discuss this

nick-harder:
@maurerle adressed the raised issues, should be good now

maurerle (approved):

Issue #95, opened 2023-07-11
Merge units definition and their bidding strategies definition into one file


nick-harder:
I believe it is a bad idea to have a seperate file for bidding strategies as it is hard to prepare scenarios and not to forget to include the same demand, power plant and storage units in the bidding strategies file. I propose to merge the definition of bidding strategies into the units file as extra parameter. How do you find this proposal @kim-mskw @maurerle @Manish-Khanra?

kim-mskw:
I like this idea. I just wondered about something similar when coding the learning.

maurerle:
I think thats good too.
We should do this after merging the refactoring in #96

nick-harder:
Issue is now fixed and merged into main

PR #97, opened 2023-07-11
Fix example03 - new bidding_strategies structure

maurerle:
* change scenario files structure according to proposal in #95


assume/common/scenario_loader.py
```diff
@@ -378,13 +375,9 @@ async def load_scenario_folder_async(
     # if vre generation is provided, add them to the vre units
     # if we have RL strategy, add price forecast to unit_params
     def powerplant_callback(unit_name, unit_params):
-        # check if we have RL bidding strategy
-        if (
-            unit_params["bidding_strategies"]["energy"] == "rl_strategy"
-            and world.forecast_providers["EOM"].price_forecast_df is not None
-        ):
-            unit_params["price_forecast"] = world.price_forecast["mcp"]
-            unit_params["res_demand_forecast"] = world.price_forecast["residual_demand"]
```
maurerle, line 387: `unit_params["res_demand_forecast"] = world.price_forecast["residual_demand"]` went missing here? unneeded?

assume/common/scenario_loader.py
```diff
@@ -242,7 +236,11 @@ async def load_scenario_folder_async(
         file_name="demand_units",
     )
 
-    heatpump_units = load_file(path=path, config=config, file_name="heatpumps")
+    heatpump_units = load_file(
+        path=path,
+        config=config,
+        file_name="heatpumps",
```
maurerle, line 242:
```suggestion
        file_name="heatpump_units",
```

The file_name is now `heatpump_units` which is consistent with everything else

assume/common/scenario_loader.py
nick-harder: this was needed for the learning agents. We will make a clean implementation with Kim once we Strat with it next week

nick-harder:
@kim-mskw is now good to go :-)

nick-harder (approved):

PR #99, opened 2023-07-12
98 error in load scenario and other fixes

nick-harder:


assume/strategies/flexable.py
```diff
@@ -255,6 +255,8 @@ def get_specific_revenue(
     else:
         price_forecast = unit.price_forecast.loc[t : t + foresight]
 
+    price_forecast = price_forecast.values
+
     possible_revenue = sum(
         market_price - marginal_cost for market_price in price_forecast
     )
```
maurerle, line 262:
```suggestion
    possible_revenue = (price_forecast - marginal_cost).sum()
```

This should work here too? Please test this befor committing

assume/units/powerplant.py
```diff
@@ -125,6 +127,7 @@ def reset(self):
         self.current_down_time = self.min_down_time
 
         self.outputs["energy"] = pd.Series(0.0, index=self.index)
+        self.outputs["energy"].iloc[:1] = self.min_power
```
maurerle, line 130:
```suggestion
        # powerplant ist turned on with min_power at simulation start
        self.outputs["energy"].iloc[:1] = self.min_power
```

assume/units/powerplant.py
```diff
@@ -175,10 +178,13 @@ def calculate_energy_operational_window(
 
         min_power, max_power = self.calculate_min_max_power(start, end)
 
+        if self.id == "pp_02":
+            logger.debug(f"min_power: {min_power}")
+
```
maurerle, line 183:
```suggestion
```

assume/units/powerplant.py
```diff
@@ -175,10 +178,13 @@ def calculate_energy_operational_window(
...
         # adjust for ramp down speed
-        min_power = max(-self.ramp_down, min_power)
+        min_power = max(current_power - self.ramp_down, min_power)
         # adjust for ramp up speed
-        max_power = min(self.ramp_up, max_power)
+        max_power = min(current_power + self.ramp_up, max_power)
```
maurerle, line 187: the change from relative changes to absolute changes breaks compatibility for multiple markets, as this would bid the already bidded power on multiple markets for the same interval.

assume/units/powerplant.py
```diff
@@ -420,22 +426,16 @@ def calc_marginal_cost_with_partial_eff(
...
-            if type(self.capacity_factor) is pd.Series
-            else self.capacity_factor * self.max_power
+            self.availability[start] * self.max_power
+            if type(self.availability) is dict
+            else self.max_power
         )
-        max_delta = max_power - base_load.max()
 
-        return min_delta, max_delta
+        return min_power, max_power
```
maurerle, line 438: I don't understand why you changed to absolute values here again..

What do you expect base_load to be?

maurerle (changes requested):
Changing from relative bids to absolute bids breaks market bidding with multiple markets for the same interval.

Why is this preferred?

Except for the change in calculate_min_max_power and calculate_energy_operational_window, this looks good!

assume/strategies/flexable.py
nick-harder: Here you are mixing up bidding and state calculations. An operational window should reflect the possible states of my power plant in absolute terms. The bidding strategy in the following step takes care of the relative aspect. Such relative formulation at the operational window calculation completely broke the bidding strategies.

assume/units/powerplant.py
nick-harder: Here you are mixing up bidding and state calculations. An operational window should reflect the possible states of my power plant in absolute terms. The bidding strategy in the following step takes care of the relative aspect. Such relative formulation at the operational window calculation completely broke the bidding strategies.

maurerle: Which bidding strategy takes care of this? Naive does not, flexable does not.

With your change, you would bid max_power on the EOM market and bid the same max_power additionally on the CRM_pos? Or am I missing something?

nick-harder: flexable does, for EOM. it should be also fixed for CRM.

maurerle: See #100 to see how i would expect working (current behavior) - your change would create a min_power of 400 whysoever :shrug:

nick-harder: ok, jetzt sehe ich was du meinst, hab es korrigiert, dein test ist auch in diesem commit drin. Ich glaube, jetzt funktioniert es

assume/units/powerplant.py
```diff
@@ -418,24 +420,20 @@ def calc_marginal_cost_with_partial_eff(
...
+        end_excl = end - self.index.freq
+        base_load = self.outputs["energy"][start:end_excl]
         # check if min_power is a series or a float
-        min_power = (
-            self.min_power[start:end].max()
-            if type(self.min_power) is pd.Series
-            else self.min_power
-        )
-        min_delta = min_power - base_load.min()
+        min_power = self.min_power - base_load.min()
```
maurerle, line 426: if my availability is 0 i can not even bid min_power right?

nick-harder: true. added the check to the operation window function since if it is zero nothing can be done basically

maurerle (approved):

PR #100, opened 2023-07-13
adds power plant test

maurerle:
* tests that feedback for multiple markets works


nick-harder:
this test is included in the commit #99

PR #104, opened 2023-07-17
Preparing for learning functionality

nick-harder:
-changed scenario loading to include global system info in the context
-added res load forecast to the forecast agent
-added forwarding data_dict to the bidding strategies
-removed redundant call of calculate_bids inside units
-added learning_utils to the common functions
-added learning_strategy
-added rl params to the bidding strategy parameters


assume/common/learning_utils.py
maurerle, line 1: This file should go into something like `assume/strategies/learning/utils.py`.
I would like to have torch as an optional dependency, which is required for RL, but is not used in all other files too?

assume/common/scenario_loader.py
```diff
@@ -291,10 +291,15 @@ async def load_scenario_folder_async(
     save_frequency_hours = config.get("save_frequency_hours", None)
     sim_id = f"{scenario}_{study_case}"
 
-    bidding_params = config.get("bidding_strategy_params", {})
-    await world.setup(start, end, save_frequency_hours, sim_id, bidding_params, index)
-
-    # add Agents to world
+    bidding_strategy_params = config.get("bidding_strategy_params", {})
+    if "load_learned_path" not in bidding_strategy_params:
```
maurerle, line 295:
```suggestion
    if "load_learned_path" not in bidding_strategy_params.keys():
```
i would always be explicit here

assume/common/scenario_loader.py
```diff
@@ -359,8 +363,17 @@ async def load_scenario_folder_async(
...
+        unit_operator.context.data_dict["price_forecast"] = world.forecast_providers[
+            "EOM"
+        ].forecasts["price_forecast"]
+        unit_operator.context.data_dict[
+            "res_demand_forecast"
+        ] = world.forecast_providers["EOM"].forecasts["residual_load_forecast"]
+
     for company_name in set(all_operators):
         world.add_unit_operator(id=str(company_name))
+        unit_operator_callback(world.unit_operators[company_name])
```
maurerle, line 376: unit_operator = world.unit_operators[company_name]
and move line 367-373 here?

assume/common/units_operator.py
```diff
@@ -214,7 +214,11 @@ async def submit_bids(self, opening: OpeningMessage):
         products = opening["products"]
         market = self.registered_markets[opening["market_id"]]
         logger.debug(f"{self.id} setting bids for {market.name} - {products}")
-        orderbook = await self.formulate_bids(market, products)
+        orderbook = await self.formulate_bids(
+            market=market,
+            products=products,
+            data_dict=self.context.data_dict,
```
maurerle, line 220: this is not needed here

assume/common/units_operator.py
```diff
@@ -267,16 +276,18 @@ async def formulate_bids(self, market: MarketConfig, products: list[tuple]):
...
                 }
 
-                for unit_id, unit in self.units.items():
-                    # take price from bidding strategy
-                    bids = unit.calculate_bids(
+                bids = {
+                    unit_id: unit.calculate_bids(
                         market_config=market,
                         product_tuple=product,
+                        data_dict=data_dict,
```
maurerle, line 283: just pass `self.context.data_dict` here instead

maurerle (changes requested):
Its WIP, but your requested feedback, so I did but some feedback to it :)

assume/common/learning_utils.py
kim-mskw: that is kinda hard to do. Since we need it at several places due to our decisison on how to implement the transformation from GPU to CPU.

assume/common/units_operator.py
kim-mskw: Why?

assume/common/learning_utils.py
nick-harder: I would say that cpu version of torch is an easy dependency and would not make our package hard to install, so we can keep the cpu distribution of PyTorch inside (our main point is the learning functionality and we are supposed to also deliver learned strategies). In a simple run these can be executed on a cpu and pretty fast. We would pack the cuda dependency as a seperate package, since that one is a bit of a pain in one place to install. So if someone wants to run a full scale learning on a GPU(s) they would need the cuda distribution.

assume/strategies/learning_strategies.py
```diff
@@ -0,0 +1,255 @@
...
+        end = operational_window["window"][1]
+
+        # in rl_units operator in ASSUME
+        # TODO consider that the last forecast_length time steps cant be used
+        # TODO enable difference between actual residual load realisation and residual load forecast
+        # currently no difference so historical res_demand values can also be taken from res_demand_forecast
+        obs = []
+        forecast_len = 30  # in metric of market
+
+        obs.extend(data_dict.res_demand_forecast[start : end + forecast_len])
```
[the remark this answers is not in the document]
nick-harder: fixed

assume/strategies/learning_strategies.py
```diff
@@ -0,0 +1,255 @@
...
+        operational_window: OperationalWindow,
+    ):
+        start = operational_window["window"][0]
+        end = operational_window["window"][1]
+
+        # in rl_units operator in ASSUME
+        # TODO consider that the last forecast_length time steps cant be used
+        # TODO enable difference between actual residual load realisation and residual load forecast
+        # currently no difference so historical res_demand values can also be taken from res_demand_forecast
+        obs = []
```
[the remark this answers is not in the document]
nick-harder: fixed

assume/common/learning_utils.py
maurerle: sounds good

assume/common/scenario_loader.py
kim-mskw: I would suggest leaving that in the callback function, since it is the preparation of the forecast role functionality.

assume/common/units_operator.py
kim-mskw: Why?

maurerle: You never need to pass `self.x` attributes to functions with the same `self`, as you just passed `self.context.data_dict` below instead ;)
So this is addressed

assume/common/scenario_loader.py
maurerle: Yes, if that is needed for further things, thats fine

maurerle (approved):

PR #106, opened 2023-07-19
-fixed price forecast and res load forecast funcs

nick-harder:
-fixed storage unit set_dispatch


adamsjohanna (commented):
If I run example_01c for the crm_case, the values for the forecasts for the CRM_pos are empty and therefore function save_forecasts does not work. - I added a check
Also in units_operator.py write_learning_params() I added an if clause to check if "energy" is in bidding strategies - this must be changed later when RL should also work for CRM.
Should agents later have different learnings for different markets?

adamsjohanna (approved):

PR #109, opened 2023-07-21
Feature/multi bidding

maurerle:
first steps to multi bidding of various products in one market opening


assume/common/utils.py
```diff
@@ -66,7 +66,7 @@ def plot_orderbook(orderbook: Orderbook, results: list[dict]):
     from matplotlib.lines import Line2D
 
     bids = defaultdict(list)
-    orderbook = sorted(orderbook, key=itemgetter("node_id"))
+    orderbook.sort(key=itemgetter("node_id"))
```
nick-harder, line 69: I find always better to have x = x+1 formulations rather then inplace formulations. Somehow its easier to read those and understand. is there a particular reason for such formulation?

assume/units/powerplant.py
nick-harder, line 1: I find it a bad idea to have such a huge function for operational windows like this with several if else conditions. Makes reading the code way to hard, and also trying to maintain or change it. I would use the previous form

nick-harder (changes requested):
Please review the several comments, especially the one in calculate operation windows functions.

adamsjohanna (approved):

PR #110, opened 2023-07-26
Added CRM strategy for storages

adamsjohanna:
The strategy is split in flexablePosCRMStorage and fleaxableNegCRMStorage analogue to the strategy for powerplants.
In flexABLE the strategy for storages is too complex and cannot be integrated into our implementation, therefore the strategy is  not the same as in flexABLE.


assume/strategies/flexable_storage.py
```diff
@@ -33,54 +34,186 @@ def calculate_bids(
...
+        possible_revenue += (market_price - marginal_cost) * theoretic_power_discharge
+        theoretic_SOC -= theoretic_power_discharge
+
+    if unit.current_SOC - theoretic_SOC != 0:
+        possible_revenue = possible_revenue / (unit.current_SOC - theoretic_SOC)
+
+    return possible_revenue
+
+    """
+    def calculatingBidPricesSTO_CRM(self, t):
```
maurerle, line 195: should we keep this here?

assume/units/powerplant.py
```diff
@@ -197,6 +197,9 @@ def calculate_energy_operational_window(
         # adjust max_power if sold positive reserve capacity on control reserve market
         max_power = max_power - self.outputs["capacity_pos"].at[start]
 
+        # ensure min_power < max_power
+        min_power = min(min_power, max(0, max_power))
```
maurerle, line 201: under which circumstances does this fail?
I would expect an assertion `assert min_power <= max_power` to fail never?

assume/units/storage.py
```diff
@@ -302,106 +324,227 @@ def calculate_energy_operational_window(
...
+        product_tuple,
+        **kwargs,
     ):
-        end_excl = end - self.index.freq
-        duration = (end - start).total_seconds() / 3600
+        return super().calculate_bids(
+            market_config=market_config,
+            product_tuple=product_tuple,
+        )
+    """
```
maurerle, line 445: can be deleted

maurerle (approved):
I mentioned some things, but otherwise this is fine, and you can merge it when you adressed the changes..?

Somehow the operational_window stuff still awaits my refactoring, but it takes longer than thought :/

maurerle (commented):
I also added those changes in my branch, we suggest the new mechanism which contains most features in the bidding strategies

assume/strategies/flexable_storage.py
adamsjohanna: No, it was just to have a comparison to flexABLE

assume/units/powerplant.py
adamsjohanna: But we add something to min_power and substract from max_power. So it is still possible to have min_power > max_power

assume/units/storage.py
adamsjohanna: Done

assume/units/powerplant.py
adamsjohanna: Is there a nicer formulation or can I leave it like that?

PR #116, opened 2023-08-03
Refactor ramp_up/ramp_down and add test for basic_unit

maurerle:


adamsjohanna (commented):
I think we still need the ramp up and down or a return of max_power and min_power for the crm market

maurerle:
The process is as follows:

1. `storage.calculate_min_max_discharge(...)` returns min_power, max_power without considering the ramping, because we don't know the amount of the previous step (which might be in the future too)
2. once we know the power of the previous_step (or calculated/received a value from a forecast) we need to check if our `min_power` is inbetween the ramping constraints

Here, both of them can be relevant, but mostly `ramp_down` is important for `min_power` and the `ramp_up` constraint is important for `max_power`.
So I think there is only one ramping function needed, which checks if a wanted value `power` is acceptable if we previously had `previous_power`

adamsjohanna:
Ok, so we then have to call the function twice if we want to check for min and max_power as it is right now, but only use the same function. Now I got it 👍

maurerle:
Exactly!
I updated the PR to be compatible with latest main

adamsjohanna (approved):

PR #126, opened 2023-08-07
add market clearing for a series of products day ahead market

nick-harder:


assume/markets/clearing_algorithms/__init__.py
```diff
@@ -1,13 +1,15 @@
 from assume.common.market_objects import market_mechanism
 
 from .all_or_nothing import pay_as_bid_aon, pay_as_clear_aon
-from .base import pay_as_bid, pay_as_clear
-from .pyomo_clearing import nodal_pricing_pyomo
+from .complex_clearing import pay_as_clear_opt
+from .nodal_pricing import nodal_pricing_example
+from .simple import pay_as_bid, pay_as_clear
```
maurerle, line 6: moving base to simple is fine
having a separate file per complex clearing does not make a lot of sense to me.
I liked they way of categorizing by the pyomo dependency better..?
Not a problem

assume/markets/clearing_algorithms/complex_clearing.py
```diff
@@ -0,0 +1,147 @@
...
+                        model.P[order["bid_id"]],
+                        order["volume"],
+                    )
+                else:
+                    min_max_power_expr = (
+                        order["volume"],
+                        model.P[order["bid_id"]],
+                        order["acceptance_ratio"] * order["volume"],
+                    )
+
```
maurerle, line 71: doesn't this miss out the option to not use a bid at all?
As I see, the value should always be between `acceptance_ratio * volume` and `volume`, but it should also be possible to set the volume to 0?

assume/markets/clearing_algorithms/nodal_pricing.py
```diff
@@ -10,7 +10,9 @@
 SOLVERS = ["glpk", "cbc", "gurobi", "cplex"]
 
 
-def nodal_pricing_pyomo(market_agent: MarketRole, market_products: list[MarketProduct]):
+def nodal_pricing_example(
```
maurerle, line 13: why should this be called "example"?
Its an implementation of nodal pricing using pyomo for the optimization, so I would prefer the previous name?

examples/examples.py
```diff
@@ -0,0 +1,54 @@
...
+    example = "small"
+    data_format = "timescale"  # "local_db" or "timescale"
+
+    if data_format == "local_db":
+        DATABASE_URI = getenv(
+            "DATABASE_URI", f"sqlite:///./examples/local_db/assume_db_{example}.db"
+        )
+    elif data_format == "timescale":
+        DATABASE_URI = getenv(
+            "DATABASE_URI", "postgresql://assume:assume@localhost:5432/assume"
```
maurerle, line 44: I would remove the getenv here, as it is very unlikely to call this script with params

examples/examples.py
```diff
@@ -0,0 +1,54 @@
...
+
+log = logging.getLogger(__name__)
+
+os.makedirs("./examples/outputs", exist_ok=True)
+EXPORT_CSV_PATH = str(getenv("EXPORT_CSV_PATH", "./examples/outputs"))
+
+os.makedirs("./examples/local_db", exist_ok=True)
+
+availabe_examples = {
+    "small": {"scenario": "example_01a", "study_case": "example_01a"},
```
maurerle, line 16: shouldn't we just say `small` instead of `example_01a`. Same for all the other scenarios..?

examples/inputs/example_01a/config.yml
```diff
@@ -18,26 +18,4 @@ example_01a:
...
-          first_delivery: 1h
-      opening_frequency: 24h
-      opening_duration: 24h
-      volume_unit: MWh
-      maximum_bid_volume: 100000
-      maximum_bid_price: 3000
-      minimum_bid_price: -500
-      price_unit: EUR/MWh
-      market_mechanism: pay_as_clear
+      market_mechanism: pay_as_clear
```
maurerle, line 21: I would like to keep the day_ahead_market..? or is there any reason to remove it?

examples/inputs/example_01c/config.yml
```diff
@@ -20,7 +20,7 @@ eom_only_case:
       price_unit: EUR/MWh
       market_mechanism: pay_as_clear
 
-crm_case:
+example_01c_eom_crm:
```
maurerle, line 23: this is redundant too

examples/paper.py
```diff
@@ -1,97 +0,0 @@
...
-    plt.legend()
-    savefig("overview-average_cost.pdf")
-    savefig("overview-average_cost.png")
-    savefig(Path("overview-average_cost.svg"))
-    plt.clf()
-
-
-if __name__ == "__main__":
-    run_cases()
-    plot_and_save_results()
```
maurerle, line 97: I would have kept this as a reference of how to do an analysis and run multiple simulations at once but as all ltm_cases are not existing anymore thats fine

experimental/experimental_examples.py
maurerle, line 1: What does this file add in comparison to the examples.py above?

experimental/outputs/example_01_clearing_example_01a_complex_clearing/demand_meta.csv
maurerle, line 1: output files should not be comitted to the repo.
To remove them, you have to --amend or force push your branch, or we should squash the commit at the end

assume/common/outputs.py
maurerle, line 1: This is not needed anymore when using the main branch, it has been fixed before (by removing product from the market_orders, as they are not needed here)

maurerle (changes requested):
Still a few things needed to address.
Especially the output files.
I would suggest to split work into multiple smaller PRs in the future if the changes are not related to the market clearing anymore ;)

maurerle:
From what I understand, for now the new pyomo market clearing should always have the same result as the one without optimization if `acceptance_ratio` is not set?

At that point we also have to decide if we want to have multiple smaller market mechanisms or tend to a bigger one which has supports all the common features together.

experimental/experimental_examples.py
nick-harder: in examples we have everything what works. And whatever we are using to implement functions or test something, this should be seperate from working examples. Such as new day ahead clearing or rl

assume/common/outputs.py
nick-harder: eventhough my branch was on the latest stand, the sqlite didn't work without this fix

assume/markets/clearing_algorithms/nodal_pricing.py
nick-harder: because this example works only for a specific number of nodes (3), which is not a general case and should be only used as an example

assume/markets/base_market.py
```diff
@@ -254,6 +254,8 @@ async def clear_market(self, market_products: list[MarketProduct]):
 
         await self.store_market_results(market_meta)
 
+        self.all_orders = []
```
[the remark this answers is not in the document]
nick-harder: true, fixed

assume/markets/clearing_algorithms/nodal_pricing.py
maurerle: Then, we should fix this in a second issue instead of renaming it..? ;)

assume/markets/clearing_algorithms/complex_clearing.py
nick-harder: yes true, missed that. the nodal clearing one has the same problem in this case

experimental/experimental_examples.py
maurerle: If it does not work at all, it should be on a separate branch and not on main.
If it does work somehow or with WIP its fine to have it without a separate file.
On current main you can run `assume -s example_01a -c acceptance_ratio -db "postgresql://assume:assume@localhost:5432/assume"` which is WIP (agents are not aware of this) or edit the args at the bottom of `cli.py` to run something in debug mode.

I would not add an additional `experimental` folder at all..?

assume/markets/clearing_algorithms/nodal_pricing.py
nick-harder: yup agree

assume/markets/clearing_algorithms/complex_clearing.py
maurerle: I think it should work as 0 is respected here:

https://github.com/assume-framework/assume/blob/0b600abce5d5671be6596a8e5e233d5761b0f755/assume/markets/clearing_algorithms/pyomo_clearing.py#L147-L150

But adding an additional test for this would also be good, as I did not test this yet

nick-harder:
we should have fewer clearing algorithms with different functionality included to make life easier for the users. we should discuss at next time

maurerle:
You should merge the main branch to resolve and readd the day_ahead_market at that time?

maurerle (approved):

PR #130, opened 2023-08-10
Improve Data for Learning

maurerle:
This does add a LearningRole which receives current learned values.
It sends the relevant (observations, actions, rewards) via the learning role into a system wide buffer which stores it.

* it still works without torch installed, so this PR does not break anything
* basic example works fine and runs 10 epochs using `assume -s example_01_rl -db "postgresql://assume:assume@localhost:5432/assume"`
* adds a learning dashboard which can be used to investigate the results of different epochs

Learning consists of observations, actions and rewards.
These triples are always tied to one market opening.
* Observations include buffer of relevant forecasts/residual load
* Actions consist of 2 values which are calculated for now (min_p and max_p)
* Reward is always a single value which describes the outcome of the actions (the market clearing)

This would allow to model learning for multiple markets, markets with multiple products and things like that.
The action size is still hard coded and the `calculate_reward` function does not work correctly for multiple markets yet.

This does not yet include actual learning, as the LearningRole does not sample next actions from the buffer using MATD3

I would like to transfer the changes to units_operator into a subclass for learning or something like that..


assume/strategies/learning_strategies.py
kim-mskw, line 1: Why do we not take the kwargs for the learning mode anymore?

README.md
```diff
@@ -90,6 +90,12 @@ to run pre-commit directly, you can use
 pre-commit run --all-files
 ```
 
+Use Learning Capabilities
+---------------------------------
+If you want to use the reinforcement learning strategies of Assume please make sure to install torch on top of the requirements listed in the requirements.txt. If you Ressources do not have a GPU available use the CPU only version of torch, which can be installed like this:
+
+pip install torch==2.0.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+
```
maurerle, line 98: Actually for me it just works fine having `pip install torch` - which install torch=2.0.1 from pypi without problems.
So it would be good to double check in a new env if this might be enough..

Then it would be just `pip install -e .[learning]` to install assume with learning dependencies

kim-mskw: On windows it definetly did not work for me with just the pip install. But good bpoint let's see.

assume/strategies/learning_strategies.py
```diff
@@ -302,12 +310,5 @@ def calculate_reward(
         ):
             profit = profit - unit.hot_start_cost / 2
```
kim-mskw, line 311: Moved reward calculation to avoid redundance defintion

kim-mskw:
Something does not work here: 
at some point randomly we get ththis error in the middle of the simulation:

  File "C:\Users\tg3533\assume\assume\reinforcement_learning\buffer.py", line 103, in add
    self.rewards[self.pos] = reward.copy()
    ~~~~~~~~~~~~^^^^^^^^^^
TypeError: float() argument must be a string or a real number, not 'dict'

I do not know what triggers it or why it appears, but we need to fix that somehow.

assume/common/units_operator.py
```diff
@@ -290,28 +301,67 @@ async def formulate_bids(
...
+                        index=[start],
+                    )
+                    data["unit"] = unit_id
+                    unit_rl_strategy_dfs.append(data)
+                    all_observations.append(unit.outputs["rl_observations"][start])
+                    all_actions[i, :] = unit.outputs["rl_actions"][start]
+                    all_rewards.append(unit.outputs["rl_rewards"][start])
+
+                    i += 1
+            all_observations = np.array(all_observations)
```
adamsjohanna, line 357: Here, I have an error because all_observations consists of several tensors.
If line 452 is changed to "all_observations.append(np.array(unit.outputs["rl_observations"][start]))" it runs without error.

README.md
adamsjohanna: For me it did work on windows :-)

adamsjohanna (approved):

assume/common/units_operator.py
maurerle: Thats somewhat weird because we are converting it to an array in 357 but if this is needed, thats finde

PR #131, opened 2023-08-11
Improve forecasts

maurerle:
add price forecaster
    * Forecaster Class which contains all needed data
    * known to every unit
    * removes need for data_dict in bidding strategies
    * allows to get forecasts ad hoc, calculated or from database in the future
    * needed to integrate other strategies relying on various forecasts

TODO: remove prices and availability from units, as they are getting it from their forecast too..?


maurerle:
After thinking about it for the weekend, I quite like it, even if it seems to complicate things.

Separating the Data of Forecasts (Forecaster), the Operation Parameters (Unit) and the bidding strategy (BiddingStrategy) makes somehow sense, and gives a lot of flexibility.

This can be merged if there is nothing against it..?

nick-harder (approved):

PR #139, opened 2023-08-16
Add Units and Market from DMAS

maurerle:
This adds working and tested units, which resemble the same behavior as given here:
https://github.com/NOWUM/dmas

This allows to create a comparison of our framework to DMAS - a market model for the day-ahead market developed at the FH-Aachen.

Before merging this, #138  should be merged.
I am still creating a market scenario with this, but want to get some feedback on this


assume/common/utils.py
```diff
@@ -172,6 +173,30 @@ def plot_orderbook(orderbook: Orderbook, results: list[dict]):
     return fig, ax
 
 
+def visualize_orderbook(order_book: Orderbook, hour_count: int = 24):
```
adamsjohanna, line 176: What exactly should be displayed here?
When running the tests, no "block_id" or "link" is given, so the line 185 returns an error.

nick-harder:
multiple tests are failing, please investigate

nick-harder (approved):
great, writing tests directly for the new commits!

PR #141, opened 2023-08-16
rename min_soc and max_soc in storages

### LINKED ISSUE #137: [storage] Rename min_SOC and max_SOC to min_volume and max_volume
Currently, the `max_SOC` variable is given in MWh? instead of %, so I would suggest to switch the name to `max_volume` instead?

adamsjohanna:
Using SOC purely as % and instead of min_soc and max_soc, min_volume and max_volume 

closes #137


maurerle (approved):
Thanks!

PR #153, opened 2023-08-21
152 define bid type to 'SB' if not set

adamsjohanna:
The bid type is now defined as 'SB' in base_market in case it is not defined but requested.
Also the orders in marketconfig.all_orders and the model are deleted after each calculation. If not, the bid_ids are not unique and the model creates some errors.

### DIFF
--- a/assume/markets/base_market.py
+++ b/assume/markets/base_market.py
@@ -175,6 +175,9 @@ class MarketRole(Role):
                 ), f"minimum_bid_price {order['price']}"
 
                 if "bid_type" in order.keys():
+                    order["bid_type"] = (
+                        "SB" if order["bid_type"] == None else order["bid_type"]
+                    )
                     assert order["bid_type"] in [
                         "SB",
                         "BB",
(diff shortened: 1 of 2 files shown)


assume/markets/base_market.py
```diff
@@ -245,6 +248,11 @@ def order_matches_req(o):
         )
 
     async def clear_market(self, market_products: list[MarketProduct]):
+        # Check if order is in time slots for current opening
+        # for order in self.all_orders:
```
maurerle, line 252: This should be part of `handle_orderbook` - see L153
As multiple clearings could be open in the same time, we should see if the sent orders are part of `self.open_slots = []` defined  in L52 (currently unused)

assume/markets/base_market.py
```diff
@@ -175,6 +175,9 @@ def handle_orderbook(self, content: dict, meta: dict):
                 ), f"minimum_bid_price {order['price']}"
 
                 if "bid_type" in order.keys():
+                    order["bid_type"] = (
+                        "SB" if order["bid_type"] == None else order["bid_type"]
+                    )
```
maurerle, line 180: As bid_type is not a default param, I think this would not belong here. But this will be fixed when we switch to a market_mechanism class. So its fine

maurerle (approved):

PR #157, opened 2023-08-23
Continue learning

maurerle:
this includes various fixes for learning

* add param `disable_learning` to disable creating a learningrole in validation run
* manipulate buffer through `set_buffer`
* update actor, target_actor, critic, target_critic accordingly
* add critic per Unit to learningrole
* add `update_transition` to playback final result to LearningStrategies again (WIP - i don't know what should be set in LearningStrategy)

Currently, this returns 
`RuntimeError: mat1 and mat2 shapes cannot be multiplied (256x62 and 70x512)`
at `x1 = F.relu(self.FC1_1(xu))`

### DIFF
--- a/assume/common/scenario_loader.py
+++ b/assume/common/scenario_loader.py
@@ -255,6 +255,7 @@ async def load_scenario_folder_async(
 
     learning_config = config.get("learning_config", {})
     bidding_strategy_params = config.get("bidding_strategy_params", {})
+
     if "load_learned_path" not in learning_config.keys():
         learning_config[
             "load_learned_path"
--- a/assume/world.py
+++ b/assume/world.py
@@ -124,7 +124,7 @@ class World:
         )
 
         # initiate learning if the learning mode is on and hence we want to learn new strategies
-        if self.learning_config:
+        if self.learning_config.get("learning_mode", False):
             # if so, we initate the rl learning role with parameters
             from assume.reinforcement_learning.learning_role import Learning
 


assume/reinforcement_learning/algorithms/matd3.py
```diff
@@ -24,183 +25,111 @@ def __init__(
...
+    def update_policy(self):
+        logger.info(f"Updating Policy")
+        n_rl_agents = len(self.learning_role.rl_units)
+        for _ in range(self.gradient_steps):
+            i = 0
+            strategy: LearningStrategy
+            for u_id, strategy in self.learning_role.rl_units.items():
+                critic_target = self.learning_role.target_critics[u_id]
+                critic = self.learning_role.critics[u_id]
+                actor = self.learning_role.rl_units[u_id].actor
```
kim-mskw, line 61: Okay, so you solved this now by giving all units to the learning role and the learning role is in the algorithm and hence you can access the actor directly. Seems like a good work around for now. 

Why do we still have the if i%100==0: though I thought this is handeld by the sheduling of the recuurent task?

nick-harder: this ensures that when you have many agents, they don't learn on the same sample from the buffer. we can remove this for now, but it improved the performance for me

nick-harder: also, the main point was not to compute next actions so many times as this is the most intensive task and requires a ton of time

nick-harder:
I fixed the issue with not matching matrix sized and some other bugs, the code is now running. If it actually learns is not validated

assume/reinforcement_learning/algorithms/matd3.py
kim-mskw: Ah okay, I misinterpreted this then

nick-harder:
warum failed 3.11 immer? codecov will immer was...

PR #158, opened 2023-08-24
Dokumentation improvements

DominikBurkert:


maurerle (changes requested):
Thank you for the improvements!

It probably makes sense to stick to the documentation of Sphinx (the tool we use to create the docs) and use their example as a reference on how to document things - and which items should be documented where:
https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#an-example-class-with-docstrings

I am not an expert on documentation either, but I think that it would be better to not document the list of available methods, as this is generated automatically afaik..?

Maybe, you can run the documentation locally (see https://github.com/assume-framework/assume#create-documentation) to see how it looks best?

kim-mskw:
> Thanks Florian we ware planning on checking how it looks and thought pushing it would be the easiest way. but testing it locally makes more sense!

It probably makes sense to stick to the documentation of Sphinx (the tool we use to create the docs) and use their example as a reference on how to document things - and which items should be documented where:
https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#an-example-class-with-docstrings

I am not an expert on documentation either, but I think that it would be better to not document the list of available methods, as this is generated automatically afaik..?

Maybe, you can run the documentation locally (see https://github.com/assume-framework/assume#create-documentation) to see how it looks best?

README.md
```diff
@@ -56,13 +56,30 @@ Afterwards, activate the environment:
 conda activate assume-framework
 ```
 
-After these steps you can also run the example simulation:
+Quick Start
+-----------
+There are 3 ways to run an exemplar. <br>
+1.) local (without database and grafana) <br>
+2.) with docker (with database and grafana) <br>
+3.) use CLI to run simulations
```
maurerle, line 64: I would phrase that differently.
How you run the program, and which "storage" you are using for the results are two pairs of shoes ;)

README.md
```diff
@@ -56,13 +56,30 @@ Afterwards, activate the environment:
 conda activate assume-framework
 ```
 
-After these steps you can also run the example simulation:
+Quick Start
+-----------
+There are 3 ways to run an exemplar. <br>
```
maurerle, line 61:
```suggestion
There are 3 ways to run a simulation. <br>
```

assume/strategies/extended.py
```diff
@@ -18,15 +22,29 @@ def calculate_bids(
...
         bids = []
         for product in product_tuples:
             start = product[0]
             end = product[1]
 
-            min_power, max_power = unit.calculate_min_max_power(start, end)
-            current_power = unit.outputs["energy"].at[start]
+            min_power, max_power = unit.calculate_min_max_power(
+                start, end
+            )  # max_power describes the maximum power output of the unit
```
maurerle, line 44: I prefer longer comments to be before or after the code, most of the time.
`max_power describes the maximum power which can be generated throughout the timeframe from start to end`
would probably be better.

But it would also be better to have this comment inside of `calculate_min_max_power` so that one can hove above the method and see its documentation if one is unsure what the return params are..?

assume/strategies/flexable.py
```diff
@@ -7,6 +7,10 @@
 
 
 class flexableEOM(BaseStrategy):
+    """
+    A strategy that bids on the EOM-market.
```
maurerle, line 11: by placing two bids onto the market.
One for inflexable amount and a second one for flexable bidding amount at a higher price

assume/strategies/naive_strategies.py
```diff
@@ -3,33 +3,67 @@
 
 
 class NaiveStrategy(BaseStrategy):
+    """
+    A naive strategy that bids the marginal cost of the unit on the market.
+    """
+
+    """
```
maurerle, line 10: This does not look right..?

maurerle (approved):

PR #159, opened 2023-08-24
Minor orderbook corrections to ensure empty orders are not handed to the clearing algorithm

adamsjohanna:
The complex clearing does not work, if orders have a volume of 0. Since those orders are not necessary in any case (that I'm aware of), those empty orders are deleted before handing over the orderbook to the clearing mechanism.


assume/markets/clearing_algorithms/complex_clearing.py
```diff
@@ -217,11 +217,16 @@ def primal_eql_dual(m):
     eps = 1e-4
     solvers = check_available_solvers(*SOLVERS)
 
-    solver = SolverFactory(solvers[0])
+    solver = (
+        SolverFactory("gurobi") if "gurobi" in solvers else SolverFactory(solvers[0])
```
maurerle, line 221: If you want to prefer gurobi, you can reorder the list of Solvers at the top of the file ;)
I think it makes sense to prioritize commercial solvers if you have them available anyway.

assume/markets/clearing_algorithms/complex_clearing.py
```diff
@@ -217,11 +217,16 @@ def primal_eql_dual(m):
...
+    )
     if solver.name == "gurobi":
         options = {"cutoff": -1.0, "eps": eps}
     else:
         options = {}
+        raise ValueError(
+            f"Solver {solver} does not support cutoff, might not find solution."
```
maurerle, line 228: in other cases, setting a timelimit might be good, which returns the best solution found

maurerle (commented):
Maybe you can fix the SOLVERS list, other than that, it is fine

assume/markets/clearing_algorithms/complex_clearing.py
adamsjohanna: I'm not that familiar with all the solver options, but yea, it should be possible to run it with different solvers. Maybe we could also implement the option to choose solver settings in the configs? At least the time limit could be a nice parameter to enter manually.

maurerle (approved):

PR #165, opened 2023-08-31
Documentation improvements

fritzphilipp:


maurerle (approved):
Thanks!
I would not document internal threeliner methods ;)

PR #167, opened 2023-09-04
Integration of complex clearing

adamsjohanna:
The complex clearing allows to use block bids and min_acceptance_ratio through an optimization problem that is solved iteratively.
When using block orders, the order fields 'volume', 'accepted_volume' and 'accepted_price' is saved as dict. Therefore during the handling of the orderbook some adjustments had to be made.
To store the resulting orderbooks in the output files the block orders are separated into single orders (see utils.py, separate_block_orders() ).


adamsjohanna:
How is it possible, that the tests are successful on my laptop, but not when uploaded here?

maurerle:
Seems like this does not work with GLPK

glpsol complains on my machine because of:
`Invalid option '--eps'; try /usr/bin/glpsol --help`

nick-harder (approved):

PR #171, opened 2023-09-06
revalidate given steps and store best strategy

maurerle:


assume/common/scenario_loader.py
```diff
@@ -542,19 +552,37 @@ def run_learning(world: World, inputs_path: str, scenario: str, study_case: str)
...
-        avg_reward = world.output_role.get_episode_sum_reward()
-        path = world.learning_config["load_learned_path"]
-        if best_reward < avg_reward:
-            world.learning_role.save_params(directory=path)
-
-        if episode == world.learning_role.training_episodes - 1:
-            world.learning_role.save_params(directory=path)
-
+        
+        if episode % world.learning_config.get("validation_episodes_interval", 5) == 0:
```
nick-harder, line 562: This should be episode+1, otherwise it will run a validation at the very start as well

assume/common/scenario_loader.py
```diff
@@ -542,19 +552,37 @@ def run_learning(world: World, inputs_path: str, scenario: str, study_case: str)
...
+            old_path = world.learning_config["load_learned_path"]
+            new_path = f"{old_path}_eval"
+            # save validation params in validation path
+            world.learning_role.save_params(directory=new_path)
+            world.reset()
```
nick-harder, line 567: Do we still need this world.reset? Don't we anyway create everything from scratch?

nick-harder (changes requested):

assume/common/scenario_loader.py
maurerle: Yes, currently we are keeping the asyncio loop and the world object. We could change this, but this is a "different" reset than the ones used in the units. But it is needed to remove the existing agents and recreate them

nick-harder: oh ok, thanks

assume/common/scenario_loader.py
maurerle: addressed

nick-harder (approved):

PR #179, opened 2023-09-12
Building heatpump dr integration

nick-harder:
this branch introduces the functionality of DSM units and their operation


assume/optitest.py
maurerle, line 1: this should be moved to the test directory and converted to a pytest test

assume/common/outputs.py
maurerle, line 1: Thats good :)

assume/common/base.py
maurerle, line 1: remove Union here
maybe have technology always as a list[str]?

assume/common/forecasts.py
```diff
@@ -127,6 +148,30 @@ def calc_forecast_if_needed(self):
...
+
+    def get_eom_price_forcast(self) -> pd.Series:
+        """
+        Returns the heating demand of the building
+
+        :rtype: pd.Series
+        """
+        return self.forecasts.get("price_EOM", pd.Series(0, self.index))
+
     def get_registered_market_participants(self, market_id):
```
maurerle, line 175: this whole section seems redundant.. Doesnt the same happen if i run self["price_EOM"] for example? Why do I need the getter for this?

assume/world.py
```diff
@@ -223,7 +224,7 @@ async def async_add_unit(
         Create and add a new unit to the world.
 
         Params
-        ------
+        ------sssssssssssssssssssssssssssssssssssssssssss
```
maurerle, line 227: remove

assume/world.py
```diff
@@ -396,7 +397,7 @@ def add_unit(
                 id=id,
                 unit_type=unit_type,
                 unit_operator_id=unit_operator_id,
-                unit_params=unit_params,
+                unit_params=unit_params if unit_params is not None else {},
```
maurerle, line 400: This should not be necessary, by having a better default for unit_params

maurerle (changes requested):
I did not yet revise the new files, but looked at the changes to the existing framework

assume/units/dst_components.py
```diff
@@ -0,0 +1,314 @@
...
+        @self.component_block.Constraint(time_steps)
+        def cooling_factor_constraint(m, t):
+            return m.power_in[t] == m.cool_out[t] / m.cooling_factor
+
+    def define_objective(self):
+        # Define the objective function specific to HeatPump if needed
+        pass
+
+
+class Storage:
```
maurerle, line 152: This should not be called Storage, as we already have a Storage class

nick-harder:
stale. Reopen when active and updated

PR #181, opened 2023-09-13
-removed outdated nonfunctional units

nick-harder:
Electrolyzer and heauput units are outdated and are non functional. Their new versions will be added in the near future.


maurerle:
Please fix the import ;)

PR #186, opened 2023-09-14
-fixing docker file as renderer is not a must

nick-harder:


maurerle (approved):
Oh, that line slipped through.. Sorry

PR #191, opened 2023-09-15
Try to migrate the existing database, by adding missing columns

### LINKED ISSUE #188: Writing to db of comlex orders after running a simple simulation before that doesn't work
I get an error sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedColumn) column "min_acceptance_ratio" of relation "market_orders" does not exist

This only happens if you run a simple simulation beforehand. I believe this happens because this column doesn't exist in the db.

maurerle:
supports float and text columns
simplifies database engine usage as pandas now fully supports sqlalchemy v2

fixes #188

### DIFF
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import pandas as pd
 from dateutil import rrule as rr
 from mango import Role
+from pandas.api.types import is_numeric_dtype
 from sqlalchemy import inspect, text
 from sqlalchemy.exc import ProgrammingError
 
@@ -84,10 +85,10 @@ class WriteOutput(Role):
 
         # Loop throuph all database tables
         # Get list of table names in database
-        table_names = inspect(self.db.bind).get_table_names()
+        table_names = inspect(self.db).get_table_names()
         # Iterate through each table
         for table_name in table_names:
-            with self.db() as db:
+            with self.db.begin() as db:
                 # Read table into Pandas DataFrame
                 query = text(
                     f"delete from {table_name} where simulation = '{simulation_id}'"
@@ -101,7 +102,7 @@ class WriteOutput(Role):
         query = text("select distinct simulation from market_meta")
 
         try:
-            with self.db() as db:
+            with self.db.begin() as db:
                 simulations = db.execute(query).fetchall()
         except Exception:
             simulations = []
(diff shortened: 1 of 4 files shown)


assume/common/outputs.py
```diff
@@ -212,9 +213,39 @@ async def store_dfs(self):
...
+        with self.db.begin() as db:
+            # Read table into Pandas DataFrame
+            query = f"select * from {table} where 1=0"
+            db_columns = pd.read_sql(query, db).columns
+        for column in df.columns:
+            if column not in db_columns:
+                try:
+                    # TODO this only works for float and text
+                    column_type = "float" if is_numeric_dtype(df[column]) else "text"
+                    query = f"ALTER TABLE {table} ADD COLUMN {column} {column_type}"
```
nick-harder, line 243: Do I understand it correctly, here we do not copy the whole thing again, but only directly commit his extra column to the db? So not downloading all entries and pushing again?

assume/common/outputs.py
```diff
@@ -212,9 +213,39 @@ async def store_dfs(self):
...
+                        df.to_sql(table, db, if_exists="append")
+
             self.write_dfs[table] = []
 
+    def check_columns(self, table: str, df: pd.DataFrame):
+        """
+        If a simulation before has been started which does not include an additional field
+        we try to add the field.
+        For now, this only works for float and text.
+        An alternative which finds the correct types would be to use
```
nick-harder, line 232: What other types there might be? anyway we cannot write dicts or enything else.

assume/common/outputs.py
maurerle: yes, we are adding a column through alter table. This requires knowledge of the data type of the new column and how this is specified in the respective Databas system (in postgresql/sqlite they are luckily named the same)

assume/common/outputs.py
maurerle: we do not detect bool or datetime values. But for most use cases it will be enough

nick-harder (approved):

assume/common/outputs.py
nick-harder: oh ok, cool, thanks!

assume/common/outputs.py
nick-harder: so maybe we can for now add a raise if the type is not detected? so we see if the problem ever occurs. Or it won't work anyway?

maurerle: If the field is bool (0 or 1) or int, we will add it as float numeric.
Datetime might be added as unix timestamp numeric or raise an error.

This will probably be enough to visualize the values, even if it is not the proper datatype.
And in other cases it will raise anyway. So I don't think we need additional error raising

PR #193, opened 2023-09-16
Enable saving results only at the end of the simulation

nick-harder:
-adds a function to store all results at the end
-fixes cvs_path creating an empty folder when not defined


assume/common/outputs.py
```diff
@@ -40,18 +40,18 @@ def __init__(
         start: datetime,
         end: datetime,
         db_engine=None,
-        export_csv_path: str = "",
-        save_frequency_hours: int = 24,
+        export_csv_path: str = None,
+        save_frequency_hours: int = None,
```
maurerle, line 44: export_csv_path = "" did the job, why the change..?

assume/common/outputs.py
```diff
@@ -40,18 +40,18 @@ def __init__(
...
         learning_mode: bool = False,
     ):
         super().__init__()
 
         # store needed date
         self.simulation_id = simulation_id
-        self.save_frequency_hours = save_frequency_hours
+        self.save_frequency_hours = save_frequency_hours or (end - start).days * 24
```
maurerle, line 51: When would you want to disable saving results in given intervals?
Stating "same_process=False" should fix consuming performance, by using a separate process

The remaining data will always be written in `on_stop` of the outputs role

assume/common/outputs.py
```diff
@@ -40,18 +40,18 @@ def __init__(
...
 
         # make directory if not already present
-        self.export_csv_path = export_csv_path
+        self.export_csv_path = export_csv_path or 0
         if self.export_csv_path:
             self.p = Path(self.export_csv_path, simulation_id)
             shutil.rmtree(self.p, ignore_errors=True)
```
maurerle, line 57: here we are removing the folder and recreating it properly

assume/common/outputs.py
```diff
@@ -40,18 +40,18 @@ def __init__(
...
         super().__init__()
 
         # store needed date
         self.simulation_id = simulation_id
-        self.save_frequency_hours = save_frequency_hours
+        self.save_frequency_hours = save_frequency_hours or (end - start).days * 24
 
         # make directory if not already present
-        self.export_csv_path = export_csv_path
+        self.export_csv_path = export_csv_path or 0
```
maurerle, line 54:
```
export_csv_path = ""
if export_csv_path:
  print("hi")
```

does the same, the `or 0` does not add anything..?

examples/examples.py
```diff
@@ -7,7 +7,8 @@
 log = logging.getLogger(__name__)
 
 csv_path = "./examples/outputs"
-os.makedirs(csv_path, exist_ok=True)
+if csv_path:
+    os.makedirs(csv_path, exist_ok=True)
```
maurerle, line 11: csv_path is always defined above..
you should just remove the makedirs line entirely, as this is taken care of in outputs.py in line 58

maurerle (changes requested):

assume/common/outputs.py
nick-harder: for me it always creates an empty folder in the root without a name and writes the files over there. So it doesn't work

assume/common/outputs.py
nick-harder: as of right now this saving each day consumes a big chunk of time as multithread is not working. We should have this until second thread works. Also, sometime you really don't need all that data every day of the run...

assume/common/outputs.py
nick-harder: as I said before, on my system defining export_csv_path as "" makes a no name folder. This fixes it

examples/examples.py
nick-harder: cool

assume/common/outputs.py
nick-harder: so the problem was actually because of the mkdir in the main examples.py file and not here. I have now removed it from the main file and it seems to work fine

assume/common/outputs.py
nick-harder: I believe this is also not because of the db, but the csv files, those require a lot of time

nick-harder:
@maurerle soll es durch?

assume/common/outputs.py
maurerle: I don't think that we should set this to None..

maurerle:
Ich wollte es noch nachstellen, ob das = None tatsächlich notwendig ist..

nick-harder:
ach ja stimmt, jetzt ist 24 wieder default

maurerle (approved):

PR #203, opened 2023-09-25
allow running simulation without port assignment

### LINKED ISSUE #202: Several runs in parallel
I tried to start several runs in parallel, but it raised an error that port 9099 is occupied. I solved this issue by passing a different port explicitly when initializing world. Is this the intended way of doing that @maurerle? Could we somehow automate this so people can start several simulations in a script without manually defining a port? Please give me some hints and I would implement this.

maurerle:
This fixes #202 by using the external_connection Container, which does not require a port.

Until https://gitlab.com/mango-agents/mango/-/merge_requests/91 is merged and released, we therefore need to switch to mango-agents from gitlab


nick-harder (approved):

nick-harder:
@maurerle failing

PR #205, opened 2023-09-25
Disable writing data during learning process

nick-harder:
-no more outputs written when learning
-adjusted learning dashboard

### DIFF
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -341,19 +341,20 @@ class WriteOutput(Role):
         await self.store_dfs()
         if self.db is None:
             return
-        queries = [
-            f"select 'avg_price' as variable, market_id as ident, avg(price) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
-            f"select 'total_cost' as variable, market_id as ident, sum(price*demand_volume_energy) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
-            f"select 'total_volume' as variable, market_id as ident, sum(demand_volume_energy) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
-            f"select 'capacity_factor' as variable, market_id as ident, avg(power/max_power) as value from market_dispatch ud join power_plant_meta um on ud.unit_id = um.\"index\" and ud.simulation=um.simulation where um.simulation = '{self.simulation_id}' group by variable, market_id",
-        ]
-        dfs = []
 
-        learning_queries = self.learning_queries()
-        if learning_queries:
-            queries.extend(learning_queries)
+        queries = []
+        if self.learning_mode:
+            queries.extend(self.learning_queries())
+        else:
+            queries = [
+                f"select 'avg_price' as variable, market_id as ident, avg(price) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
+                f"select 'total_cost' as variable, market_id as ident, sum(price*demand_volume_energy) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
+                f"select 'total_volume' as variable, market_id as ident, sum(demand_volume_energy) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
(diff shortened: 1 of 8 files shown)


nick-harder:
@maurerle I have also adjusted the learning dashboard, but I couldn't get the last two graphs to work - Bid price and bid volume, and also the noise. Could you pleae take a look at them? Ideally the bids ones should display all bids submitted by the learning agent. Thanks!

assume/common/outputs.py
```diff
@@ -341,19 +341,20 @@ async def on_stop(self):
...
+        queries = []
+        if self.learning_mode:
+            queries.extend(self.learning_queries())
+        else:
+            queries = [
+                f"select 'avg_price' as variable, market_id as ident, avg(price) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
+                f"select 'total_cost' as variable, market_id as ident, sum(price*demand_volume_energy) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
+                f"select 'total_volume' as variable, market_id as ident, sum(demand_volume_energy) as value from market_meta where simulation = '{self.simulation_id}' group by market_id",
+                f"select 'capacity_factor' as variable, market_id as ident, avg(power/max_power) as value from market_dispatch ud join power_plant_meta um on ud.unit_id = um.\"index\" and ud.simulation=um.simulation where um.simulation = '{self.simulation_id}' group by variable, market_id",
+            ]
```
maurerle, line 354: so we never have learning results and avg_prices and stuff at the same time..?

assume/common/outputs.py
```diff
@@ -372,13 +373,10 @@ async def on_stop(self):
                 with self.db.begin() as db:
                     df.to_sql("kpis", self.db, if_exists="append", index=None)
         except ProgrammingError as e:
-            self.db.rollback()
+            # self.db.rollback() not working, no rollback function
```
maurerle, line 376: remove this line - its not needed anymore , as we have the proper `with self.db.begin()`

maurerle (changes requested):
I would find it much better, if we split this up by
`output_agent_addr` and
`learning_output_agent_addr` in the data_dict

As we are already checking if a output_agent_id and addr is given, we can use this to distinguish if we should write..?

then we can set this in world, that market and units operator do not have an output agent and units_operator only knows a `learning_output_agent` - therefore writes only learning..?
I can prepare a draft for this if this is unclear?

assume/common/outputs.py
nick-harder: no, since no market orders and market results are written, we don't have this info during training at all. During evaluation runs it is written

maurerle: I thought, that we would need reward and results at the same time during evaluation runs..?

assume/world.py
```diff
@@ -213,7 +216,12 @@ def add_unit_operator(
         unit_operator_agent._role_context.data_dict = {
             "output_agent_addr": self.output_agent_addr[0],
             "output_agent_id": self.output_agent_addr[1],
-            "learning_mode": self.learning_mode,
+            "learning_agent_addr": self.learning_agent_addr[0]
+            if self.learning_mode
+            else None,
+            "learning_agent_id": self.learning_agent_addr[1]
+            if self.learning_mode
+            else None,
```
maurerle:
```suggestion
        unit_operator_agent._role_context.data_dict = {}
        if not self.learning_mode:
            unit_operator_agent._role_context.data_dict.update({
                "output_agent_addr": self.output_agent_addr[0]
                "output_agent_id": self.output_agent_addr[1]
            })
        else:
            unit_operator_agent._role_context.data_dict.update({
                "learning_agent_addr": self.learning_agent_addr[0]
                "learning_agent_id": self.learning_agent_addr[1]
            })
```

something like this..?

maurerle (changes requested):
Sorry, that's also great to not hardcode the learning_agent!
I just meant to provide the output agent twice with a different param, as it is used in a different context (writing learning output instead of other output)

maurerle (approved):
Looks good!
Thank you!

PR #206, opened 2023-09-25
Correct cost calculations and storage operation

adamsjohanna:
This commit adapts the use of the right price forecast for storages, the cost calculations in flexable.py according to flexABLE (minor correction and inclusion of avg_operation_times) and the calculation of total_costs as reward_function


assume/strategies/learning_strategies.py
```diff
@@ -271,25 +271,22 @@ def create_observation(
...
             scaled_price_forecast = (
-                unit.forecaster["price_forecast"].loc[start:].values
-                / scaling_factor_price
+                unit.forecaster["price_EOM"].loc[start:].values / scaling_factor_price
             )
             scaled_price_forecast = np.concatenate(
                 [
                     scaled_price_forecast,
-                    unit.forecaster["price_forecast"].iloc[
+                    unit.forecaster["price_EOM"].iloc[
```
maurerle, line 281: why is this not `price_forecast` anymore? Isn't price_EOM the same..?

maurerle:
This looks good! Is the storage now operating as expected?
Hopefully, we are not breaking any other time assumptions through this :)
Maybe you can run some other examples just to be sure..?

maurerle (approved):
looks good!

PR #208, opened 2023-09-26
Fix outputs function

nick-harder:
-rl params were not written when not in learning
-should be for eval
-error handling was wrong

### DIFF
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -357,7 +357,7 @@ class WriteOutput(Role):
         for query in queries:
             try:
                 df = pd.read_sql(query, self.db)
-            except (OperationalError, ProgrammingError):
+            except Exception as e:
                 continue
 
             dfs.append(df)
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -357,10 +357,13 @@ class WriteOutput(Role):
         for query in queries:
             try:
                 df = pd.read_sql(query, self.db)
-            except OperationalError:
+            except (OperationalError, ProgrammingError):
                 continue
 
             dfs.append(df)
+
+        if not dfs:
+            return
 
         df = pd.concat(dfs)
         df.reset_index()
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -357,8 +357,10 @@ class WriteOutput(Role):
         for query in queries:
             try:
                 df = pd.read_sql(query, self.db)
-            except Exception as e:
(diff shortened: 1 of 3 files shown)


assume/common/units_operator.py
```diff
@@ -395,8 +395,13 @@ def write_learning_to_output(self, start: datetime, marketconfig: MarketConfig):
...
 
-        db_aid = self.context.data_dict.get("learning_output_agent_id")
-        db_addr = self.context.data_dict.get("learning_output_agent_addr")
+        db_aid = self.context.data_dict.get(
+            "learning_output_agent_id"
+        ) or self.context.data_dict.get("output_agent_id")
+        db_addr = self.context.data_dict.get(
+            "learning_output_agent_addr"
+        ) or self.context.data_dict.get("output_agent_id")
+
```
maurerle, line 404: why is this needed?
it must be output_agent_addr at least..?

assume/common/outputs.py
```diff
@@ -357,7 +357,7 @@ async def on_stop(self):
         for query in queries:
             try:
                 df = pd.read_sql(query, self.db)
-            except (OperationalError, ProgrammingError):
+            except Exception as e:
```
maurerle, line 360: We don't want to catch every exception without notice.
You might want to add a second except block with different hadling..?

assume/common/units_operator.py
nick-harder: if it is only output agent, we don't write the RL params at all. It would be still a good idea to write them (for example during evaluations). So this is always written

nick-harder: oh the address, haha, saw it, thanks

assume/common/outputs.py
nick-harder: yeah true

assume/common/units_operator.py
maurerle: But then you should just add the learning_output_agent config on validation runs too - instead of changing this here?

assume/common/outputs.py
```diff
@@ -359,6 +359,8 @@ async def on_stop(self):
                 df = pd.read_sql(query, self.db)
             except (OperationalError, ProgrammingError):
                 continue
+            except Exception as e:
+                raise e
```
maurerle, line 363: This does not do anything - it just catches and raises again.
I think you need this:

```suggestion
            except Exception as e:
                logger.error("could not read query: %s", e)
                continue
```

maurerle (approved):

PR #209, opened 2023-09-26
-adding eval episodes

nick-harder:
-restructing load scenario
-adjusting outputs


nick-harder:
Somehow always when running in RL mode (the RL base example) I get some error but only in the very first episode. I am not sure why and couldn't figure it out. Could you please check it @maurerle?

maurerle (approved):

Issue #133, opened 2023-08-11
Remarks from Remark


maurerle:
I just discussed a few code related things with Eike which are:

- [x]  missing documentation for market and market mechanisms
  -  unclear when which manipulations are necessary, which items have to be sent back
  -  unclear which parameters are needed
  - how one can implement own market mechanism
- [x]  market clearing seems to do more than just clearing the market
  -  one option for the future would be to create MarketClearing as an object instead of a callback, which has the following functions:
```python
class MarketClearing:
  grid: Any # <- we can store things here for additional data like grid
  def clear_market(orderbook, products) -> accepted: Orderbook, rejected: Orderbook, on_hold: Orderbook
    ...

  def calculate_meta(accepted: Orderbook) -> meta_dict:
    ...
```

- [x]  currently traders which did not get any offer, are not getting informed about it
- [x]  currently traders are not informed about if an offer is rejected (which could free up reserved space in a bidding agent)
- [ ]  clearing related things
  - [ ]  clearing can have 4 different cases of intersection, which might not be handled in the way they should be from an economic perspective
  - [x]  it can happen, that some orders are sent back splitted due to the way our market mechanism currently works for `pay_as_clear` and `pay_as_bid`

nick-harder:
@maurerle I believe now all the issues are adressed or?

nick-harder:
can we close this issue @maurerle ?

nick-harder:
as there was no new activity in this issue and the problems have been resolved, this issue is now closed

PR #212, opened 2023-09-27
Adressing some major issues with learning

nick-harder:
-fixed evaluation average reward calculation
-fixed saving learned strategies
-fixes profit and regret calculations
-fixed missing unit_id in rejected orders
-fixed issue when mcp is tensor
-fixed some dashboards
-fixed use of variable cost while inputs use fixed cost -added check if learning is on but no RL strategies are given -query for evarge reward in eval raises an error if not available


assume/common/outputs.py
```diff
@@ -380,6 +384,8 @@ async def on_stop(self):
 
             dfs.append(df)
 
+        # remove all empty dataframes
+        dfs = [df for df in dfs if not df.empty]
```
maurerle, line 388: why should there be an empty dataframe? are you sure this still happens?

assume/markets/clearing_algorithms/simple.py
```diff
@@ -112,20 +112,17 @@ def clear(
...
-                    split_supply_order = supply_order.copy()
-                    split_supply_order["volume"] = diff
                     supply_order["accepted_volume"] = supply_order["volume"] - diff
+
                     # changed supply_order is still part of to_commit and will be added
                     # only volume-diff can be sold for current price
                     gen_vol -= diff
 
-                    # add left over to supply_orders again
-                    supply_orders.insert(0, split_supply_order)
```
maurerle, line 126: We need to fix the market clearing in a different way..
As we are simply adding demand - then add supply until demand is matched -> then add again to the demand, we might have the split situation more than once per clearing.

The leftover will therefore not be traded at afterwards.

maurerle (commented):
We should address the generated issue in the clearing in a proper way to still add the leftover, everytime we have enough demand/supply.

Having the same amount of orders as submitted is also an open request for clearing improvements from #133 

Other than that it looks good!

assume/common/outputs.py
nick-harder: Idk why, but I was getting this FutureWarning again. So I added this just in case to remove the warning

assume/markets/clearing_algorithms/simple.py
nick-harder: This clearing is anyway not suitable for complex orders and other things. So I would day we remove this splitting to save bids for future and have a true simple clearing.

nick-harder: I have reverted the clearing to the initial state as it works fine now anyway

maurerle (approved):

PR #213, opened 2023-09-28
Editing dashboards

nick-harder:
-adjust learning dashboards
-fix simple market causing issues with writing bids
-adjust dashboards for cashflow


adamsjohanna:
I think it would be nice to also have the description for each plot in the learning dashboard. I started to add some, but for the Regret etc. I don't even know how it is defined, so maybe you can continue here @nick-harder

adamsjohanna (approved):

maurerle:
This introduces a regression.
The clearing is now incorrect, as the test (which was also adjusted here) should clear the whole Volume of 500

PR #214, opened 2023-09-29
fix market clearing if more supply is available

maurerle:
tests for verification are added


nick-harder:
These extra bids were causing several problems when writing to the DB as bids were going missing. Does this fix that behavior?

maurerle:
yes this fixes this

nick-harder (approved):

PR #216, opened 2023-09-29
update version to 0.2 - fix dependency

maurerle:
This removes the git dependency and enables the 0.2 release of assume


maurerle:
Should be merged after #214 

Supersedes #215 - which reverts features to comply with current mango

nick-harder (approved):

PR #221, opened 2023-10-22
RL Doku Proposal

kim-mskw:
Made a first suggestion for a learning dokumentation. Added a paragraph for the provided exampels and uploaded the workshop examples as proper examples in the folder.

I copied something together. it definitely needs some work, but I wanted you feedback on the general idea first, before I finish it up. What do you guys think?


maurerle:
I think that this is in general very good and helpful!

I added the new files to the index, so that you can navigate them using: 
`make -C docs html && python -m http.server --directory docs/build/html`

You might want to convert your docs using a Markdown to RST conversion tool.

Other OS's than Windows care about capitalization in file names, so I fixed that too :)

kim-mskw:
I did change the format hopefullz in the right waz. I, however, have problems buidling the docs. The port seems not reachable somehow. Does it work for someone else or did I break something?

Also I add github badges that link the colab tutorials.

kim-mskw:
Sooo, I think I fixed everything :) please review the changes, as we should push the draft before the workshop tomorrow, and ideally we should add the description of the other example files as well.

maurerle (approved):
Looks good!

maurerle:
Btw @kim-mskw after you have run the make command - you can run:
`python -m http.server --directory docs/build/html 8090` <- with any open port (must be above 1024)

Maybe something is already running on port 8000 on your machine :shrug:

PR #222, opened 2023-10-23
Mqtt fixes

maurerle:
This includes fixes to running the MQTT container with docker for a distributed simulation.

For distributed simulations, one can start
`python examples/distributed_world_manager.py` in one terminal and run
`python examples/distributed_world_agent.py` in another one.

A registry of which agents are available (and must be online before simulation can start) and a central yaml config are still needed.
So this is an early fix set, which should not change anything to existing behavior


assume/markets/base_market.py
```diff
@@ -397,7 +397,10 @@ async def clear_market(self, market_products: list[MarketProduct]):
 
         for meta in market_meta:
             logger.debug(
-                f'clearing price for {self.marketconfig.name} is {meta["price"]:.2f}, volume: {meta["demand_volume"]}'
+                "clearing price for %s is %.2f, volume: %f",
+                self.marketconfig.name,
```
nick-harder, line 401: can we please stick to one format for strings? In the whole code we are using f"{}" format. it would be confusing to have it different in some places.

examples/distributed_config.py
nick-harder, line 1: all these filed should be under a folder for this specific example. Otherwise, the whole examples folder is getting very busy. Also, since this would be for now a not very often used feature, having it in a seperate folder would be better. How do you think?

assume/markets/base_market.py
maurerle: I had this view too for quite a while.
But unfortunately, formatting with f-strings is evaluated every time - even if the logger is set to warning (so the price is formatted even if the line is not printed).
Using it with the % notation does only execute the formatting when loglevel is set to debug.

But I would not like to switch to the %s notation on all other places.. :shrug:
So I think we can't really do anything but to use the %s format in logger situations and format strings everywhere else

maurerle:
Weird that `get_sum_reward` did not have a length and I needed to handle that case?
This does not occur on my device though.

nick-harder (approved):

assume/markets/base_market.py
nick-harder: hmm, good to know, let's keep it this way

PR #223, opened 2023-10-27
Add shields badges to readme

maurerle:
New badges can be seen here:

https://github.com/assume-framework/assume/blob/shields-badges/README.md


maurerle:
We should also add SPDX license headers to our source files and register here:
https://api.reuse.software/register
so that we get a REUSE compliant badge like amiris does:
https://gitlab.com/dlr-ve/esy/amiris/amiris

Being REUSE compliant has non-technical benefits - it allows other researchers to be clear about the terms under which this project can be used and contributed to.

nick-harder (approved):

PR #224, opened 2023-10-30
add amiris scenario loader

maurerle:
This allows loading scenarios which were originally written for AMIRIS by DLR


nick-harder (approved):

nick-harder:
look great! really cool functionality! But it would be important to also have a documentation about the current limitation of this scenario loader. Could you please add it @maurerle ?

PR #242, opened 2023-11-06
Add reuse headers to every file

maurerle:
The reuse specification is quite strict.
I don't like the added files in the inputs folder a lot, but that's the way it is.
I see myself in the comment from the reuse page:
> think that .license files clutter the directory, or find it very strange that even insignificant non-code files get licensing headers.

https://reuse.software/faq/#tradition


nick-harder (approved):

nick-harder:
Ich hoffe, das ganze war von REUSE automatisiert, hahaha

maurerle:
Ja es war tatsächlich nur:

```
pipx install reuse
pipx run reuse init
pipx run reuse annotate --license AGPL-3.0-or-later --recursive --copyright "ASSUME Developers" --exclude-year --skip-unrecognised .
pipx run reuse lint
```

Dann merge ich es jetzt :)

PR #247, opened 2023-11-10
Data Request mechanism

maurerle:
This adds a new data_request mechanism which can be used to request the power generation of a unit or the market price of a market.
This is needed for contracts, which are developed in a follow-up PR

* add data_request mechanism to market and units_operator
* do not shadow on_stop and setup methods
* unify usage of start_time and end_time

This allows workflows like:
1. request market_price of last week from market
2. wait for result from market
3. calculate contract result (e. g. hours in which the price is above or below some value) in the contract_market_agent

### DIFF
--- a/assume/common/market_objects.py
+++ b/assume/common/market_objects.py
@@ -36,9 +36,13 @@ class Order(TypedDict):
     :param end_time: the end time of the order
     :type end_time: datetime
     :param volume: the volume of the order
-    :type volume: int
+    :type volume: float
     :param price: the price of the order
-    :type price: int
+    :type price: float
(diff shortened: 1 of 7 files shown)


nick-harder (changes requested):
Everything looks fine, but could you please also add some tests for the new functionality and updated setup() function for the market operator? Just to be sure we have working features on main.

nick-harder:
@maurerle hi Florian, so are no additional tests required for this functionality? I saw a commit with fix tests, but no extra tests. Or are they a part of another commit?

nick-harder (approved):

PR #248, opened 2023-11-10
Contract Market with feed in policy and market premium

maurerle:
This PR implements a feed in policy, which calculates the value a special demand agent pays on top of the market prices as a feed-in-tariff.
The contract is accepted on a monthly base for example, and evaluated on a weekly base.
For this, the generation and market_price of the last week is requested using the data_request mechanism developed in #247 - then the resulting orderbook is created and sent to the contract participants using the already existing `clearing` process.

This works in general, though the results are not correct yet. I think the volume and price is applied for the whole series and the dashboard needs some polishment too.

### DIFF
--- a/assume/markets/clearing_algorithms/contracts.py
+++ b/assume/markets/clearing_algorithms/contracts.py
@@ -27,6 +27,19 @@ log = logging.getLogger(__name__)
 
 
 class PayAsBidContractRole(MarketRole):
+    """
+    This market role handles contracts between agents.
+    Contracts can be long term agreements which are valid for a longer period of time.
+    Each contract has an evaluation frequency in which the actual cashflow of the contract is executed.
+    This can include a calculation depending on the market result and dispatched energy.
+    It can be limited which agent can agree to a offered contract at this market by using an `eligible_lambda` in the contract properties.
+    The available contracts can be configured through the `available_contracts` dictionary.
+
+    Args:
+        marketconfig (MarketConfig): The market configuration.
+        limitation (str): a string for limitations - either being "only_co2emissionless" or "only_renewables"
+
+    """
     required_fields = [
         "sender_id",
         "contract",
@@ -38,11 +51,9 @@ class PayAsBidContractRole(MarketRole):
         self,
         marketconfig: MarketConfig,
         limitation: str = "only_co2emissionless",
-        market_indices: dict[str, str] = {},
     ):
         super().__init__(marketconfig)
         self.limitation = limitation
-        self.market_indices = market_indices
         self.futures = {}
 
     def setup(self):
@@ -92,6 +103,8 @@ class PayAsBidContractRole(MarketRole):
                 ]
             elif self.limitation == "only_renewables":
                 requirement = lambda x: x in ["demand", "wind", "solar", "biomass"]
+            else:
+                log.error(f"unknown limitation {self.limitation}")
             return all(
                 [requirement(info["technology"]) for info in content["information"]]
             )
@@ -162,8 +175,6 @@ class PayAsBidContractRole(MarketRole):
                     if supply_order["price"] <= demand_order[
                         "price"
                     ] and self.check_working(supply_order, demand_order):
-                        # TODO this might match a different demand, if different policies are available
-                        # so we should not reject only if check_working is false but check compatibility beforehand?
                         supply_order["accepted_volume"] = supply_order["volume"]
                         to_commit.append(supply_order)
                         gen_vol += supply_order["volume"]
--- a/assume/markets/clearing_algorithms/contracts.py
+++ b/assume/markets/clearing_algorithms/contracts.py
@@ -1,0 +1,659 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import asyncio
+import logging
+from datetime import datetime, timedelta
+from itertools import groupby
+from operator import itemgetter
+from typing import Callable
+
+import pandas as pd
+from dateutil import rrule as rr
+from dateutil.relativedelta import relativedelta as rd
+
+from assume.common.market_objects import (
+    ClearingMessage,
+    MarketConfig,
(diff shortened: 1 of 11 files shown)


nick-harder:
@maurerle why do you believe the feed-in premium should be handled by an extra demand agent, ant not an additional market party? 
Also, it would be better to add functionality when it is 100% operational and tested. So please same here, check functionality, and add tests for the new market clearing algorithms. Thanks :-)

nick-harder:
@maurerle is this now ready?

maurerle:
I'd still like to add some tests - but I don't think that I find time this week.
So we can merge this :)

nick-harder:
@maurerle there is also no description of the new methods, no docstrings, and any mention of the new clearings in the documentation. It would be helpful to have them. Let's merge when tests and documentation are there, it can wait :-)

assume/markets/clearing_algorithms/contracts.py
```diff
@@ -0,0 +1,659 @@
...
+    MetaDict,
+    Order,
+    Orderbook,
+)
+from assume.markets.base_market import MarketRole
+
+log = logging.getLogger(__name__)
+
+
+class PayAsBidContractRole(MarketRole):
```
nick-harder, line 29: would it be also a good idea to have some description docstrings here? for the automated documentation? or the one in text is sufficient from your point fo view?

assume/markets/clearing_algorithms/contracts.py
```diff
@@ -0,0 +1,659 @@
...
+
+        Args:
+            content (dict): message content with registration message and agent information
+            meta (MetaDict): message meta
+
+        Returns:
+            bool: True if agent fulfills requirements
+        """
+        if self.limitation:
+            if self.limitation == "only_co2emissionless":
```
nick-harder, line 85: maybe we could make these technologies as a part of the config instead of hard coding them at this spot?

assume/markets/clearing_algorithms/contracts.py
```diff
@@ -0,0 +1,659 @@
...
+
+                dem_vol += -demand_order["volume"]
+                to_commit: Orderbook = []
+
+                while supply_orders and gen_vol < dem_vol:
+                    supply_order = supply_orders.pop(0)
+                    if supply_order["price"] <= demand_order[
+                        "price"
+                    ] and self.check_working(supply_order, demand_order):
+                        # TODO this might match a different demand, if different policies are available
```
nick-harder, line 165: should this TODO be adressed directly now or is this a work for the future?

assume/scenario/loader_amiris.py
```diff
@@ -408,7 +523,7 @@ async def load_amiris_async(
     # To use this with amiris run:
     # git clone https://gitlab.com/dlr-ve/esy/amiris/examples.git amiris-examples
     # next to the assume folder
-    scenario = "Simple"  # Germany2019 or Austria2019 or Simple
+    scenario = "Germany2019"  # Germany2019 or Austria2019 or Simple
     base_path = f"../amiris-examples/{scenario}/"
```
nick-harder, line 527: shouldn't this be moved/copied to an example file to showcase how to use this feature?

assume/strategies/extended.py
```diff
@@ -2,8 +2,11 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+import dateutil.rrule as rr
```
nick-harder, line 5: why the name is extended? maybe a more specific name would be better?

examples/world_script_policy.py
```diff
@@ -0,0 +1,139 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import logging
```
nick-harder, line 5: why do we need a seperate file for this? We are using config files and input folders for all other examples and adding them into the dict. What is the beenfit of having a seperate file here? Maybe a tutorial notebook would be more beneficial in the future?

nick-harder (changes requested):
I have left some comments

assume/markets/clearing_algorithms/contracts.py
maurerle: This is actually not needed at all as the orders are grouped by the `market_getter`, which includes grouping by contract. So bids with different contracts can never be matched together.
So this is not an issue at all.
Thanks!

assume/strategies/extended.py
maurerle: Sure - but as it was that way before, I would leave this as is in this PR and change it once we have a proposal?
I don't know a much better name - it just includes some not basic strategies..

examples/world_script_policy.py
maurerle: I like to have some things scripted to not create a lot of csv files additionally.
From my view this makes it easy to see what is going on, when having only a few agents and focusing on the market.
I will create a tutorial for the contracts in the future

assume/markets/clearing_algorithms/contracts.py
maurerle: I wanted to improve this once we have more than 2 possible limitations of market participants.
What do you mean with "part of the config"?

assume/markets/clearing_algorithms/contracts.py
maurerle: I added docstrings as well, thanks

assume/markets/clearing_algorithms/contracts.py
nick-harder: since now we have additional configs for each market, we could list such technologies there

nick-harder (approved):

assume/markets/clearing_algorithms/contracts.py
maurerle: Good point.
But I don't know if this will only stay "a list of allowed technologies" - one could also specifiy to limit this market to agents with a max_power > 10MW or whatever - so I would like to keep this loose for now

PR #251, opened 2023-11-12
add fancy completion using argcomplete

maurerle:
While watching [an interesting talk about tab-completion ](https://media.ccc.de/v/2023-252-woher-weiss-die-shell-was-sie-tun-soll-wenn-ich-tab-drcke- ) I tried to set this up for assume too.

[Screencast from 2023-11-12 14-46-15.webm](https://github.com/assume-framework/assume/assets/25026204/b47dfe5d-d3a0-4837-a8ca-92dc33bd8f55)

This suggests possible database paths and scenarios, as well as parameters for case_study based on the context.
By moving the script to the top folder, assume does not need to be imported if only calling the --help for example.

This should also work on macOS by running:

`pip install -e .`
`eval "$(register-python-argcomplete assume)"`

And also for Powershell on Windows using:

`register-python-argcomplete --shell powershell assume | Out-String | Invoke-Expression` (though I did not test it)

### DIFF
--- a/assume/common/scenario_loader.py
+++ b/assume/common/scenario_loader.py
@@ -254,7 +254,8 @@ async def load_scenario_folder_async(
 
     # load the config file
     path = f"{inputs_path}/{scenario}"
-    config = yaml.safe_load(open(f"{path}/config.yaml", "r"))
+    with open(f"{path}/config.yaml", "r") as f:
+        config = yaml.safe_load(f)
     if not study_case:
         study_case = list(config.keys())[0]
     config = config[study_case]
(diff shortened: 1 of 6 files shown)


assume/common/scenario_loader.py
```diff
@@ -254,7 +254,8 @@ async def load_scenario_folder_async(
 
     # load the config file
     path = f"{inputs_path}/{scenario}"
-    config = yaml.safe_load(open(f"{path}/config.yaml", "r"))
+    with open(f"{path}/config.yaml", "r") as f:
```
nick-harder, line 257: just for my knowledge, what is the difference between these two approaches?

nick-harder (approved):

assume/common/scenario_loader.py
maurerle: when using open() in a single the file handler i opened but never actually closed, so you are relying on the garbage collector for closing it.

By using `with open(...) as f:` you are making sure that the file handler is actually closed directly afterwards.
It is best to interact with file ressources as short as possible and clean it up as soon as possible.

https://stackoverflow.com/a/7409814

PR #252, opened 2023-11-23
Adding docs for base unit classes and power plant

nick-harder:
-added base units class description
-added power plant description
-restructured contents

### DIFF
--- a/assume/units/powerplant.py
+++ b/assume/units/powerplant.py
@@ -144,7 +144,11 @@ class PowerPlant(SupportsMinMax):
         """
         Initialize the marginal cost of the unit.
         """
-        self.marginal_cost = self.calc_simple_marginal_cost()
+
+        if not self.partial_load_eff:
+            self.marginal_cost = self.calc_simple_marginal_cost()
+        else:
+            self.marginal_cost = None
 
     def execute_current_dispatch(
         self,
--- a/docs/source/units/powerplant.rst
+++ b/docs/source/units/powerplant.rst
@@ -1,0 +1,54 @@
+.. SPDX-FileCopyrightText: ASSUME Developers
+..
+.. SPDX-License-Identifier: AGPL-3.0-or-later
+
+Power Plant Class
+==================
+
+Overview
+--------
+The PowerPlant class represents a unit for power generation. It models the behavior of a power plant, including its operational constraints
+such as minimum and maximum power output, ramping rates, minimum up and down times, and minimum operating time. It also returns the
+marginal cost of the unit for different power outputs.
+
+Initialization Parameters
+-------------------------------
+.. csv-table::
+   :class: full-width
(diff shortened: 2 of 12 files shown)


assume/units/powerplant.py
```diff
@@ -144,7 +144,11 @@ def init_marginal_cost(self):
         """
         Initialize the marginal cost of the unit.
         """
-        self.marginal_cost = self.calc_simple_marginal_cost()
+
+        if not self.partial_load_eff:
+            self.marginal_cost = self.calc_simple_marginal_cost()
+        else:
+            self.marginal_cost = None
```
maurerle, line 151: It think it would look better to not negate the logic?

Why should None be returned if partial_load_eff is set?

nick-harder: because if it is set and not None, cal_marginal_cost function returns this value and ignores the partial efficiency.

nick-harder: or we should fix the calc_marginal_cost function, either one is valid I believe

nick-harder: @maurerle so what do you think?

docs/source/units/powerplant.rst
maurerle, line 1: I don't think it is very helpful to duplicate about the same content as here:
https://assume.readthedocs.io/en/latest/assume.units.html#module-assume.units.powerplant

This is generated through automodule from the doc strings in the project.
If we add this twice, things will diverge and we will have old wrong docs over time.

Same for the BaseUnit - I think a small paragraph with a link to the reference above would be better?
I like the powerplant_params table loaded from csv - we should use that for the examples list maybe in the future?
But as far as I can see, the specific powerplant_params are already documented in the automodule above right?

Automodule of course has the benefit, that it always stays in line with the code

maurerle: Maybe we can update the doc strings instead to be more helpful
And use the units.rst to describe the general concept and point to the class descriptions of the powerplant, demand and co for the specific information?

nick-harder: How would you suggest to have it? I would still prefer more text based description, because many people are scared of such code generated documentations and don't read them. Try to look at it as if you are not a computer scinetist, but a person just starting with coding and modeling. 
But in general this is a valid point, maybe we @maurerle and @kim-mskw can meet next week to discuss the styling and make some guidelines for it?

maurerle: Of course.
I think we should take this as an example:
https://pypsa.readthedocs.io/en/latest/import_export.html
Which covers links to the API docs on a lot of places to dig deeper.
And shows how to actually use the framework.

Having a list of all Attributes to initialize objects as given here:
https://pypsa.readthedocs.io/en/latest/components.html
is also helpful, but it does not include methods of the classes (which is good).

PR #255, opened 2023-11-29
changed buffer add function call

kim-mskw:
changed buffer add function call to not only add the first value of the observations 50 times.

I am wondering how it learned anything with these observations in the buffer...


nick-harder (approved):

nick-harder:
oh wow, that is weird error. I believe at some point observation contained past and future observations, so this was required. But at some point we probably shifted to use only one observation since next one could be sampled easily, but this part stayed

PR #257, opened 2023-12-01
Added the minimal example

nick-harder:
-added the first small jupyter example
-added it to the documentation
-improved docstrings for world class
-improved docstrings for forecast class
-adjusted the structure of documentation index
-adjusted structure of automated docs
-small fixes in several other docstings

### DIFF
--- a/assume/common/forecasts.py
+++ b/assume/common/forecasts.py
@@ -10,13 +10,22 @@ import pandas as pd
 
 class Forecaster:
     """
-    A Forecaster can provide timeseries for forecasts which are derived either from existing files, random noise or actual forecast methods.
+    A Forecaster can provide timeseries for forecasts which are derived either from existing files,
+    random noise, or actual forecast methods.
 
-    :param index: the index of the forecasts
-    :type index: pd.Series
+    Args:
+    - index (pd.Series): The index of the forecasts.
 
-    Methods
-    -------
+    Methods:
+    - __getitem__(self, column: str) -> pd.Series: Returns the forecast for a given column.
+    - get_availability(self, unit: str) -> pd.Series: Returns the availability of a given unit.
+    - get_price(self, fuel_type: str) -> pd.Series: Returns the price for a given fuel type or zeros if
+    the type does not exist.
+
+    This class represents a forecaster that provides timeseries for forecasts derived from existing files,
+    random noise, or actual forecast methods. It initializes with the provided index. It includes methods
+    to retrieve forecasts for specific columns, availability of units, and prices of fuel types, returning
+    the corresponding timeseries as pandas Series.
     """
 
     def __init__(self, index: pd.Series):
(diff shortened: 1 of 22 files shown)


assume/common/forecasts.py
```diff
@@ -10,13 +10,22 @@
...
+    A Forecaster can provide timeseries for forecasts which are derived either from existing files,
+    random noise, or actual forecast methods.
+
+    Args:
+    - index (pd.Series): The index of the forecasts.
+
+    Methods:
+    - __getitem__(self, column: str) -> pd.Series: Returns the forecast for a given column.
+    - get_availability(self, unit: str) -> pd.Series: Returns the availability of a given unit.
+    - get_price(self, fuel_type: str) -> pd.Series: Returns the price for a given fuel type or zeros if
```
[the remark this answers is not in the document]
nick-harder: If they are removed here, will they still be visible in hints when coding? So you can see them directly?

examples/examples.py
```diff
@@ -58,8 +58,8 @@
     - local_db: without database and grafana
     - timescale: with database and grafana (note: you need docker installed)
     """
-    data_format = "timescale"  # "local_db" or "timescale"
-    example = "learning_small"
+    data_format = "local_db"  # "local_db" or "timescale"
+    example = "small"
```
maurerle, line 62: Is this change inteded?

assume/common/forecasts.py
```diff
@@ -212,13 +281,33 @@ def calculate_EOM_price_forecast(self):
...
+        The calculation involves the following steps:
+        1. Determines the fuel price based on the fuel type of the power plant.
+        2. Calculates the fuel cost by dividing the fuel price by the efficiency of the power plant.
+        3. Calculates the emissions cost based on the CO2 price and emission factor, adjusted by the
+        efficiency of the power plant.
+        4. Considers any fixed costs specified for the power plant.
+        5. Aggregates the fuel cost, emissions cost, and fixed cost to obtain the marginal cost of the
+        power plant.
+
+        Returns the resulting marginal cost as a float value.
```
maurerle, line 308: If we want, we can change the docstring format from the default sphinx one:
https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#an-example-class-with-docstrings

to Google Doc strings or Numpy doc strings:
https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google
or 
https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html#example-numpy

which adds the napoleon plugin.
But I think we should use the same code style all over assume.

maurerle (approved):
We now have two different doc styles used in Assume.
Feel free to fix this now or in some time later.
I think that the added explanations are very useful and we can bulk convert the existing docs probably too.

I'll just approve without merge now.
I did not view the docs myself, but only looked at the github diff.

examples/examples.py
nick-harder: yes, since we don't expected people to already have docker running when using this small example.

assume/common/forecasts.py
nick-harder: please explain, I don't understand

assume/common/forecasts.py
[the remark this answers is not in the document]
maurerle: I just write:
`object.` then the IDE hints  show which methods are available, which is mostly sufficient for me?

Other than that, I like clicking on the breadcrumbs in the menu to see an outline of my current open file:

image

assume/common/forecasts.py
maurerle: Actually this isn't even google Style docs but something weird inbetween - Probably just a Copilot suggestion?

nick-harder: ok, I get it now. So what I have written, it is basically google form with some extra explanation at the end. We can discuss this next time if we wanna keep it and convert existing stuff to it. I find it personally extremely helpful. Let's keep it for now and talk about it during a development meeting.

PR #258, opened 2023-12-04
Add collab tonotebooks

nick-harder:
-added rl as notebook to the repository
-adjusted config to show collab at the top


kim-mskw:
Love how easy and simple it look :) pretty cool! 

I was wondering if we should give an outlook and link the more complex implementations there as well? 
For example: Here we defined the market, in the advanced example this is done with YAML files. 

Or do you think that would be too confusing?

nick-harder:
@kim-mskw yeah, these more compelx notebook will follow later. I wanted to construct dictsa and save them as csv files to demonstrate how to create a scenario and run it. For now I would like to find a good format for us to follow

kim-mskw:
Yeah I know that it will follow later :) I was just wondering if we should mention or link the more complex option already somehow so that it is easier to draw the conclusion. Could be part of the structure then I guess.

nick-harder:
@kim-mskw do you mean to add it directly in the notebook or what exactly? I don't understand. COuld you please make a suggestion directly? :-)

kim-mskw:
Discussed in person

kim-mskw (approved):

PR #262, opened 2023-12-21
Documentation of Unit example

Manish-Khanra:
Documentation of an electrlyser unit example


docs/source/ASSUME_Architecture.png
maurerle, line 1: There is already https://github.com/assume-framework/assume/blob/main/docs/source/img/architecture.svg
If added - this picture would belong into the same folder, but I would just use the .svg where you need it.

examples/notebooks/02_custom_unit_example.ipynb
```diff
@@ -0,0 +1,494 @@
...
+    "ASSUME is a tool designed for researchers, planners, utilities, and anyone interested in understanding how energy markets work. It's like a user-friendly toolbox, available for free, that you can customize to fit your specific needs.\n",
+    "\n",
+    "In the following figure the architecture of the framework is depicted. It can be roughly devided into two parts. On the left side, we have the market elements, and on the right side, we have the market participants, which we'll call 'units' here. These two parts are connected because the market participants make orders in the markets. Now, there's something special on the right side, shown in yellow. That's where we talk about the framework's ability to learn and adapt. But before we dive into all the technical details of ASSUME, let's first get familiar with some important terms and concepts that are the building blocks of electricity market modeling."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "![Model Architecture](ASSUME\\docs\\source\\ASSUME_Architecture.png)"
```
maurerle, line 35: use svg here..?

maurerle (commented):
Adding the tutorial to the list is needed for it to show up on the docs too:
https://github.com/assume-framework/assume/blob/b3aa9c185e88061cbdfc87c38b3eaaac06b70fbf/docs/source/examples_basic.rst

nick-harder:
@kim-mskw could you please check this commit please? You have a good eye for details and consistency :-)

nick-harder:
@Manish-Khanra the part of adding the new unit type is missing. Also, you need something like %%add_to RLStrategy from tutorial 4 RL to be able to have functions inside classes up and running. Please add it where needed. Also, at the beginning, the part about docker and limitations is repeated twice. Please check

docs/source/examples_basic.rst
maurerle, line 1: This does not look right - it should not undo the renaming

nick-harder (approved):

maurerle (changes requested):
The tutorial in #266  is supposed to be number 5 - the market comparison tutorial is supposed to stay number 6 - I think.

nick-harder:
> The tutorial in #266 is supposed to be number 5 - the market comparison tutorial is supposed to stay number 6 - I think.

@maurerle it was decided to merge unit and strategy, so it is 3, RL is 4 and market is 5

maurerle (approved):
the electrolyzer png works on the docs but is broken on the notebook itself..
This is due to the paths being different.

We could fix this, by loading the picture with python and displaying it - resulting in the images being embedded as an output of a cell.
We could also fix this by adding a new folder /img next to the notebook, which includes this image a second time.
Both is really stupid.

We could have something like:

```python
from IPython.display import Image, display
display(Image('docs/source/img/Electrolyzer.png'))
```

nick-harder:
yes true, I changed the way how it is displayed, is ugly, but this is the only way how it works for docs, but also in collab and locally

maurerle (approved):

PR #264, opened 2024-01-11
Add MASTR based OEDS loader - move scenario loaders to separate folder

maurerle:
This adds the documented OEDS loader which can be used to create scenarios from MASTR for NUTS locations, if an open-energy-data-server is available.

The scenario loaders are moved to a separate `scenario` folder and all begin with `loader_`

Documentation is added accordingly, which reflects the possibilities and a small usage guide of it


assume/__init__.py
```diff
@@ -3,11 +3,6 @@
...
 
 from assume.common import MarketConfig, MarketProduct
-from assume.common.scenario_loader import (
-    load_custom_units,
-    load_scenario_folder,
-    run_learning,
-)
 from assume.world import World
 
 __version__ = "0.0.1"
```
nick-harder, line 8: not relevant to this, but general question, do we also need to update the version here as well?

nick-harder (approved):

assume/__init__.py
maurerle: Actually this version is not needed at all, but is a common quasi-standard:
https://stackoverflow.com/a/459185

The way I manage this in other packages is by using:
`from importlib.metadata import version`
sets the value from the pyproject.toml metadata (which is what we want, without duplicating version information.

I will create PR for this

PR #265, opened 2024-01-11
[tutorial] add basic example of EOM and LTM comparison

maurerle:
This still awaits proper renaming of the examples to tutorials, but can be merged once I tested everything.


examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+   "source": [
+    "# 5. market comparison tutorial\n",
+    "\n",
+    "This tutorial describes a showcase [from the scientific paper \"Market Abstraction of Energy Markets and Policies - Application in an Agent-Based Modeling Toolbox\"](https://link.springer.com/chapter/10.1007/978-3-031-48652-4_10).\n",
+    "\n",
+    "It will describe how one can add a new market to the scenario and create a small comparison study of the results using matplotlib.\n",
+    "\n",
+    "The outline of this tutorial is as follows. We will start with a basic summary of the implementation of reinforcement learning (RL) in ASSUME and its architectrue (1. ASSUME & Learning Basics) . If you need a refresher on RL in general, please visit our readthedocs (https://assume.readthedocs.io/en/latest/). Afterwards, we install ASSUME in this Google Colab (2. Get ASSUME running) and then we dive into the learning_strategies.py file and explain how we need to adjust conventional bidding strategies to incorporate RL (3. Make ASSUME learn).\n",
+    "\n",
+    "**As a whole, this tutorial covers the following codding tasks:**\n",
```
nick-harder, line 17: Isn't this description still from the RL tutorial?

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+   "source": [
+    "Now we can run the first scenario which does only include the energy-only market, we call this \"eom_case\""
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
nick-harder, line 122: I believe it would be beneficial to show here how the config file in example_2c looks like, similar to tutorial 2. Or just by reading and printing the file. Otherwise it is hard to see (just in case someone doesn't figure out to open the downloaded files and check inside)

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+    "from functools import partial\n",
+    "import matplotlib.pyplot as plt\n",
+    "import pandas as pd\n",
+    "from sqlalchemy import create_engine\n",
+    "\n",
+    "from assume import World\n",
+    "from assume.scenario.loader_csv import load_scenario_folder\n",
+    "\n",
+    "# make sure that you have a database server up and running\n",
+    "DB_URI = \"postgresql://assume:assume@localhost:5432/assume\""
```
nick-harder, line 107: I would suggest to have a local db here in the tutorial, since the main use case would be to run it in collab. And to have this as comment below in case it is executed locally with a docker in place

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now we can run the second case which also includes the long term market as can be seen in the config.yaml of it:"
```
nick-harder, line 142: Here also maybe just printing it or showing it as a dict? Easier for users to compare directly with the previous cell

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+    "world.run()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "zMyZhaNM7NRP"
+   },
+   "source": [
+    "## 3. Visualize the results\n",
```
nick-harder, line 171: really nice with the visualization. Could you please also here change it to a local db and have the timescale as comment for local use? We need to make sure, that all these notebooks work when started from collab

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "4JeBorbE6FYr"
+   },
+   "source": [
+    "# 5. market comparison tutorial\n",
+    "\n",
+    "This tutorial describes a showcase [from the scientific paper \"Market Abstraction of Energy Markets and Policies - Application in an Agent-Based Modeling Toolbox\"](https://link.springer.com/chapter/10.1007/978-3-031-48652-4_10).\n",
```
nick-harder, line 11: Ideally I would expect people to take a look at the work, but we cannot expect that from everyone? Could you please provide a small theoretical part here, maybe directcly from the paper, introducing the research question and exactly what markets are investigated here?

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "4JeBorbE6FYr"
+   },
+   "source": [
+    "# 5. market comparison tutorial\n",
```
nick-harder, line 9:
```suggestion
    "# 5. Altering market configuration: Effects of a Long Term Market\n",
```

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Here one can see that the dispatch stays about the same, but a large amount has been traded as a base band on the LTM while the daily variances are traded on the EOM.\n",
+    "\n",
+    "This also brings us to the end of this short tutorial on multiple markets."
```
nick-harder, line 330: here, a short summary of what people learned (in this case how to adjust configs for an extra market), and how to visualize the results. I believe what is also missing here is the part about bidding strategies, how exactly are they condifured for 2 markets. Maybe it makes sense to have a section in the tutorial on that? Just a brief mention of what is the bidding strategy, how it is called, and how exactly is it given via the inputs

examples/notebooks/05_market_comparison.ipynb
kim-mskw: I would agree on that and maybe explain with a couple sentence, that you just have to add the second market config in a descriptive way?

examples/notebooks/05_market_comparison.ipynb
kim-mskw: I would say the necessity for this depends on the aim of tutorial. Similar to the new name that you @nick-harder suggested. Is it to show the efects of long-term markets and hence get a different angle on the paper results or to showcase how markets can be configured? 

If it is the later, I would not include any theory besides a description of the markets and how this description is translated to the market-config.

examples/notebooks/05_market_comparison.ipynb
kim-mskw: I would agree on that. This is a great opportunity to showcase the interplay of different bidding strategies.

examples/notebooks/05_market_comparison.ipynb
nick-harder: yes, that and with what parameters, like the frequency, opening hours and so on. Just to describe what market are we adding exactly

examples/notebooks/05_market_comparison.ipynb
nick-harder: we could have both. What if we asked the following RQ? Here is the questions, and we will investigate these market configs. How do we put this into ASSUME? and thus the tutorial.

examples/notebooks/05_market_comparison.ipynb
nick-harder: @maurerle yes, exactly, this is also pretty cool, and a good contribution of the paper and a nice insight people can get from the tutorial

environment_docs.yaml
```diff
@@ -22,4 +22,4 @@ dependencies:
   - sphinx-book-theme
   - nbsphinx
   - nbsphinx_link
-  - sphinxcontrib-mermaid
+  - git+https://github.com/maurerle/sphinxcontrib-mermaid@maurerle
```
nick-harder, line 25: why do we need to rely on this custom version?

nick-harder (approved):

kim-mskw (approved):

environment_docs.yaml
maurerle: The initialization of mermaid js is currently broken on the upstream version - see here:
https://github.com/assume-framework/assume/issues/272

PR #266, opened 2024-01-12
documentation bidding strategy

paragpatil39:
Documentation of bidding strategy


nick-harder:
@paragpatil39 you should also add the tutorial to the documentation, as suggested by Florian in #262, so it will be visible in the readthedocs page

nick-harder:
merged with #262 tutorial and PR

PR #273, opened 2024-01-19
Dokumentation Updates

kim-mskw:
- add unit operator stuff
- rearranged load_params function in learning role, to be algorithm unspecific and actually work with continue learning
- some fixes in documentation (see commits)


examples/examples.py
```diff
@@ -60,7 +60,7 @@
     - timescale: with database and grafana (note: you need docker installed)
     """
     data_format = "local_db"  # "local_db" or "timescale"
-    example = "small"
+    example = "small_learning_1"
```
nick-harder, line 63: I would't have the learning example as default on the main, since torch is not installed by default

examples/inputs/example_02a/config.yaml
```diff
@@ -12,12 +12,12 @@ base:
   learning_config:
     observation_dimension: 50
     action_dimension: 2
-    continue_learning: False
-    load_model_path: None
+    continue_learning: True
+    load_model_path: C:\Users\tg3533\assume\examples\inputs\learned_strategies\example_02a_base
```
nick-harder, line 16: also having an absolute path for the policies loading is not ideal, please use relative paths

maurerle (approved):

docs/source/bidding_agents.rst
```diff
@@ -1,49 +0,0 @@
...
-Both are part of :doc:`assume.common`.
-
-A Unit includes a Forecaster which gives Data on Forecasts, as well as a BiddingStrategy, which can make use of the information given in the unit, as well as its forecasts to create a valid Orderbook for the given market.
-
-This three Classes give flexbility regarding where the data for forecasts, availability or prices comes from (Forecaster), what Operation Parameters are given (Unit) as well as the actual bidding behavior which can either be rule-based or based on Reinforcement Learning (BiddingStrategy).
-
-
-.. mermaid::
-
-    classDiagram
```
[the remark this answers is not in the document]
kim-mskw: I feel like it adds relatively little value to the unit operator functionalities. I see its benefit though for the units, yes! Question is how we want to document the units. We could reuse this part. I really do not get the intention of the first paragraph though.

maurerle:
I still get `AttributeError: 'RLStrategy' object has no attribute 'actor'` when running:
`assume -s example_02a  -db sqlite:///./examples/local_db/test_mini_rl.db`
Though I fixed some things which seem to have not looked right, merged main and made the tests pass :)

maurerle:
current main does also fail on `load_actor_params` - seems that this issue exists longer..

There is `world.learning_role.compare_and_save_policies()` which somehow does not what seems to be intended (it does not save policies and lacks the retrieval of the compared policies and has that weird `self.rl_eval_rewards` which does not do anything.
I think it would be good to go through this and fix loading and comparing scenarios.

maurerle:
Turns out that `compare_and_save_policies` is a weird method which is never used but should be fixed somehow.
Instead of using this method of the learning_role - the correct function to use instead of `learning_role.save_params` would have been `learning_role.rl_algorithm.save_params`.

Now, everything works again :blush:

kim-mskw:
> Turns out that `compare_and_save_policies` is a weird method which is never used but should be fixed somehow. Instead of using this method of the learning_role - the correct function to use instead of `learning_role.save_params` would have been `learning_role.rl_algorithm.save_params`.
> 
> Now, everything works again 😊

Actually, it does not. If continue_learning=True we should read the actor and critic params, which is currently not done. This is what I tried to fix for the redundant config variable load_model_path. I am on it though.

PR #276, opened 2024-01-24
added an automatic assignment of RL units to one RL unit operator

kim-mskw:
see title, that's it

### DIFF
--- a/assume/world.py
+++ b/assume/world.py
@@ -384,11 +384,12 @@ class World:
                     self.learning_role.rl_strats[id] = bidding_strategies[product_type]
 
                     # if we have learning strategy we need to assign the powerplant to one  unit_operator handling all leanring units
-                    self.logger.warning(
-                        f"Your chosen unit-operator {unit_operator_id} for the learning unit {id} was overwritten, with Operator RL, since al learning units need to be handeled by one unit operator."
-                    )
+                    if unit_operator_id != "Operator-RL":
+                        self.logger.warning(
+                            f"Your chosen unit-operator {unit_operator_id} for the learning unit {id} was overwritten with 'Operator-RL', since all learning units need to be handeled by one unit operator."
+                        )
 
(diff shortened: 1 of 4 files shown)


assume/world.py
```diff
@@ -383,6 +383,13 @@ async def async_add_unit(
                 ):
                     self.learning_role.rl_strats[id] = bidding_strategies[product_type]
 
+                    # if we have learning strategy we need to assign the powerplant to one  unit_operator handling all leanring units
+                    self.logger.warning(
+                        f"Your chosen unit-operator {unit_operator_id} for the learning unit {id} was overwritten, with Operator RL, since al learning units need to be handeled by one unit operator."
```
maurerle, line 388:
```suggestion
                        f"Your chosen unit-operator {unit_operator_id} for the learning unit {id} was overwritten, with Operator RL, since all learning units need to be handeled by one unit operator."
```

examples/examples.py
maurerle, line 1: we still want to keep the non learning as default ;)

assume/world.py
adamsjohanna: Do we need this warning, if the operator is overwritten, even if "Operator RL" was chosen already?

maurerle: Good point - we only need it `if unit_operator_id != "Operator RL"`

While we are at it - I don't like spaces in names a lot and would prefer "Operator-RL" instead if that is fine for you?

examples/examples.py
```diff
@@ -60,7 +60,7 @@
     - timescale: with database and grafana (note: you need docker installed)
     """
     data_format = "local_db"  # "local_db" or "timescale"
-    example = "small"
+    example = "small_learning_1"
```
adamsjohanna, line 63:
```suggestion
    example = "small"
```

kim-mskw: Thanks!

PR #279, opened 2024-01-31
Redispatch v1

nick-harder:
This commit brings redispatch model into ASSUME. Both cost based and market based redispatch are possible. And clearing can be done using bay as bid or pay as clear mechanisms. This module uses pypsa to model the network, check for congestions, and perform redispatch.


assume/common/market_objects.py
```diff
@@ -136,6 +138,8 @@ class MarketConfig:
     price_tick: float | None = None  # steps in which the price can be increased
     supports_get_unmatched: bool = False
     eligible_obligations_lambda: eligible_lambda = lambda x: True
+    network_path: str = " "
+    solver: str = "glpk"
```
maurerle, line 142: I don't like having these here a lot - this does not have a lot todo with how the market is configured concretely - I would suggest to have something like the bidding_params with
`bidding_params = unit_params.get("bidding_params", self.bidding_params)`

so instead of these two add a dict with `market_params` which contains these

assume/markets/clearing_algorithms/network_clear.py
```diff
@@ -0,0 +1,375 @@
...
+
+            redisp_results = pd.DataFrame(self.redisp_network.generators_t.p)
+
+            # the following loop adds positive redispatch to the respective generator and/or reduce negative redispatch from original day ahead generation plan
+            for col in self.network.generators_t.p_set.columns:
+                redisp_results[f"total_{col}"] = (
+                    redisp_results[col]
+                    + redisp_results[f"posredisp_{col}"]
+                    - redisp_results[f"negredisp_{col}"]
+                )
```
maurerle, line 351: You can do this quite surely without a loop

assume/markets/clearing_algorithms/network_clear.py
```diff
@@ -0,0 +1,375 @@
...
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import numpy as np
+import pandapower as pp
+import pandas as pd
+import pypsa
+from shapely.errors import ShapelyDeprecationWarning
+
+warnings.filterwarnings("ignore", category=ShapelyDeprecationWarning)
+warnings.filterwarnings("ignore", category=FutureWarning)
```
maurerle, line 20: Why is this needed? Can't you install a recent version instead?
I don't like filtering warnings a lot

assume/common/market_objects.py
nick-harder: please make a suggestion in code

assume/markets/clearing_algorithms/network_clear.py
nick-harder: sorry, this whole file was just a draft, sorry you read it, redispatch.py is the main file

nick-harder:
@maurerle we have finished the implementation, redispatch.py is now the working version. @rqussous will check again the logic behind the redispatch and network. @maurerle if you could take a look at the integration and meta data, that would be awesome! Thank you all!

nick-harder:
should be merged after #295

maurerle (approved):

PR #280, opened 2024-02-01
Fixed RL evaluations

kim-mskw:
Fixed RL evaluations to use compare_and_save_policies function instead of comparing in 

- enabled the origignal function compare_and_save_policies()
- was basically redundant in code
- also changed the collect initial experience to be overwritten to 1 when it is 0 because otherwise the buffer gives an error, because it has no entries to sample from

### DIFF
--- a/assume/reinforcement_learning/learning_role.py
+++ b/assume/reinforcement_learning/learning_role.py
@@ -5,6 +5,7 @@
 import logging
 from collections import defaultdict
 from datetime import datetime
+from pathlib import Path
 
 import torch as th
 from dateutil import rrule as rr
@@ -54,7 +55,10 @@ class Learning(Role):
         self.training_episodes = learning_config["training_episodes"]
         self.learning_mode = learning_config["learning_mode"]
         self.continue_learning = learning_config["continue_learning"]
-        self.trained_policies_path = learning_config["trained_policies_path"]
+        self.trained_policies_save_path = learning_config["trained_policies_save_path"]
+        self.trained_policies_load_path = learning_config.get(
+            "trained_policies_load_path", self.trained_policies_save_path
+        )
 
         cuda_device = (
             learning_config["device"]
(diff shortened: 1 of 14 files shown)


assume/reinforcement_learning/learning_role.py
```diff
@@ -232,8 +237,14 @@ def compare_and_save_policies(self) -> None:
...
+            Curretnly the best evaluation metrik is still assessed by the development team and preliminary we use the avergae rewards.
         """
-        modes = ["reward", "profit", "regret"]
+        # modes = ["reward", "profit", "regret"]
+        modes = ["reward"]
+
+        # add current rewrad to list of all rewards
+        self.rl_eval_rewards.append(current_avg_reward)
+
         for mode in modes:
```
nick-harder, line 248: do we even need this for loop now if we are using only a single mode?

assume/reinforcement_learning/learning_role.py
```diff
@@ -243,26 +254,27 @@ def compare_and_save_policies(self) -> None:
...
                 and self.rl_eval_regrets[-1] != 0
             ):
-                self.max_eval_regret = self.rl_eval_regrets[-1]
-                dir_name = "lowest_regret"
-                value = self.max_eval_regret
+                # self.max_eval_regret = self.rl_eval_regrets[-1]
+                # dir_name = "lowest_regret"
+                # value = self.max_eval_regret
+
+                return NotImplementedError("Evaluation Strategie is not implemnted yet")
```
nick-harder, line 274:
```suggestion
                return NotImplementedError("Evaluation Strategy is not implemented yet")
```

nick-harder (changes requested):
I didn't fully get it, so are we now saving when each individual reward is max or the average among all agents? what did we decided from out last long discussion on this topic?

assume/reinforcement_learning/learning_role.py
kim-mskw: I inteded to keep the variability for the future but make it as easy as possible and still have the average reward as the target. But Folrian and I made it more generic and still have the avg reward.

kim-mskw:
> get it, so are we now saving when each individual reward is max or the average among all agents? what did we decided from out last long discussion on this topic?

we still did it with the average among all agents since we did not decide on an alternative. With Florians upcoming commit it is programmed in such a way, that we can easily change that in the future though.

PR #281, opened 2024-02-05
Adjust docstrings

adamsjohanna:
Adjust and add docstrings for all modules used in the advanced orders tutorial (in process).

### DIFF
--- a/assume/common/forecasts.py
+++ b/assume/common/forecasts.py
@@ -33,7 +33,6 @@ class Forecaster:
         Returns the forecast for a given column.
 
         Args:
-            self: The instance of Forecaster.
             column (str): The column of the forecast.
 
         Returns:
@@ -48,7 +47,6 @@ class Forecaster:
         Returns the availability of a given unit as a pandas Series based on the provided index.
 
         Args:
-            self: The instance of Forecaster.
             unit (str): The unit.
 
         Returns:
@@ -68,7 +66,6 @@ class Forecaster:
         not exist.
 
         Args:
-            self: The instance of Forecaster.
             fuel_type (str): The fuel type.
 
         Returns:
@@ -116,7 +113,6 @@ class CsvForecaster(Forecaster):
         If the column does not exist in the forecasts, a Series of zeros is returned. If the column contains "availability", a Series of ones is returned.
 
         Args:
-            self: The instance of CsvForecaster.
             column (str): The column of the forecast.
 
         Returns:
@@ -138,7 +134,6 @@ class CsvForecaster(Forecaster):
         a Series, it sets the forecast for the given column. If data is None, no action is taken.
 
         Args:
-            self: The instance of CsvForecaster.
             data (pd.DataFrame | pd.Series | None): The forecast data.
             prefix (str): The prefix of the column.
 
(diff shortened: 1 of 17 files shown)


assume/common/forecasts.py
```diff
@@ -169,6 +174,9 @@ def calc_forecast_if_needed(self):
 
         This method calculates additional forecasts if they do not already exist, including
         "price_EOM" and "residual_load_forecast".
+
+        Args:
+            self: The instance of CsvForecaster.
```
maurerle, line 179: Self should not be part of the args anywhere - as this is implicit for classes

assume/common/forecasts.py
```diff
@@ -15,8 +15,8 @@ class Forecaster:
     to retrieve forecasts for specific columns, availability of units, and prices of fuel types, returning
     the corresponding timeseries as pandas Series.
 
-    Args:
-    - index (pd.Series): The index of the forecasts.
+    Attributes:
+        index (pd.Series): The index of the forecasts.
```
maurerle, line 19: It is correct to have this called Args:

If you look at this example:
https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

the def __init__ is documented in the class doc string with Args

You would use Attributes for Class-Variables which are independent of the actual instance

assume/markets/clearing_algorithms/complex_clearing.py
```diff
@@ -341,22 +360,24 @@ def clear(
...
+    Calculates the surplus of an order given the market clearing prices and results of the market clearing.
 
     Args:
         order (dict): The order
-        market_clearing_prices (dict): The market clearing prices
-        instance (pyomo.ConcreteModel): The solved pyomo model containing the results of the market clearing
-        children (list[dict]): The linked child bids of the given order
+        market_clearing_prices (dict): The market clearing prices.
+        instance (pyomo.ConcreteModel): The solved pyomo model containing the results of the market clearing.
+        children (list[dict]): The linked child bids of the given order.
```
maurerle, line 368: I do not think punctuation is required here, but I don't mind :smile:

assume/common/forecasts.py
adamsjohanna: So it's not double, if it is added in both class definition and init()? Then I will add it for all others too

maurerle: yes - Attributes document the class attributes and args document the arguments to the constructor.

In most cases those are quite similar - as args to the constructor are then set as an attribute, so yes you can add them

maurerle (changes requested):
I just pushed a commit which removes the docs for the self args using the regex search and replace of vscodium:

search for:
```
        Args:
            self:.*
```

replace with:
```
        Args:
```
globally :)

but you can look through it

assume/common/forecasts.py
adamsjohanna: Thanks for the explanation 👍

assume/markets/clearing_algorithms/complex_clearing.py
maurerle: I would rather not have any punctuation in the args sections - as it is not a real sentence anyway :smile: 
But thats probably the least important thing

maurerle (approved):
Looks fine
You could build the docs once to confirm it looks nice if you haven't done that already.

So feel free to merge

PR #283, opened 2024-02-06
add additional clearing and strategy docs

maurerle:
* add additional clearing and strategy docs
* update copyright
* use version from pyproject.toml in docs version


nick-harder:
@maurerle could you please also fix the requirements here so we can make a release?

nick-harder (approved):

PR #289, opened 2024-02-09
New strategies allocation using market name instead of product type

nick-harder:


assume/strategies/flexable.py
```diff
@@ -340,8 +340,8 @@ def calculate_bids(
             min_power[start] = unit.calculate_ramp(
                 op_time, previous_power, min_power[start], current_power
             )
-            bid_quantity = min_power[start] - previous_power
-            if bid_quantity >= 0:
+            bid_quantity = previous_power - min_power[start]
```
adamsjohanna, line 343: Just a reminder: Now that the bid_quantity is positive, we have to be careful with the dispatch if product_type = energy_neg, because we then have to subtract this part. (Not implemented yet)

assume/units/powerplant.py
```diff
@@ -20,32 +20,6 @@ class PowerPlant(SupportsMinMax):
...
-        min_operating_time (float, optional): The minimum duration that the power plant must operate once started, in hours.(Defaults to 0 hours.). Defaults to 0.
-        min_down_time (float, optional): The minimum downtime required after a shutdown before the power plant can be restarted, in hours.(Defaults to 0 hours.). Defaults to 0.
-        downtime_hot_start (int, optional): The downtime required after a hot start before the power plant can be restarted, in hours.(Defaults to 8 hours.). Defaults to 8.
-        downtime_warm_start (int, optional): The downtime required after a warm start before the power plant can be restarted, in hours.( Defaults to 48 hours.). Defaults to 48.
-        heat_extraction (bool, optional): A boolean indicating whether the power plant can extract heat for external purposes.(Defaults to False.). Defaults to False.
-        max_heat_extraction (float, optional): The maximum amount of heat that the power plant can extract for external use, in some suitable unit.(Defaults to 0.). Defaults to 0.
-
-    Args:
```
maurerle, line 48: Why did you delete this?
Args: should be left - not Attributes right?

assume/strategies/flexable.py
maurerle: For me it is unclear why it is is needed to switch to positive here?
Neither the Commit message, nor the PR state this?
Shouldn't this be separated?

nick-harder: yes, true, this is a bug, but the problem is bigger, we need to fix it in a separate commit.

assume/units/powerplant.py
nick-harder: fixed

maurerle (approved):

PR #292, opened 2024-02-12
Fix bidding strategy assignment

nick-harder:
fix several spots where bidding strategy was assigned using the old convention


nick-harder:
@maurerle I am not sure if scenario_loaders amiris and oeds is correct, please double check if it is in line with your convention

assume/strategies/naive_strategies.py
```diff
@@ -76,7 +76,7 @@ def calculate_bids(
         return bids
 
 
-class NaiveDAStrategy(BaseStrategy):
```
maurerle, line 79: I think this is much better renamed to NaiveProfileStrategy? I find it weird to tie the name to the DAM?
But if you think it is better like this, it is fine for me.

Same for the NaiveEOM -> NaiveSingleBidStrategy?

maurerle (approved):

assume/strategies/naive_strategies.py
nick-harder: yeah true, will change it

PR #293, opened 2024-02-12
Remove empty bid as method of bidding strategy

nick-harder:
Removes the clearing of empty bids from unit operator into the bidding strategies themselves. This allows for the user to control this behavior and keep bids with volume of 0 if needed (such as for redispatch market)

### DIFF
--- a/assume/strategies/naive_strategies.py
+++ b/assume/strategies/naive_strategies.py
@@ -73,6 +73,10 @@ class NaiveSingleBidStrategy(BaseStrategy):
             else:
                 op_time = min(op_time, 0) - 1
 
+        bids = self.remove_empty_bids(bids)
+
+        bids = self.remove_empty_bids(bids)
+
         return bids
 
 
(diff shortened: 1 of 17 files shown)


assume/strategies/naive_strategies.py
```diff
@@ -73,6 +73,10 @@ def calculate_bids(
             else:
                 op_time = min(op_time, 0) - 1
 
+        bids = self.remove_empty_bids(bids)
+
+        bids = self.remove_empty_bids(bids)
```
maurerle, line 78: why is this needed twice?

maurerle (approved):
If you are sure about the results of the flexable strategies tests, this is fine.

You could add two small tests which test the new functions independently to check that:
1. does not remove volume !=0
2. removes volume == 0

assume/strategies/naive_strategies.py
nick-harder: oh yeah thats a mistake

PR #294, opened 2024-02-12
Remove hard coded EOM references from code base

nick-harder:
- price_EOM and load_EOM were hardcoded into strategies -> changed to use market_id 
- forecasts relied on demand_EOM units -> now all demand units are considered 
- forecaster checks which units have bidding strategies for this particular market when calculating forecasts 
- renamed name in marketconfig to market_id for better readability and consistency 
- adjusted tutorials and tests to comply with new definitions
- removed unused code from dmas strategies


docs/source/market_config.rst
```diff
@@ -12,7 +12,7 @@ A full list of configurations and explanation is given here.
  ============================= =====================================================
   market config item            description
  ============================= =====================================================
-  name                          string name
+  market_id                          string name
```
maurerle, line 15: indent

assume/common/forecasts.py
```diff
@@ -174,27 +182,38 @@ def calc_forecast_if_needed(self):
...
+            if config["product_type"] != "energy":
+                self.logger.warning(
+                    f"Price forecast could be calculated for {market_id}. It can only be calculated for energy only markets for now"
+                )
+                continue
+
+            if f"price_{market_id}" not in self.forecasts.columns:
+                self.forecasts[
+                    f"price_{market_id}"
+                ] = self.calculate_EOM_price_forecast(market_id=market_id)
```
maurerle, line 211: calculate_market_price_forecast would suit the method name much better now right?

maurerle (approved):
Did you check that the dashboards still show everything correctly?
But I think the market_config.name was never used there right? - They already refer to the market_id everywhere I think?

I still need the DMAS code though - but I can readd it

assume/common/forecasts.py
nick-harder: yes true

Issue #277, opened 2024-01-25
Clarify usage of fixed_cost, variable_cost and relative_margin


maurerle:
As marginal cost is the change in the total cost that arises when the quantity produced is incremented by one unit - it should not include fixed_costs which must be payed every hour.

Variable cost as suggested here covers the case of having an additional cost on top of the price calculation, which is respected in the marginal cost - this could also be seen as runtime cost in €/MW of the unit

Additionally, it would be good to have a relative margin on the price, which should be added after adding the fixed_cost/variable_cost, by multiplying the cost.

We should document that and make sure that this works as expected :)

_Originally posted by @maurerle in https://github.com/assume-framework/assume/pull/269#discussion_r1466133854_

maurerle:
@nick-harder and I discussed to remove the fixed_cost (which is not included in marginal cost) and renamt the variable_cost (which is included in marginal cost) to "extra_cost".

A relative_margin would be added after marginal cost calculation and could be added in the bidding strategy and is therefore not needed.

This simplifies things a lot

PR #295, opened 2024-02-13
Fix missing fixed costs and fix forecaster

nick-harder:
-fix missing fixed cost in marginal cost calculations -> now power plant and storage units consider both variable and fixed cost
- before all calculations were missing fixed costs completely since all input files provide only fixed costs
-fix forecaster keeping old values even if new values are given -> now values are overwritten 
-fix learning_strategy missing cost calculations
-fix tests to account for missing fixed cost in marginal cost


maurerle (commented):
Did you discuss this with @adamsjohanna ?
We need to do some cleanup regarding the usage of fixed costs as discussed in #277 
I think we should get it right, when we are at fixing it

assume/units/powerplant.py
```diff
@@ -76,7 +74,7 @@ def __init__(
         downtime_warm_start: int = 48,  # hours
         heat_extraction: bool = False,
         max_heat_extraction: float = 0,
-        location: tuple[float, float] = (0.0, 0.0),
+        location: Tuple[float, float] = (0.0, 0.0),
```
maurerle, line 77: this change is the same for all supported versions, but I don't mind

assume/units/powerplant.py
```diff
@@ -91,16 +89,30 @@ def __init__(
             **kwargs,
         )
 
+        self.id = id
+        self.unit_operator = unit_operator
+        self.technology = technology
+        self.bidding_strategies = bidding_strategies
+        self.index = index
```
maurerle, line 96: I would like to remove these 5 lines - as this already happens in the call to super() and should not be done twice

maurerle (commented):
I left some remarks on the issue I did not understand :)

assume/strategies/flexable.py
```diff
@@ -345,8 +345,8 @@ def calculate_bids(
             min_power[start] = unit.calculate_ramp(
                 op_time, previous_power, min_power[start], current_power
             )
-            bid_quantity = min_power[start] - previous_power
-            if bid_quantity >= 0:
+            bid_quantity = previous_power - min_power[start]
+            if bid_quantity <= 0:
```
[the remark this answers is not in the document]
nick-harder: This is a bid for negative control reserve. There was just a small bug that was bidding negative value for supply instead of positive values. So the negative CRM results were always empty. This simply reverses the sign, nothing much to it

assume/strategies/learning_strategies.py
```diff
@@ -446,6 +449,7 @@ def calculate_reward(
         unit.outputs["profit"].loc[start:end_excl] += profit
         unit.outputs["reward"].loc[start:end_excl] = reward
         unit.outputs["regret"].loc[start:end_excl] = opportunity_cost
+        unit.outputs["total_costs"].loc[start:end_excl] = costs
```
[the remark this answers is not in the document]
nick-harder: Following how it is everywhere else, total_cost us used in all code

assume/units/powerplant.py
nick-harder: What do you mean by versions?

assume/units/powerplant.py
nick-harder: Oh, this was copilot...sorry, will remove it

maurerle (approved):

PR #302, opened 2024-02-20
Fix datetime warning

maurerle:
This drops the timezone information after conversion and therefore stays compliant with the current behavior.
This is a follow-up to #285

It also includes tests for the helper functions, to make sure that they adhere to the offset correctly.
Running the tests in CI might not be always helpful, as they probably run in UTC and therefore do not have weird issues.

The existing and current behavior is to always convert datetimes into UTC and never store the Timezone with them.

### DIFF
--- a/assume/common/mango_serializer.py
+++ b/assume/common/mango_serializer.py
@@ -2,19 +2,20 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
-import calendar
 import pickle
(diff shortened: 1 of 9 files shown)


nick-harder:
@maurerle I believe the problem is that the version of datetime for python 3.10 doesn't have UTC in it, that was also a problem for me, and why I considered dropping 3.10 version

nick-harder (approved):

PR #303, opened 2024-02-20
Add a tutorial for the advanced order types and documentation for the complex clearing

adamsjohanna:


assume/markets/clearing_algorithms/complex_clearing.py
```diff
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+import cProfile
```
nick-harder, line 5: why is this needed?

assume/strategies/advanced_orders.py
```diff
@@ -166,20 +166,21 @@ def calculate_bids(
...
+                    "only_hours": product_tuples[0][2],
+                    "price": mean_price,
+                    "volume": bid_quantity_block,
+                    "bid_type": "BB",
+                    "min_acceptance_ratio": 1,
+                    "accepted_volume": {product[0]: 0 for product in product_tuples},
+                }
+            )
 
         bids = self.remove_empty_bids(bids)
```
nick-harder, line 185: since there is already a check for volume != 0 this is not needed

examples/examples.py
```diff
@@ -23,8 +23,8 @@
         "study_case": "dam_with_complex_clearing",
     },
     "small_with_BB_and_LB": {
-        "scenario": "example_01d",
-        "study_case": "dam_with_complex_clearing",
+        "scenario": "example_01c",
```
nick-harder, line 26: did you merge c and d? and also I saw you added a new e example, is it also in the dict of available examples now?

examples/examples.py
```diff
@@ -64,7 +64,7 @@
     - local_db: without database and grafana
     - timescale: with database and grafana (note: you need docker installed)
     """
-    data_format = "local_db"  # "local_db" or "timescale"
+    data_format = "timescale"  # "local_db" or "timescale"
```
nick-harder, line 67: I would prefer to keep local db in the exmaple on the main

nick-harder:
@adamsjohanna also the tutorial has to be added to the docs, see docs folder for details

assume/strategies/advanced_orders.py
adamsjohanna: We only check the volume for the block order, so we need this here.

assume/strategies/flexable_storage.py
```diff
@@ -335,15 +335,6 @@ def calculate_bids(
                 )
                 theoretic_SOC += delta_soc
                 previous_power = bid_quantity + current_power
-                time_delta = (end - start) / timedelta(hours=1)
```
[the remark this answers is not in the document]
adamsjohanna: This was just double

nick-harder (approved):

PR #304, opened 2024-02-20
This commit prepares assume to integrate proper nodal pricing

nick-harder:
-delete old examplery nodal pricing
-it will be replaced with pypsa model similar to redispatch 
-provide better description and structure for markets API documentation


maurerle:
I don't really see the reason in removing things without replacing them with better things, while we have a differen WIP PR open which improves this?

nick-harder:
@maurerle I have now added the new nodal clearing similar to redispatch structure, but without all the extra funcitons. it uses the same input data and same example folder. Please take a look

nick-harder:
@maurerle yeah, this one has to be significantly changed to be in line with the new version. I will work on it this week (hopefully) and let you know

assume/strategies/naive_strategies.py
```diff
@@ -319,3 +319,82 @@ def calculate_bids(
...
+            bids.append(
+                {
+                    "start_time": product[0],
+                    "end_time": product[1],
+                    "only_hours": product[2],
+                    "price": marginal_cost,
+                    "volume": volume,
+                    "max_power": unit.max_power if volume > 0 else unit.min_power,
+                    "min_power": min_power[start] if volume > 0 else unit.max_power,
+                    "node": node,
```
maurerle, line 390: if this is the only difference here, I would rather have a check in the NaiveStrategy to add its node if it is required in market_config.additional_fields?

assume/markets/clearing_algorithms/nodal_pricing.py
```diff
@@ -3,283 +3,228 @@
...
+        assert self.payment_mechanism in ["pay_as_bid", "pay_as_clear"]
 
     def setup(self):
         super().setup()
 
     def clear(
-        self, orderbook: Orderbook, market_products: list[MarketProduct]
-    ) -> (Orderbook, Orderbook, list[dict]):
+        self, orderbook: Orderbook, market_products
+    ) -> Tuple[Orderbook, Orderbook, List[dict]]:
```
maurerle, line 96: this should be lowercase for built-in typing

maurerle (commented):
I would like to have closer look at the example when I have time, but so far it looks good

assume/strategies/naive_strategies.py
nick-harder: good point

maurerle (approved):

PR #305, opened 2024-02-20
Add basic grid visualization

maurerle:
This adds a way to visualize a given grid at the end of the Grafana dashboard.
This requires nodes to have a shape included in the input data.

It works by writing the grid structure data to the database at the start of the simulation.

It includes some further fixes too:
* It switches the "node_id" to "node" correctly.
* A check for the needed additional_fields in the marketdesign is enforced when using `redispatch`
* Make nodal_pricing also use the same grid function.
* let loader_csv load the csv files and let the market_mechanism handle pandas Dataframes

What is still needed is to include writing the current load at each bus (and generator?) with the market result.
Then we can extend the query with a join of this data and have nice visualizations of the grid usage at each time.

image


assume/markets/clearing_algorithms/redispatch.py
```diff
@@ -37,16 +38,23 @@ class RedispatchMarketRole(MarketRole):
...
         self.network = pypsa.Network()
         # set snapshots as list from the value marketconfig.producs.count converted to list
         self.network.snapshots = range(marketconfig.market_products[0].count)
-        self.solver = marketconfig.param_dict.get("solver", "glpk")
-        network_path = marketconfig.param_dict.get("network_path")
+        self.network_data = marketconfig.param_dict.get("grid_data")
+        assert self.network_data
+        # set backup marginal cost
+        backup_marginal_cost = marketconfig.param_dict.get("backup_marginal_cost", 10e4)
+        read_pypsa_grid(
```
nick-harder, line 52: I find this a good approach, but only to be used with buses and lines. The redispatch requires specific handling of power plants to create all those upward and downward power plants. This should stay in here to be able to see how exactly redispatch is handled and modeled. The functions need t be corrected

examples/inputs/example_01d/buses.csv
```diff
@@ -1,3 +1,3 @@
-name,v_nom,carrier,x,y
-north,380,AC,9.598,53.5585
-south,380,AC,13.607,51.0769
+name,v_nom,carrier,x,y,shape
```
nick-harder, line 1: I really really doubt users would input directly geopandas shapes in here, We should work with coordinates like x y or lat long and convert it inside code ourselves

nick-harder (changes requested):
very cool funcitonality. But the utils function has to be adjusted since this is only a very specific case for the redispatch, which requires special manipulation of power plants, and it should stay in the market so people can see how this is modeled. Buses and lines can go into utils. Also, having direclly points inside inpout csv files is not the best way, we should work with long lat as coordinates and convert it outselves or have points, but not only the points from geopandas

examples/inputs/example_01d/lines.csv
```diff
@@ -1,2 +1,2 @@
-name,bus0,bus1,s_nom,x,r
-0,north,south,5000,0.01,0.001
+name,bus0,bus1,s_nom,x,r,shape
+0,north,south,5000,0.01,0.001,"LINESTRING (9.598 53.5585, 13.607 51.0769)"
```
maurerle, line 2: So @nick-harder you would rather have lon and lat in buses/powerplants and let assume create an according `LINESTRING` from the bus0 and bus1 connection on the go instead of here?
Sounds good too.

Then we do not need to have this information in every file

nick-harder: yes exacty, only the coordinates in the busses should be enough to construct the whole thing.

nick-harder:
@maurerle another comment is that maybe we need a seperate dashboard for this functionality? Ramiz told me that for him it didn't work very well and was extremeley laggy for large networks with many nodes and lines. Putting it into a seperate we at least insure that the main dashboard is accesable

assume/markets/clearing_algorithms/redispatch.py
```diff
@@ -37,16 +38,25 @@ class RedispatchMarketRole(MarketRole):
...
         super().__init__(marketconfig)
-
         self.network = pypsa.Network()
         # set snapshots as list from the value marketconfig.producs.count converted to list
         self.network.snapshots = range(marketconfig.market_products[0].count)
-        self.solver = marketconfig.param_dict.get("solver", "glpk")
-        network_path = marketconfig.param_dict.get("network_path")
+        assert self.grid_data
+        # set backup marginal cost
+        backup_marginal_cost = marketconfig.param_dict.get("backup_marginal_cost", 10e4)
```
nick-harder, line 50: sometimes we use self.market_config, sometimes just the market config. Should we use one version only?

assume/markets/grid_utils.py
```diff
@@ -0,0 +1,131 @@
+# SPDX-FileCopyrightText: ASSUME Developers
```
nick-harder, line 1: should this file be in markets? maybe common is a better place for it?

assume/markets/grid_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+
+    def add_lines(network: pypsa.Network, lines: pd.DataFrame) -> None:
+        network.madd(
+            "Line",
+            names=lines.index,
+            **lines,
+        )
+
+    def add_generators(
```
nick-harder, line 90: this function is also specific for redispatch, since generators are loaded as Load. This won't be the case for normal networks

nick-harder (changes requested):
Some minor things replated to redispatch specific functions and location of the grid utils need to be adressed

assume/markets/grid_utils.py
maurerle: I hope I fixed this as expected.

nick-harder:
@maurerle I am getting the following error when trying to test it. I am tetsing it using local_db

`ERROR:assume.common.outputs:tried writing grid data to non postGIS database
Traceback (most recent call last):
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1969, in _exec_single_context
    self.dialect.do_execute(
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 922, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: near "EXTENSION": syntax error

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/nick/GitHub/assume_dev/assume/common/outputs.py", line 275, in store_grid
    db.execute(text("CREATE EXTENSION IF NOT EXISTS postgis;"))
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1416, in execute
    return meth(
           ^^^^^
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/sql/elements.py", line 517, in _execute_on_connection
    return connection._execute_clauseelement(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1639, in _execute_clauseelement
    ret = self._execute_context(
          ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1848, in _execute_context
    return self._exec_single_context(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1988, in _exec_single_context
    self._handle_dbapi_exception(
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 2344, in _handle_dbapi_exception
    raise sqlalchemy_exception.with_traceback(exc_info[2]) from e
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1969, in _exec_single_context
    self.dialect.do_execute(
  File "/Users/nick/miniconda3/envs/assume-framework/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 922, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "EXTENSION": syntax error
[SQL: CREATE EXTENSION IF NOT EXISTS postgis;]
(Background on this error at: https://sqlalche.me/e/20/e3q8)`

compose.yml
```diff
@@ -5,7 +5,9 @@
 version: "3.9"
 services:
   assume_db:
-    image: timescale/timescaledb:latest-pg15
+    image: timescale/timescaledb-ha:pg15-all-oss
```
nick-harder, line 8: this version is not working on my macbook. I am getting the following error. Can we make this postgis as optional dependency?
 ⠧ assume_db Pulling                                                                                                                                                                                                            
no matching manifest for linux/arm64/v8 in the manifest list entries

nick-harder (changes requested):
everyhting works now! There is only one issue with docker compose as it doesn't work on my macbook, and probably also on other arm powered devices

compose.yml
maurerle: I think we should switch to `timescaledev/timescaledb-ha:pg15-all-oss` which already has arm support.
See https://github.com/timescale/timescaledb-docker-ha/issues/259

Once the official timescale org image has it too, we can switch back.
Otherwise we could remove postgis from it, but I doubt that it will be pleasent to switch to manually afterwards :shrug: 

So I will switch to the dev image.

maurerle:
@nick-harder 
the `db query error: pq: relation "lines_geo" does not exist` should not happen if you have a database with postgis support - if not - you can not use the grid visualization feature

nick-harder:
@maurerle with the new dev distribution I am getting an error again. I have deleted all containers, and DB files, and ran docker compose from scratch. Here is the error I am getting:

sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) can't adapt type 'Point'
[SQL: INSERT INTO buses_geo (name, v_nom, carrier, x, y, wkt_srid_4326, simulation, geometry) VALUES (%(name__0)s, %(v_nom__0)s, %(carrier__0)s, %(x__0)s, %(y__0)s, %(wkt_srid_4326__0)s, %(simulation__0)s, %(geometry__0)s), (%(name__1)s, %(v_nom__1)s, %(carrier__1)s, %(x__1)s, %(y__1)s, %(wkt_srid_4326__1)s, %(simulation__1)s, %(geometry__1)s)]
[parameters: {'name__0': 'north', 'simulation__0': 'example_01d_base', 'geometry__0': <POINT (9.598 53.558)>, 'carrier__0': 'AC', 'y__0': 53.5585, 'wkt_srid_4326__0': 'SRID=4326;POINT (9.598 53.5585)', 'v_nom__0': 380.0, 'x__0': 9.598, 'name__1': 'south', 'simulation__1': 'example_01d_base', 'geometry__1': <POINT (13.607 51.077)>, 'carrier__1': 'AC', 'y__1': 51.0769, 'wkt_srid_4326__1': 'SRID=4326;POINT (13.607 51.0769)', 'v_nom__1': 380.0, 'x__1': 13.607}]
(Background on this error at: https://sqlalche.me/e/20/f405)

maurerle:
@nick-harder  I did not get this running this container.
If you are sure that you have this error using the timescaledb-dev image, it is an upstream issue and will be fixed magically in the future..?

nick-harder:
@maurerle I found the mistake, tha package geoalchemy2 was missing, but I saw this error only very deep inside the error message. Should we add it as a requirement for the network distribution?

nick-harder (approved):

maurerle:
@nick-harder I thought to have it written to fail gracefully.
I added a dumb geoalchemy2 import so that it runs in the import error and falls back to normal `to_sql` - which should work too

PR #311, opened 2024-02-28
Pypsa loader

maurerle:
This adds a new loader for arbitrary pypsa networks.

Examples are given from the pypsa examples as:
* `ac_dc_meshed`
* `scigrid_de`
* `storage_hvdc`

These only contain a few hours of simulations, as the corresponding snapshots are not that large.
Besides of the `loads` and `generators`, the buses are loaded and used for the grid visulaization and calculation, if a redispatch or grid related market is used.

Documentation of the pypsa loader has been added too.

image

This relies on the commits from #305  and has to be merged afterwards therefore


assume/scenario/loader_pypsa.py
```diff
@@ -0,0 +1,232 @@
...
+        study_case (str): study case name
+        network (str): database uri to connect to the OEDS
+        marketdesign (list[MarketConfig]): description of the market design which will be used with the scenario
+    """
+    index = network.snapshots
+    index.freq = "h"
+    start = index[0]
+    end = index[-1]
+    sim_id = f"{scenario}_{study_case}"
+    print(f"loading scenario {sim_id}")
```
nick-harder, line 40: logger?

assume/scenario/loader_pypsa.py
```diff
@@ -0,0 +1,232 @@
...
+
+    await world.setup(
+        start=start,
+        end=end,
+        save_frequency_hours=4,
```
nick-harder, line 45: why is this hardcoded?

assume/scenario/loader_pypsa.py
```diff
@@ -0,0 +1,232 @@
...
+    marketdesign: list[MarketConfig],
+):
+    """
+    This initializes a scenario from the given pypsa grid.
+
+    Args:
+        world (World): the world to add this scenario to
+        scenario (str): scenario name
+        study_case (str): study case name
+        network (str): database uri to connect to the OEDS
```
nick-harder, line 32: I believe this is wrong description, please check all docstrings

assume/scenario/loader_pypsa.py
```diff
@@ -0,0 +1,232 @@
...
+        "loads": network.loads,
+    }
+    for market_config in marketdesign:
+        market_config.param_dict["grid_data"] = grid_data
+        world.add_market(mo_id, market_config)
+
+    # naive_eom
+    default_strategy = {mc.market_id: "naive_redispatch" for mc in marketdesign}
+
+    bidding_strategies = {
```
nick-harder, line 66: I am against this approach. we should allow loading from folder, where a network is placed, but also other files can be placed, such as power plants or demand units, and their bidding strategies and other things specified. The current implementation is rather a prototype one

assume/scenario/loader_pypsa.py
```diff
@@ -0,0 +1,232 @@
...
+from assume import World
+from assume.common.forecasts import NaiveForecast
+from assume.common.market_objects import MarketConfig, MarketProduct
+
+
+async def load_pypsa_async(
+    world: World,
+    scenario: str,
+    study_case: str,
+    network: pypsa.Network,
```
nick-harder, line 22: why is network passed? should we use the same structure as other loaders, where we specify the path, and network and other things are loaded from there. I don't find this approach in line with the rest of the framework

assume/scenario/loader_pypsa.py
```diff
@@ -0,0 +1,232 @@
...
+        "gas": default_strategy,
+        "biomass": default_strategy,
+        "nuclear": default_strategy,
+        "wind": default_strategy,
+        "solar": default_strategy,
+        "demand": default_strategy,
+    }
+
+    world.add_unit_operator("powerplant_operator")
+    for _, generator in network.generators.iterrows():
```
nick-harder, line 79: we should allow the use of external csv files as extra inputs as well. would it work?

assume/scenario/loader_pypsa.py
```diff
@@ -0,0 +1,232 @@
...
+from assume.common.forecasts import NaiveForecast
+from assume.common.market_objects import MarketConfig, MarketProduct
+
+
+async def load_pypsa_async(
+    world: World,
+    scenario: str,
+    study_case: str,
+    network: pypsa.Network,
+    marketdesign: list[MarketConfig],
```
nick-harder, line 23: we should also keep this in line with other loader and load the market from the config file and not pass it manually

nick-harder (changes requested):
thanks for such a nice commit! look good, but it is rather in a prototype status right now since many functions are not in line with how the rest of the framework operates. I have left some comments. Please let me know what do you think about that

assume/scenario/loader_pypsa.py
maurerle: The idead is to have a network loaded from somewhere.
Whatever is available here:
https://pypsa.readthedocs.io/en/latest/import_export.html
* HDF5
* PyPower
* Pandapower
* netCDF

If you want to load a pypsa Network from CSV you can use this:
https://pypsa.readthedocs.io/en/latest/api_reference.html#pypsa.Network.import_from_csv_folder
and put the loaded network into the `pypsa_loader`

assume/scenario/loader_pypsa.py
maurerle: As said in the comment above, you can add your extra CSV inputs to the pypsa network beforehand and have them run in the simulation?
Wouldn't this be enought?

nick-harder: my main point is to be able to define bidding startegies for all units, without this functionality I don't see any reason why have such a loader. Network plus market is awesome combination, but markets and bidding strategies need to be configurable, if we want to do it right.

maurerle: I just addressed this in f5a2bd18393f07b4fd9fa397c147fa23b7057864
So you can now configure your market design and the used bidding strategies individually.

nick-harder (approved):

PR #313, opened 2024-02-29
Add Code of Conduct

maurerle:
I think we once talked about having a Code of Conduct and I just came across the one from https://www.contributor-covenant.org which I found suitable, so we might want to have this too, as it does not hurt.

I don't think we need a CoC now, but it does not hurt to have one and is easier to add at this stage.

I found it relatively unclear who the community leaders are in this CoC, somehow all current members are also community leaders? But I think that will be more clear once a solid userbase is established.

What do you think about this?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,138 @@
+<!--
+SPDX-FileCopyrightText: ASSUME Developers
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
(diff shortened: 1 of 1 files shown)


nick-harder:
@maurerle I find it a nice idea to have it already from the beginning, the content is also fine, nothing too specific, but still a good base

nick-harder:
@kim-mskw @Manish-Khanra @adamsjohanna @paragpatil39 it would be great to hear your opinion here as well!

adamsjohanna:
I also think it's a good thing to have.
But maybe we should also have a responsible person in our group to check this from time to time. Or make it a fixed point of agenda each second meeting or so. For now, it's just a thing of 5 seconds, but to have this and not use it in any way does not make much sense to me.

kim-mskw:
> I also think it's a good thing to have. But maybe we should also have a responsible person in our group to check this from time to time. Or make it a fixed point of agenda each second meeting or so. For now, it's just a thing of 5 seconds, but to have this and not use it in any way does not make much sense to me.

I would double down on this point. I like the idea and there are even implications for misbehaving and those need to be reinforced in some way. Maybe we can have an anonymous way to submit concerns about inappropiate behavior that we then can discuss in the development meeting if something arises?

nick-harder:
@kim-mskw @maurerle @Manish-Khanra as discussed today, I have added our emails to the enforcement part. Please take a look. Please also react to this message if everything else is fine, so we will merge this commit

nick-harder:
@Manish-Khanra what would be your opinion? Please respond so we can close the issue

Manish-Khanra:
As per our discussions in the previous development meeting, we should
proceed with the CoC clause proposed by Florian as an initial step, with a
commitment to regularly revisit and refine it as necessary.

PR #319, opened 2024-03-12
use ISO-8601 date format for direct yaml parsing

maurerle:


nick-harder:
@maurerle depending on users not to forget T and followi this exact format is somehow not very intuitive. Is there some specific reason for this?

maurerle:
Mostly just me wondering, why the yaml does not read this into datetime directly ^^
So we can keep the current behavior in favor of usability.

PR #323, opened 2024-03-18
Interoperability tutorial

maurerle:
This adds the interoperability tutorial used for my workshop.

I had to adjust some code to fix values which are declared in pypsa as well as as keyword arguments.
This seems to break if not done right.
It looks a little like an ugly fix. Though I did not came up with something better yet.

This also fixes the first hour in AMIRIS and the Warning "PerformanceWarning highly fragmented pandas DataFrame" when running example_02.

Furthermore, it is now possible to set a refresh_interval of 5s for continuous updates.


assume/common/grid_utils.py
```diff
@@ -27,16 +27,19 @@ def add_generators(
         index=network.snapshots,
         columns=generators.index,
     )
-
+    gen_c = generators.copy()
```
nick-harder, line 30: this gen_c is created but never used? Shouldn't it be passed as **gen_c instread of **generators?

assume/common/grid_utils.py
```diff
@@ -235,22 +245,15 @@ def read_pypsa_grid(
...
-            "Line",
-            names=lines.index,
-            **lines,
-        )
+        network.import_components_from_dataframe(lines, "Line")
 
     # setup the network
     add_buses(network, grid_dict["buses"])
     add_lines(network, grid_dict["lines"])
+    add_loads(network, grid_dict["loads"])
```
nick-harder, line 256: this shouldn't be here since we define different loads for different types of network. we have loads for redispatch and for nodal and these use different functions

nick-harder (approved):

PR #324, opened 2024-03-20
fix bugs in tutorial 6

adamsjohanna:
When running in Colab, the Tutorial 6 returns Warnings when running the simulation with flexableEOM strategy:
"WARNING:pyomo.repn.plugins.lp_writer:EXPORT Suffix 'dual' found on 1 block:    dual LP writer cannot export suffixes to LP files.  Skipping."

This is not fixed yet! Does anyone know how to fix this? I run pip install pyomo in the beginning of the notebook. It does not help to install it again...

Also the visualization did not return the wanted output. This is fixed by this PR.

### DIFF
--- a/examples/notebooks/06_advanced_orders_example.ipynb
+++ b/examples/notebooks/06_advanced_orders_example.ipynb
@@ -136,7 +136,19 @@
    },
    "outputs": [],
    "source": [
-    "!pip install assume-framework\n",
+    "!pip install assume-framework"
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


nick-harder:
@adamsjohanna I believe this is a pure collab problem and the glpk solver that is used there. I didn't figure out a way to surpress this warning. But it is fine, everything is still working I believe, or?

adamsjohanna:
Yes, with the few fixes for the visualization, it is running :-)

PR #331, opened 2024-03-26
add assume_cli as a package using assume

maurerle:
It is needed to be an extra package so that the completion is fast when using `assume -h` for example.

Without this, it tries to do `from cli import cli`, while cli is not an installed package, just a script lying around. 
This makes it possible to run scenarios in other folders than the cloned one too.

For examples using `assume -i . -s my_example -c tiny` works, if I created one myself.

This also correctly sets the -l LOGLEVEL switch to set the same level output on the stdout output as in the log_file

### DIFF
--- a/assume/world.py
+++ b/assume/world.py
@@ -89,7 +89,7 @@ class World:
         addr: Union[tuple[str, int], str] = "world",
         database_uri: str = "",
         export_csv_path: str = "",
-        log_level: str = "WARNING",
+        log_level: str = "INFO",
         distributed_role: Optional[bool] = None,
     ) -> None:
         logging.getLogger("assume").setLevel(log_level)
--- a/assume/world.py
+++ b/assume/world.py
@@ -37,7 +37,6 @@ from assume.units import BaseUnit, Demand, PowerPlant, Storage
 
 file_handler = logging.FileHandler(filename="assume.log", mode="w+")
 stdout_handler = logging.StreamHandler(stream=sys.stdout)
-stdout_handler.setLevel(logging.WARNING)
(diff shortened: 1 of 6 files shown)


assume/world.py
```diff
@@ -89,7 +89,7 @@ def __init__(
         addr: Union[tuple[str, int], str] = "world",
         database_uri: str = "",
         export_csv_path: str = "",
-        log_level: str = "INFO",
+        log_level: str = "WARNING",
```
nick-harder, line 92: why do we want to log only warning? seeing the progress of loading and others is pretty helpful. Or is this for a special case only?

assume_cli/cli.py
```diff
@@ -101,7 +98,7 @@ def cli(args=None):
         "-l",
         "--loglevel",
         help="logging level used for file log",
-        default="INFO",
+        default="WARNING",
```
nick-harder, line 101: same here, why only warnings?

nick-harder (changes requested):
why do we switch from logging info to warning?

assume/world.py
```diff
@@ -36,7 +37,6 @@
 
 file_handler = logging.FileHandler(filename="assume.log", mode="w+")
 stdout_handler = logging.StreamHandler(stream=sys.stdout)
-stdout_handler.setLevel(logging.WARNING)
```
maurerle, line 39: Before, we always only warnings in stdout as well.
And LogLevel did only set the output level in `assume.log`

We can change this to show INFO if you want to :)

nick-harder (approved):

PR #334, opened 2024-03-26
add python 3.12 to tests, update docs to use python 3.12 in conda

maurerle:
I did not find any issues using python 3.12, lets see if the pipeline finds any.

I also updated the Dockerfile to use the newer smaller image


maurerle:
Will be possible with the release of torch 2.4 which officially supports python 3.12

nick-harder:
@maurerle torch 2.4.1 has been released and the support for 3.12 has been added. we can now go on with this PR. Should something extra be added?

nick-harder (approved):

PR #335, opened 2024-04-10
Market coupling

nick-harder:


assume/strategies/learning_advanced_orders.py
```diff
@@ -196,6 +196,7 @@ def calculate_bids(
                         "volume": bid_quantity_inflex,
                         "bid_type": "SB",
                         "bid_id": f"{unit.id}_SB_{len(bids) + 1}",
+                        "node": unit.node,
```
kim-mskw, line 199: Is there a particular reason, why you did not add the node to the normal leanring?

kim-mskw:
I thought about this more and have the following claim: with a DC OPF in pypsa the differences between the nodal pricing clearing and this should be neglectable, if we do not have strategic behavior of units
- the DC OPF does not do a lot more than the constraints you added
- I mean we still bid out marginal costs and the available capacity, which equals the not self dispatching implemented now
- I will check that once my student is ready with their case study

nick-harder:
@kim-mskw I discussed it further in my team and the conclusion was, that in order to replicate the exact behavior with pypsa we need to define links instead of lines, and also not just a link, but a link in each direction, so 2 links per line. In that case the results will be exactly the same. When using our current formulation due to voltage calculations, which pypsa does, the results slightly differ at some points. We have tested it yesterday and so the result. You can test it also in this branch by running the nodal example and the zonal example, and comparing their price. So I am now 95% sure this is the correct way to go to implement zonal market coupling. Also, this gives us more flexibility and better value in the long term, not making us dependedent on pypsa. Pypsa models can be still used to create and load szenarios, just as a market clearing this new market cleaering needs to be used.

assume/strategies/learning_advanced_orders.py
nick-harder: missed it, thanks

compose.yml
maurerle, line 1: This should not be part of this PR

examples/examples.py
maurerle, line 1: You probably do not want to change the default here again ;)

maurerle (commented):
I left two comments, otherwise this can be merged.

kim-mskw (commented):
Okay I see, thanks for the thourough discussion :) and the nice adition to Assume :)

assume/markets/clearing_algorithms/complex_clearing.py
kim-mskw, line 1: Looks good :)

kim-mskw (approved):

PR #338, opened 2024-04-12
Fix handling of rejected bids

nick-harder:
1. fix accepted_volume and accepted price being removed if already defined by the market
2. remove logging noise from pypsa scripts
3. remove unrequired clean up of all_orders from newer market clearing algorithms

### DIFF
--- a/assume/markets/base_market.py
+++ b/assume/markets/base_market.py
@@ -493,19 +493,23 @@ class MarketRole(MarketMechanism, Role):
             rejected_orderbook,
             market_meta,
         ) = self.clear(self.all_orders, market_products)
+
         self.all_orders = []
+
         for order in rejected_orderbook:
-            if isinstance(order["volume"], dict):
-                order["accepted_volume"] = {
-                    start: 0.0 for start in order["volume"].keys()
-                }
-                order["accepted_price"] = {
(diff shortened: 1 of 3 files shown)


assume/markets/base_market.py
```diff
@@ -493,19 +493,23 @@ async def clear_market(self, market_products: list[MarketProduct]):
...
-                    start: 0.0 for start in order["volume"].keys()
-                }
-                order["accepted_price"] = {
-                    start: market_meta[i]["price"]
-                    for i, start in enumerate(order["volume"].keys())
-                }
-            else:
-                order["accepted_volume"] = 0.0
-                order["accepted_price"] = market_meta[0]["price"]
+            if "accepted_volume" not in order and "accepted_price" not in order:
```
maurerle, line 500: why is that? if it is rejected it should be 0 or not given?
When would this lead to problems?

nick-harder: this can lead to problems when markt handles the accepted volume and accepted price by itself with different values, for example pay as bid and not as clear. So this checks if the market handled it by itself or not, and if it did, it doesn't change anything.

maurerle (approved):

PR #341, opened 2024-04-19
Improve output metrics and do not wait to write output in simulation

maurerle:
This makes it possible to define additional output queries which are calculated as KPIs at the end of the simulation.

Additionally, we do not wait for the writing of the data in the output container in our simulation anymore.
This speeds up the simulation a little bit if the latest mango-agents-assume version is installed


nick-harder (changes requested):
if this works only with the lates version of mango, should we add a requirement tor the version?

maurerle:
Its not a hard requirement and the latest version is installed by default.
Old environments will just not have this until someone executes `pip install -U  mango-agents-assume` or reinstalls in a fresh environment.

So due to backward compatibility I would not bump the dependency :shrug: but good that we thought about it

nick-harder (approved):

PR #344, opened 2024-04-26
Tests and classes of Dst components, steel plant, flexibility script

Manish-Khanra:


assume/common/forecasts.py
```diff
@@ -178,6 +195,24 @@ def set_forecast(self, data: pd.DataFrame | pd.Series | None, prefix=""):
...
+    def set_operation_states(self, unit: str, data: pd.DataFrame):
+        """
+        Sets the operational state data for a given unit.
+
+        Args:
+            unit (str): The unit for which the operational state data is provided.
+            data (pd.DataFrame): The operational state data.
+        """
+        # Check if the unit already exists in the operation_states dictionary
+        if unit in self.operation_states:
```
maurerle, line 207:
```suggestion
        if unit in self.operation_states.keys():
```
I would prefer to be explicit here

assume/common/forecasts.py
```diff
@@ -389,6 +424,34 @@ def save_forecasts(self, path):
...
+        if os.path.exists(csv_file):
+            # Load existing CSV file
+            existing_df = pd.read_csv(csv_file, index_col=0)
+            # Update existing DataFrame with new data
+            existing_df.update(concatenated_data)
+            # Update the DataFrame
+            concatenated_data = existing_df
+
+        # Save the concatenated data to a CSV file
+        concatenated_data.to_csv(csv_file)
```
maurerle, line 453: This variable `concatenated_data` is not declared anywhere?

Manish-Khanra:
@nick-harder Could you plese run the example_04_base_case_steel_plant_2019 as check whats the issue? I am getting the below error with our new update:
RuntimeWarning: invalid value encountered in cast new_result = trans(result).astype(dtype)

nick-harder:
> @nick-harder Hi Nick, Could you please run the example_04_base_case_steel_plant_2019, and check whats the issue? I am getting the below error with our new update: RuntimeWarning: invalid value encountered in cast new_result = trans(result).astype(dtype)

@Manish-Khanra works fine for me! I would suggest to complete delete the folder local_db and try again. this error looks like it is cming from writing data at some point

Manish-Khanra:
@nick-harder also tried deleting the folder, still getting the same error.

nick-harder:
@Manish-Khanra now I am getting the following error:
  File "/Users/nick/GitHub/assume/assume/units/steel_plant.py", line 414, in calculate_marginal_cost
    +value(self.components["electrolyser"].b.electrolyser_operating_cost[t])
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyomo/core/expr/numvalue.pyx", line 229, in pyomo.core.expr.numvalue.value
  File "pyomo/core/expr/numvalue.pyx", line 215, in pyomo.core.expr.numvalue.value
ValueError: No value for uninitialized NumericValue object electrolyser_operating_cost[0]

could you please take a look at it and let me know when it's fixed. I will look into your other issue further afterwards. thanks!

Manish-Khanra:
> @Manish-Khanra now I am getting the following error: File "/Users/nick/GitHub/assume/assume/units/steel_plant.py", line 414, in calculate_marginal_cost +value(self.components["electrolyser"].b.electrolyser_operating_cost[t]) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "pyomo/core/expr/numvalue.pyx", line 229, in pyomo.core.expr.numvalue.value File "pyomo/core/expr/numvalue.pyx", line 215, in pyomo.core.expr.numvalue.value ValueError: No value for uninitialized NumericValue object electrolyser_operating_cost[0]
> 
> could you please take a look at it and let me know when it's fixed. I will look into your other issue further afterwards. thanks!

This error stems from the same issue. The previous error message occured when I have commented out marginal cost function. This error message is when I have again included the marginal cost function. I suspect this error is related to the sequencing of the pyomo components while creating an instance of a model called 'model_opt'. Could figure out the solution yet

Manish-Khanra:
@nick-harder I have pushed the last changes. Now the branch is ready for the final review. Once reviewed we can finally merge it with the main branch.

tests/test_dst_components.py
```diff
@@ -0,0 +1,351 @@
...
+        ramp_down=20,
+        min_operating_time=1,
+        min_down_time=1,
+        efficiency=0.8,
+        fuel_type="Electricity",
+        index=pd.date_range("2022-01-01", periods=4, freq="h"),
+    )
+
+
+def test_init_function(electrolyser_unit):
```
nick-harder, line 36: This file has a ton of functions with the same name, this won't work and will always raise an error. Please fix

assume/strategies/dsm_load_shift.py
```diff
@@ -0,0 +1,68 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import pyomo.environ as pyo
```
nick-harder, line 5: this file just consists of a bunch of functions which are not bidding startegies, therefore they shouldn't be here. If you are using them in some particular strategy, they should be moved there. Or if this is part of the steel_plant, they should be there

assume/units/dst_components.py
```diff
@@ -0,0 +1,819 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import pyomo.environ as pyo
```
nick-harder, line 5: in this file please check again if everyhting is sufficiently documented and clear enough so users and future developers can follow up with it

assume/units/steel_plant.py
```diff
@@ -0,0 +1,503 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import logging
```
nick-harder, line 5: here the same, please check if everything is documented correctly and sufficiently for future use and development

assume/units/steel_plant.py
```diff
@@ -0,0 +1,503 @@
...
+        def steel_output_association_constraint(m, t):
+            return (
+                sum(
+                    self.dsm_components["eaf"].b.steel_output[t]
+                    for t in model.time_steps
+                )
+                == model.steel_demand
+            )
+
+        # @self.model.Constraint(self.model.time_steps)
```
nick-harder, line 255: some parts of the code are commented out. Are they needed? If not, please remove them

tests/test_dst_components.py
```diff
@@ -0,0 +1,350 @@
...
+    DriPlant,
+    DRIStorage,
+    ElectricArcFurnace,
+    Electrolyser,
+    GenericStorage,
+)
+
+
+# Test Electrolyser class
+@pytest.fixture
```
nick-harder, line 19: many tests had the same name, which doesn't work. I fixed it, but please use different names in the future

nick-harder (changes requested):

nick-harder:
@Manish-Khanra from my side everyhting is fine after the few comments are adressed. I have also cleaned up the commit a bit by removing unused parts

assume/strategies/dsm_load_shift.py
Manish-Khanra: Initially, I was considering creating a folder particularly for managing different flexibility measures for the consumers, but I wasn't sure where it should belong. What do you think about creating a folder in the strategies directory and naming it dsf_measures? This way, developers can easily add and manage scripts related to any demand side flexibility measures (e.g., in this example, load shift for industries) in this folder in the future.

assume/units/dst_components.py
Manish-Khanra: Checked and updated!

assume/units/steel_plant.py
Manish-Khanra: The irrelevant once are removed and for the relevant once, a description of its usage is added.

assume/units/steel_plant.py
Manish-Khanra: Here, the documentation was indeed incomplete. Now the script has been updated with the proper description.

tests/test_dst_components.py
Manish-Khanra: After a good thought, I think we can remove the test for dst_components for individual technologies sicne we have tested them in the test for the steel plant, already. What's your opinion on this?

nick-harder: it is up to you, what you think would be best for future development. But I believe keeping individual tests is also good since maybe these components will be reused in something else.

assume/strategies/dsm_load_shift.py
nick-harder: @maurerle what is your opinion on this matter? From my perspective these functions should be a part of the unit itself since they are written in unit specific manner and cannot be reused with different unit types other than steel plant.

nick-harder:
@Manish-Khanra thanks for adressing the comments and adding the documentation. Unfortunately, I believe you forgot to push your changes since I cannot see them

assume/strategies/dsm_load_shift.py
Manish-Khanra: @nick-harder The script can be reused for the other dsm units as well. We only have to add a if statement for the  'total_power_input_constraint_with_flex' constraint, which can handle other units. The if statement will distinguish the unit type.

Manish-Khanra:
@nick-harder Thanks for responding on the open points. I didnt push it until now because of the two other open points apart from documentation. However, now I have pushed the changes.

tests/test_dst_components.py
Manish-Khanra: Okay, then we will kept the test script for dst_components. :)

assume/strategies/dsm_load_shift.py
nick-harder: @Manish-Khanra oh, ok, good to know. but let's keep these functions in the dst_components.py for now and when we have a use case where they can be reused (whenever a new DSM units comes around), we will give it a better thought where to put these functions. For now it would be best to keep DSM stuff at one location.

Manish-Khanra: Sounds plausible! So if the documentation part looks fine we can proceed with the merging :)

tests/test_dst_components.py
```diff
@@ -0,0 +1,350 @@
...
+
+def test_dri_storage_constraints(dri_storage_unit):
+    dri_storage_unit.add_to_model(unit_block=pyo.Block(), time_steps=[0, 1, 2])
+    assert hasattr(dri_storage_unit.b, "storage_min_capacity_dri_constraint")
+    assert hasattr(dri_storage_unit.b, "storage_max_capacity_dri_constraint")
+    assert hasattr(dri_storage_unit.b, "energy_in_max_capacity_dri_constraint")
+    assert hasattr(dri_storage_unit.b, "energy_out_max_capacity_dri_constraint")
+    assert hasattr(dri_storage_unit.b, "energy_in_uniformity_dri_constraint")
+    assert hasattr(dri_storage_unit.b, "energy_out_uniformity_dri_constraint")
+    assert hasattr(dri_storage_unit.b, "storage_capacity_change_dri_constraint")
```
maurerle, line 288: These tests do not provide much value.

I would like to see a test which shows how the storage can be ramped up and down and that the optimization model behaves as wanted. I think this is missing for the other DST units as well

assume/units/dst_components.py
```diff
@@ -0,0 +1,903 @@
...
+
+    def add_to_model(self, unit_block, time_steps):
+        """
+        Adds the electrolyser unit to the Pyomo model.
+
+        Args:
+            unit_block (pyomo.Block): The Pyomo block where the electrolyser will be added.
+            time_steps (list): List of time steps for which the model will be defined.
+        """
+        self.b = unit_block
```
maurerle, line 69: Why do we have self.b (a pyomo block) and self.model (the whole model) here?
Shouldn't be either of the two sufficient?

assume/strategies/naive_strategies.py
```diff
@@ -137,6 +137,83 @@ def calculate_bids(
...
+                {
+                    "start_time": product[0],
+                    "end_time": product[1],
+                    "only_hours": product[2],
+                    "price": price,
+                    "volume": -volume,
+                }
+            )
+
+        return bids
```
maurerle, line 179: The two strategies are basically the same as the NaiveSingleBidStrategy
but an additional method of the unit is called and the result is retrieved from `unit.opt_power_requirement` instead of unit.outputs["energy"].

If we would find a different way to call the calculation/optimization model before bidding, we would not need another strategy, right?
Just for my understanding.

assume/strategies/naive_strategies.py
```diff
@@ -137,6 +137,83 @@ def calculate_bids(
...
+        unit: SupportsMinMax,
+        market_config: MarketConfig,
+        product_tuples: list[Product],
+        **kwargs,
+    ) -> Orderbook:
+        bids = []
+        start = product_tuples[0][0]  # start time of the first product
+
+        if unit.flex_power_requirement is None:
+            if unit.flexibility_measure == "max_load_shift":
```
maurerle, line 194: I would just call some `unit.optimal_flex_operation` or something and leave the decision of the flexibility_measure in the unit itself, not in the strategy - to determine the optimization target function

assume/units/dst_components.py
```diff
@@ -0,0 +1,903 @@
...
+        @self.b.Constraint(time_steps)
+        def storage_capacity_change_dri_constraint(b, t):
+            return b.soc_dri[t] == (
+                (
+                    (b.soc_dri[t - 1] if t > 0 else b.initial_soc_dri)
+                    * (1 - b.storage_loss_rate_dri)
+                )
+                + ((1 - b.charge_loss_rate_dri) * b.charge_dri[t])
+                - ((1 + b.discharge_loss_rate_dri) * b.discharge_dri[t])
+            )
```
maurerle, line 903: It seems that these "classes" do not have any kind of state or methods at all.
They could be modeled as well as factory functions, which receive a configuration dict and return the created pyomo block.

so that in steel_plant one simplyhas:

`self.model.add_block(create_storage_unit(**component_data))`

I would like that much better, except you see possibility that these classes will do much more in the future.

docs/source/img/architecture-noconfig.svg
maurerle, line 1: why are these changed? looks similar to me?

tests/test_steel_plant.py
```diff
@@ -0,0 +1,129 @@
...
+    assert steel_plant.flex_power_requirement is not None
+    assert isinstance(steel_plant.flex_power_requirement, pd.Series)
+
+    instance = steel_plant.model.create_instance()
+    instance = steel_plant.switch_to_flex(instance)
+    steel_plant.solver.solve(instance, tee=False)
+
+    total_power_input = sum(
+        instance.total_power_input[t].value for t in instance.time_steps
+    )
```
maurerle, line 123: couldn't we test and expect a little bit more here to check the functionality works correct?

assume/strategies/naive_strategies.py
Manish-Khanra: Two different stategies are created keeping in mind distinguising logic of determining bid price for day ahead and redispatch , each.

assume/strategies/naive_strategies.py
Manish-Khanra: Sorry didnt get you. We are choosing the flexibility_measure in the unit itself. In Unit we hav an attribute where we can guve a flexibility_measure to the unit.

assume/units/dst_components.py
Manish-Khanra: self.b is for the unit blocks. Here the dst_components are the blocks of the master model, (here its steel plant). So if we want to use global values from the steel plant and pass it on to the dst_componnts we have to use self.model.

maurerle: Oh okay. Thats reasonable :)

docs/source/img/architecture-noconfig.svg
nick-harder: this should not be changed but the new linting tool changes it somehow for us and cannot return it back

maurerle: oh okay - probably just adding a newline at the end..

assume/strategies/naive_strategies.py
maurerle: see a9eba5b73b64b78950e04c361d96b70d4ddfbadc which was merged into this branch :)

tests/test_steel_plant.py
nick-harder: @Manish-Khanra I also support this point

Manish-Khanra: @maurerle Based on your suggestions, I have enhanced the test script for testing the fucntions such as material and energy flow balance for the process.The following checks has been added:
Hydrogen Flow: Added verification on the hydrogen output from the electrolyser to the DRI plant at each time step.

Shaft to Arc Furnace Material Flow Test: Added a check to verify the material flow from the DRI plant to the Electric Arc Furnace is consistent at each time step.

Steel Output from EAF Test: Added a check that the total steel output from the Electric Arc Furnace over all time steps equals the specified steel demand.

I hope these checks are enough to test the steel plant is functionally valid and consistent.

tests/test_dst_components.py
Manish-Khanra: @nick-harder, @maurerle I have enhanced the test script for the dst_components to ensure it thoroughly tests the functionality and constraints of the individual components. Meanwhile, I have only written the test for the electrolyser. If you give a green signal I will proceed writing similar tests for the rest of the components. The following updates were made:

    Added a dummy objective function to minimize operating costs, encouraging the model to optimize power input values.

    Included detailed checks for ramp-up and ramp-down constraints to verify that the electrolyser respects these constraints at each time step.
    
    Ensured that the power input remains within the specified minimum and maximum bounds.

assume/units/dsm_load_shift.py
```diff
@@ -0,0 +1,70 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import pyomo.environ as pyo
+
+
+class DSMUnit:
```
maurerle, line 8: I think you wanted to rename this to something else, right? :) @Manish-Khanra

assume/world.py
```diff
@@ -129,6 +129,7 @@ def __init__(
             "power_plant": PowerPlant,
             "demand": Demand,
             "storage": Storage,
+            "steel_plant": SteelPlant,
```
maurerle, line 132: I don't think the real specialty of this model is steel?
In the long term, I would like to have a more generic solution, for example, something that also works for a pulp machine or other energy-intensive industries.
Having "Demand", "PowerPlant", "Storage" and then "Steelplant" seems a little weird.
But the steel plant is more of a general modeling approach to a multi step industrial process right?

tests/test_steel_plant.py
maurerle: This looks good so far!

maurerle (approved):

assume/units/dsm_load_shift.py
Manish-Khanra: Ah, Yes. Thanks for reminding. :)
Now its renamed.

assume/world.py
Manish-Khanra: The "SteelPlant" in this model focuses on integrating technologies specific to steel production. While I aim to create a more generic solution for various energy-intensive industries in the long term, currently, the model is tailored to the steel industry. When in the next step I develop similar models for other industries like cement will help identify commonalities for a unified approach in the future.

assume/units/dsm_load_shift.py
nick-harder, line 1: @maurerle why was this file created here in units? Since it is a kind of foundational file, should it be perhaps in the common/base.py file?

tests/test_dst_components.py
```diff
@@ -0,0 +1,91 @@
...
+    total_hydrogen_production = 300
+    model.total_hydrogen_constraint = pyo.Constraint(
+        expr=sum(model.electrolyser.hydrogen_out[t] for t in model.time_steps)
+        == total_hydrogen_production
+    )
+
+    return model
+
+
+def test_electrolyser_ramping_and_power_bounds(electrolyser_model):
```
nick-harder, line 51: I find this a very nice way to test the DST components. Maybe also adding some equality checks would make everything even better and more robust? So that some particular values match exactly when everything works as expected? For example checking the total cost to have some specific value at some time steps, or the power being some paticular values. There is no need to check for all of them, just one of the values would be enough

nick-harder (changes requested):
everyhting looks great so far! Please finilize the dst component checks and we can merge it. Also, please add your changed to the docs/source/release_notes so we will have a full description when releasing a new version

nick-harder (approved):

PR #349, opened 2024-05-06
Rl save policies debugging

kim-mskw:
so the old save policies function did not really use max_eval. It was always set back to -1000 when the learning role was wirtten new. I changed that now and added an early stopping criterium. 
Something works better but something weird as well, because the max avg metrik is now always 19.28 

This is the basis for the discussion with @nick-harder ick- tomorrow

### DIFF
--- a/assume/reinforcement_learning/algorithms/matd3.py
+++ b/assume/reinforcement_learning/algorithms/matd3.py
@@ -328,7 +328,7 @@ class TD3(RLAlgorithm):
             This function implements the TD3 algorithm's key step for policy improvement and exploration.
         """
 
-        # logger.info(f"Updating Policy")
+        logger.info(f"Updating Policy")
         n_rl_agents = len(self.learning_role.rl_strats.keys())
         for _ in range(self.gradient_steps):
             self.n_updates += 1
--- a/assume/reinforcement_learning/algorithms/matd3.py
+++ b/assume/reinforcement_learning/algorithms/matd3.py
@@ -328,7 +328,7 @@ class TD3(RLAlgorithm):
             This function implements the TD3 algorithm's key step for policy improvement and exploration.
         """
 
-        logger.info(f"Updating Policy")
+        # logger.info(f"Updating Policy")
         n_rl_agents = len(self.learning_role.rl_strats.keys())
         for _ in range(self.gradient_steps):
             self.n_updates += 1
--- a/assume/reinforcement_learning/learning_role.py
+++ b/assume/reinforcement_learning/learning_role.py
@@ -45,6 +45,8 @@ class Learning(Role):
         self.buffer: ReplayBuffer = None
         self.obs_dim = learning_config["observation_dimension"]
         self.act_dim = learning_config["action_dimension"]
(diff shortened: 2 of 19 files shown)


kim-mskw:
okay I know why it is always the same, because it somehow now always loads old policies in a dead folder. I am fixing that. This is terribly confusing.

kim-mskw:
so there was a function in the learning_strategies that loaded policies and one in the learning_algo, leading to very weird behaviour after the load_scenario function. Basically in the eval run always an old policy from a folder was read. 

Technically we need the actor in the bidding strategy, because otherwise it is not initialized if we do not use learning, so learning mode off and hence no algo as well. This double initializing is weird though.

assume/reinforcement_learning/algorithms/matd3.py
```diff
@@ -328,7 +328,7 @@ def update_policy(self):
             This function implements the TD3 algorithm's key step for policy improvement and exploration.
         """
 
-        logger.info(f"Updating Policy")
+        # logger.info(f"Updating Policy")
```
nick-harder, line 331: please remove this change, I have changed it to a debug message in my commit

assume/scenario/loader_csv.py
```diff
@@ -367,7 +368,7 @@ async def load_scenario_folder_async(
     if not learning_config.get("trained_policies_save_path"):
         learning_config[
             "trained_policies_save_path"
-        ] = f"./learned_strategies/{study_case}"
+        ] = f"./learned_strategies/{study_case}/last_policies"
```
nick-harder, line 371: last_policies are set twice now, here and also in load functions of the learning strategies.

assume/reinforcement_learning/learning_role.py
```diff
@@ -45,6 +45,8 @@ def __init__(
         self.buffer: ReplayBuffer = None
         self.obs_dim = learning_config["observation_dimension"]
         self.act_dim = learning_config["action_dimension"]
+        self.early_stopping_steps = learning_config["early_stopping_steps"]
```
nick-harder, line 48: these should have some default value set by us to make things a bit easier for the user

compose.yml
```diff
@@ -17,16 +17,16 @@ services:
       - TS_TUNE_MAX_CONNS=500
     volumes:
       # needed for normal image
-      #- ./assume-db:/var/lib/postgresql/data
+      - ./assume-db:/var/lib/postgresql/data
```
nick-harder, line 20: why was the docker file changed?

examples/inputs/example_02a/config.yaml
```diff
@@ -49,7 +49,7 @@ tiny:
 
 base:
   start_date: 2019-03-01 00:00
-  end_date: 2019-04-01 00:00
+  end_date: 2019-03-02 00:00
```
nick-harder, line 52: this was changed probably only for testing or?

compose.yml
kim-mskw: Oh sorry, for my laptop the other image never works, I commited it accidently

PR #353, opened 2024-05-14
Split scenario loader function to improve speed and reduce unrequired operations

nick-harder:
This PR splits the scenario loader into two parts, keeping the files and other things in memory to avoid loading everyhting multiple times when in training


kim-mskw:
Looks good to me :) I would appreciate Florian thought though, as I am also pretty tired :D

assume/scenario/loader_csv.py
```diff
@@ -277,85 +277,143 @@ def add_units(
...
-    scenario: str,
-    study_case: str,
-    perform_learning: bool = True,
-    perform_evaluation: bool = False,
-    episode: int = 0,
-    eval_episode: int = 0,
-) -> None:
+def load_config_and_create_forecaster(
+    inputs_path: str, scenario: str, study_case: str
+) -> Dict[str, Any]:
```
maurerle, line 282:
```suggestion
) -> dict[str, object]:
```

assume/scenario/loader_csv.py
```diff
@@ -277,85 +277,143 @@ def add_units(
...
+        "powerplant_units": powerplant_units,
+        "storage_units": storage_units,
+        "demand_units": demand_units,
+        "forecaster": forecaster,
+    }
+
+
+async def async_setup_world(
+    world: World,
+    data: Dict[str, Any],
```
maurerle, line 379: `scenario_dict` or `scenario_data` would probably fit better?

maurerle (approved):
I have two minor improvements.

In general one does not need a lot from typing anymore with python 3.9 or greater :)

PR #354, opened 2024-05-14
adjusted learning loop in accordance with description in issue

kim-mskw:
image
image


assume/reinforcement_learning/learning_role.py
```diff
@@ -105,6 +106,41 @@ def __init__(
...
+        self.episodes_done = inter_episodic_data["episodes_done"]
+        self.eval_episodes_done = inter_episodic_data["eval_episodes_done"]
+        self.max_eval = inter_episodic_data["max_eval"]
+        self.rl_eval = inter_episodic_data["all_eval"]
+        self.avg_rewards = inter_episodic_data["avg_rewards"]
+        self.buffer = inter_episodic_data["buffer"]
+
+        self.initialize_policy(inter_episodic_data["actors_and_critics"])
+
+    def dump_inter_episodic_data(self):
```
maurerle, line 126: I would associate "dump" with writing something to a file/stream - "get_inter_episodic_data" would fit better.?

assume/reinforcement_learning/learning_role.py
```diff
@@ -114,21 +150,23 @@ def setup(self) -> None:
...
-        )
 
-        recurrency_task = rr.rrule(
-            freq=rr.HOURLY,
-            interval=self.train_freq,
-            dtstart=self.simulation_start,
-            until=self.simulation_end,
-            cache=True,
-        )
+        if self.perform_evaluation is False:
```
maurerle, line 154:
```suggestion
        if not self.perform_evaluation:
```

assume/scenario/loader_csv.py
```diff
@@ -571,7 +570,6 @@ def load_scenario_folder(
 
     Notes:
         - The function sets up the world environment based on the provided inputs and configuration files.
-        - If `perform_learning` is set to True and learning_mode is set, the function initializes the learning mode with the specified episode number.
         - If `perform_evaluation` is set to True, the function performs evaluation using the specified evaluation episode number.
         - The function utilizes the specified inputs to configure the simulation environment, including market parameters, unit operators, and forecasting data.
```
maurerle, line 574: `terminate_learning` is missing here.

Terminating learning sounds like starting it, but then stopping it.
I think you meant it mot like "unterbinden"?

Would `training_enabled` be a better wording for that variable? (of course negating its bool value)

So "learning_enabled" -> should we do anything learning related
"training_enabled" -> should we perform further rl training?

maurerle (commented):
The tests did not pass yet, I had some minor remarks, but it looks good :)

assume/reinforcement_learning/learning_role.py
kim-mskw: good point

assume/scenario/loader_csv.py
kim-mskw: then the distinction between the learning mode is not really clear I thought. I would especially leave it at terminating, because we end the learning process, this is also the wording of the early stopping

maurerle: Thats fine too :)

assume/scenario/loader_csv.py
```diff
@@ -360,22 +359,28 @@ async def load_scenario_folder_async(
...
     config = replace_paths(config, path)
 
-    if learning_config.get("learning_mode", False):
+    if learning_config.get("learning_mode", False) and not learning_config.get(
+        "evaluation_mode", False
+    ):
         sim_id = f"{sim_id}_{episode}"
 
-    if learning_config.get("evaluation_mode", False):
+    elif learning_config.get("learning_mode", False) and learning_config.get(
```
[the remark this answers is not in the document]
kim-mskw: it was like this before and did not work, so I would leave it like this

nick-harder (approved):

PR #355, opened 2024-05-15
Speed up simulation

nick-harder:
-refactor the function itself to use vectorization
-execute the funtion only once at the end of the simulation
-paralellize the call for all agents to improve performance

suggestion by @maurerle

The speed up is as follows (tested on small example):
current version - 24s
vectorized calculate_generation_cost function - 16s
execute calculation only at the end of the simulation - 8s

Using the small_learning_1 example:
current version - 36s
final tweaked version - 10s

### DIFF
--- a/assume/common/base.py
+++ b/assume/common/base.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 from datetime import datetime, timedelta
 from typing import TypedDict, Union
 
+import numpy as np
 import pandas as pd
 
 from assume.common.forecasts import Forecaster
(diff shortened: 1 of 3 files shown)


assume/common/base.py
```diff
@@ -176,14 +177,20 @@ def calculate_generation_cost(
...
             end (datetime.datetime): The end time for the calculation.
             product_type (str): The type of product for which the generation cost is to be calculated.
-
         """
-        if start not in self.index:
-            start = self.index[0]
+        if end != self.index[-1]:
+            return  # do nothing before the end of the simulation
+
+        start = self.index[0]
```
maurerle, line 184: Noone will know, that this basically makes sure that it only runs at the end and therefore runst from the start on..

1. add a param "run_at_end_only" and put line 181-184 into a `if run_at_end_only` check - this makes things more clear
2. refactor the `calculate_generation_cost` function to be only run in `on_stop` of the UnitsOperator for the whole time once..

Either way is fine for me

assume/common/units_operator.py
```diff
@@ -327,7 +329,14 @@ def get_actual_dispatch(
...
 
             data["unit"] = unit_id
-            unit_dispatch_dfs.append(data)
+            return data
+
+        with ThreadPoolExecutor() as executor:
+            results = executor.map(
+                lambda unit_id_unit: process_unit(*unit_id_unit), self.units.items()
+            )
+            unit_dispatch_dfs.extend(results)
```
maurerle, line 338: are you sure the ThreadPoolExecuter helps here?
From what I know, you probably need to use a `multiprocessing.Pool` to do this parallel..?

assume/common/base.py
nick-harder: I have having the problem that using this approach only the last value is written into db. I tried the on_stop idea, but the output role is killed before this one, so doesn't work as well. Do you have a suggestion?

PR #356, opened 2024-05-15
Fixes saving policy into the main folder instead of inputs

nick-harder:


kim-mskw (approved):
Thanks for fixing this

PR #358, opened 2024-05-22
This fixes the loading of actors

nick-harder:
-screate a deepcopy of the scenario data to avoid changing it during the simulation 
-load last_policies by default if not policies are specified 
-raise an error if RL policy is selected but no policies are provided and learning is deactivated


maurerle (approved):
Looks fine to me

PR #360, opened 2024-06-17
update architecture.svg

maurerle:
I updated the architecture.svg to match recent developments.
Though I am open for more input on this change


maurerle:
Side by side:
On the left, the new layout better reflects the actual usage of markets in the architecture.

image

kim-mskw (commented):
I do like the idea of @maurerle to highlight the market config more as a particular item. To shocase the adapability of the market. Would we add the other config then as well?

Also I like that the clearing has ecampels from which the markets evolve, which represents more what we have implemnted. I would kepp the position though and then add the network to the clearing?

I agree that the two sides divided/connected by world and order should be kept for easy readablility.

maurerle:
New images:



and I created another one without minimized configs:


I would just add the architecture.svg

I also updated the actor and critic usage according to the usage in the source code, and updated the LearningRole :)

kim-mskw:
That looks good. Thank you :)

With regard to the learning, I would suggest, that the actor is still connected as before to the learning role because the learning role has an algorithm that has the actors. Though we do not depict the algorithm here. To keep it simple I would not do so, but add it below the buffer and then draw the line from learning role to actor again. I can also do that if you provide me with the file.

maurerle:
Hi Kim, actually the LearningRole has a list of learning_strategies and can access the actos through these, as correctly shown in the image ;)
Though I am open for suggestions and improvements.

You can edit the file by just opening the svg with draw.io/[diagrams.net](https://app.diagrams.net/)
Just do as you like :)

kim-mskw:
> Hi Kim, actually the LearningRole has a list of learning_strategies and can access the actos through these, as correctly shown in the image ;) Though I am open for suggestions and improvements.
> 
> You can edit the file by just opening the svg with draw.io/[diagrams.net](https://app.diagrams.net/) Just do as you like :)

you are of course right, because we changed that

kim-mskw (approved):
as per the discussion, it looks good :)

PR #368, opened 2024-06-20
Fix opportunity costs

kim-mskw:
Minor changes to fix opportunity costs, they were falsy added per order, yet they are order independent. And sheduled the first eval run to x episodes after initial experience


maurerle (approved):
Looks good from my side,

Tests are failing though

PR #372, opened 2024-06-20
Fix learning when action dimension equals one

nick-harder:
fixes the raised error when using a learning startegy with action dimension of one


maurerle (approved):
This checks if the actions are a tuple and assumes a dimension of 1 if they are not.
Looks fine to me.

PR #376, opened 2024-06-25
Adjust the framework to be able to schedule storing to the learning role

nick-harder:
To be able to simulate several markets using one single RL agents we need to be able to save the rewards only after all the modeled markets have been cleared and the final reward has been calculated. This PR brings the required changes to the framework by scheduling the saving to the buffer and policy updates every train_freq. If train_freq is configured to match the closure of all modelled markets, the correct rewards is sved, thus enabling to model several market with one RL agent.


assume/reinforcement_learning/learning_role.py
```diff
@@ -182,14 +163,16 @@ def handle_message(self, content: dict, meta: dict) -> None:
...
-        if content.get("type") == "replay_buffer":
+        if content.get("type") == "save_buffer_and_update":
             data = content["data"]
             self.buffer.add(
                 obs=data[0],
                 actions=data[1],
                 reward=data[2],
             )
 
+        self.update_policy()
```
kim-mskw, line 174: The way you have sheduled the update policy now, right after the save to buffer, will also update the policy druing the inital exploration. This should not be a problem besides slowing the simualtion down right?

assume/strategies/learning_strategies.py
```diff
@@ -178,8 +178,8 @@ def calculate_bids(
         ]
 
         # store results in unit outputs which are written to database by unit operator
-        unit.outputs["rl_observations"][start] = next_observation
-        unit.outputs["rl_actions"][start] = actions
+        unit.outputs["rl_observations"].append(next_observation)
+        unit.outputs["rl_actions"].append(actions)
         unit.outputs["rl_exploration_noise"][start] = noise
```
kim-mskw, line 183: This is differen (append vs [start]) because you commented it out right? Shoudl we comment everything out then?

assume/world.py
```diff
@@ -328,7 +312,12 @@ def add_unit_operator(self, id: str) -> None:
         if self.unit_operators.get(id):
             raise ValueError(f"Unit operator {id} already exists")
 
-        units_operator = UnitsOperator(available_markets=list(self.markets.values()))
+        units_operator = UnitsOperator(
+            simulation_start=self.start,
+            simulation_end=self.end,
+            available_markets=list(self.markets.values()),
+            train_freq=self.learning_config.get("train_freq", 24),
```
kim-mskw, line 319: I do not like the way the tran_freq is now handeled by the unit operator as it would be technically the task of the learning role. Just definition wise. Is there a way to keep it in the learning role somehow? Maybe we can discuss that?

kim-mskw (changes requested):

assume/reinforcement_learning/learning_role.py
nick-harder: the update_policy function anyway checks if we are still in the initial exploration, so no updates are actually executed

assume/strategies/learning_strategies.py
nick-harder: I don't understand this point, please clarify

assume/world.py
nick-harder: @kim-mskw I have now moved the definition of the task into the world setup itself. Please take a look at the new version

nick-harder:
@maurerle @kim-mskw it would be great if you please could review this PR today so I could continue working on the sequential market testing and implementation. Thanks in advance!

assume/common/units_operator.py
```diff
@@ -573,120 +599,74 @@ def write_learning_to_output(
...
+        )
+        all_rewards = np.array(all_rewards).reshape(-1, learning_unit_count)
         rl_agent_data = (all_observations, all_actions, all_rewards)
 
         learning_role_id = self.context.data.get("learning_agent_id")
         learning_role_addr = self.context.data.get("learning_agent_addr")
 
         if learning_role_id and learning_role_addr:
-            self.context.schedule_instant_acl_message(
+            await self.context.schedule_instant_acl_message(
```
maurerle, line 664: afaik this is not an async function?

assume/scenario/loader_csv.py
```diff
@@ -815,7 +794,7 @@ def run_learning(
         if (
             episode % validation_interval == 0
             and episode
-            > world.learning_role.episodes_collecting_initial_experience
+            >= world.learning_role.episodes_collecting_initial_experience
```
maurerle, line 797: as the first episode is 1, you probably want to have > here - otherwise you would get no run without `initial_experience` when `episodes_collecting_initial_experience=1`

examples/inputs/example_02b/config.yaml
```diff
@@ -11,13 +11,13 @@ base:
 
   learning_config:
     continue_learning: False
-    trained_policies_save_path: learned_strategies/example_02b_base
+    trained_policies_save_path: Null
```
maurerle, line 14: you want `null` to set it to "None" - `Null` would be interpreted as string - that path does not exist

maurerle (commented):
I looked over the code, some things are still unclear to me. I think it would be best if kim tests this branch as well to see, that everything gives reasonable results/still works?

assume/common/units_operator.py
nick-harder: oh, idk where that came from, thanks

examples/inputs/example_02b/config.yaml
nick-harder: somehow for me only Null works, not null

nick-harder:
> I looked over the code, some things are still unclear to me. I think it would be best if kim tests this branch as well to see, that everything gives reasonable results/still works?

Thanks! I have tested learning for single agent and multi agent (2a and 2b) and everyhting works for me. But yeah, it would be great to get a second look since this brings many changes

assume/scenario/loader_csv.py
nick-harder: since it is initial_experience+validation_interval, it will be at least 2, so this should be fine. Otherwise with > the first evaluation is done very late. For example, if eval=5 and init_experience=5, the first evaluation happens after episode=15

assume/reinforcement_learning/learning_role.py
kim-mskw: Ah okay, it is weird that this is made sure of in two different ways, but then fine.

assume/world.py
kim-mskw: yeah it makes more sense there somehow than in the unit operator. The tidiest way would still be in the learning role, but I see why that is not possible

assume/strategies/learning_strategies.py
kim-mskw: Nevermind, not thta important

kim-mskw (approved):

PR #377, opened 2024-06-25
Distributed csv loader using -p cli param

maurerle:
This adds a CLI param `-p` or `--parallell` which runs each agent on a separate process in its own world.
This allocates port 9100 and following (one per UnitsOperator).

In scenarios with few unitsoperators this does not have any benefit and might slow things down as well, though it helps for larger ones:

one can compare `assume -s example_02 -p` (5000 it/s) with the serialized version `assume -s example_02` (2000 it/s).
While it does slow things down when using `assume` (100000 it/s) vs `assume -p` (78000 it/s)


assume/world.py
```diff
@@ -412,29 +507,17 @@ async def async_add_unit(
...
-
-        self._validate_unit_addition(id, unit_type, unit_operator_id)
-
-        # provided unit type does not exist yet
-        unit_class: type[BaseUnit] = self.unit_types.get(unit_type)
 
-        bidding_strategies = self._prepare_bidding_strategies(unit_params, id)
-        unit_params["bidding_strategies"] = bidding_strategies
+        # check if unit operator exists
+        if unit_operator_id not in self.unit_operators.keys():
```
nick-harder, line 512: there is a function _validate_unit_addition in place to check all these

assume/world.py
```diff
@@ -584,12 +644,16 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
         self.markets[f"{market_config.market_id}"] = market_config
 
     async def _step(self):
+        if self.distributed_role:
```
nick-harder, line 647: please add documentation as docsting to this function. It is very hard to understand and won't be understood by anyone except for the core developers

examples/distributed_simulation/main.py
```diff
@@ -0,0 +1,60 @@
+# SPDX-FileCopyrightText: ASSUME Developers
```
nick-harder, line 1: the whole idea is really great, but I doubt anyone would use it as it is very unclear what exactly is happening. Please always keep in mind, that we are cretaing this not for our own use, so only us can understand what is written here. Especially since this part is in the examples, here should be clear comments and details what is happening and why

assume_cli/cli.py
```diff
@@ -104,6 +104,13 @@ def cli(args=None):
         choices=set(logging._nameToLevel.keys()),
     )
 
+    parser.add_argument(
```
nick-harder, line 107: is there a description of available parameters for the CLI in the documentation?

nick-harder (changes requested):
good work overall, but many things are unclear, and better documentation is required. Also, naming of the functions is very confusing. There should be a clear differentiation between the normal functions and the ones for distributed execution. Please take a critical view on the changes taking a position of external collaborator and asking yourself if this new function or a piece of code is explainable/clear, and if the names are clear or rather confusing. Thanks :-)

nick-harder:
@maurerle thanks for adressing the comments! I will check it out in details this week.

I believe this code if merged before the steel plant will make main uncompatible, and I would like to avoid puting extra tasks on Manish to understand and implement the changes from this PR into the steel plant. Would it be acceptable for you if we merge steel plant first and you could resolve the merge conflicts in case those arrise?

maurerle:
I can take of the merge conflicts, no matter which gets merged first.
It is fine for me :)

nick-harder (approved):
Everything is fine now! Please @maurerle don't forget to add the description of the changes to the doc/release_notes.rst. afterwards it can be merged

nick-harder:
@maurerle would you like to finilize this PR and merge it?

PR #378, opened 2024-06-26
Add the larger example from the paper

nick-harder:
-move the complex bid example to 2e
-have exmaple_02c as example with 10 learning plants similar to the paper


maurerle:
It might be helpful to specify which paper this is referring to ;)

PR #390, opened 2024-07-16
Improve steel plant working

maurerle:
I worked a little around the DSM units code and refactored it to simplify it (hopefully).

* change dst_components to factory methods 
* remove model argument for methods which already have self.model available 
* create abstract class DSMUnit for flexibility_cost_tolerance call
* let unit decide what to do based on its objective and flexibility_measure

This PR is against the test_stell_plant_integration branch.
If this is fine for you @Manish-Khanra I would merge this?

The tests need to be adapted as well of course.


nick-harder:
@maurerle thanks for the adjustments, I believe this looks good, let's merge it

PR #397, opened 2024-07-30
Doc fix pr

nick-harder:


nick-harder:
@kim-mskw I believe Florian is on a vacation and didn't review this pull request. Could you please review and merge it, so we have a working documentation, and Florian can fix it when he is back?

kim-mskw (commented):
I guess this is fine as a quick fix until the original problem is solved

kim-mskw (approved):
I guess this is fine as a quick fix until the original problem is solved.

PR #399, opened 2024-08-06
Fix loader and examples

nick-harder:
## Description
- check if availability_df containt values >1 and warn user if so
- fix these values automatically by normalizing them
- also check in the power plant itself when calculating max_power if availability is larger than max_power
- restructure examples for easier understanding
- pack all feature example to example_01 and all learning to example_02
- add a full year DRL scenario from the fit paper
- all year examples are now as example_03

## Checklist


assume/common/utils.py
```diff
@@ -538,3 +538,30 @@ def create_nodal_incidence_matrix(lines, buses):
...
+            max_power = powerplants_df.loc[column, "max_power"]
+        except KeyError:
+            logger.warning(
+                f"Warning: Unit '{column}' not found in powerplants dataframe. Skipping normalization for this unit."
+            )
+            continue
+
+        # Normalize values greater than 1 by dividing by max_power
+        normalized_df[column] = normalized_df[column].apply(
+            lambda x: x / max_power if x > 1 else x
```
kim-mskw, line 561: If the unnormalized df has a value that is below 1 by coincidence this will not be overwritten and hence the normalisation is wrong for that value. I would suggest testing wheter one value it above 1 and then noprmlising everything. What do you think?

kim-mskw: Wait we already check if some values are bigger 1, so we just do not need this condition anymore, right?

examples/examples.py
```diff
@@ -51,22 +52,27 @@
         "scenario": "example_01d",
         "study_case": "zonal_case",
     },
-    "large_2019_eom": {"scenario": "example_02", "study_case": "base_case_2019"},
+    # example_01e is used in the tutorial notebook #3: Custom unit and bidding strategy example
+    # example_01f is used in the tutorial notebook #5: Market configuration comparison example
```
kim-mskw, line 56: Why are they not here anymore?

examples/examples.py
```diff
@@ -51,22 +52,27 @@
...
+    # example_01g is used in the tutorial notebook #6: Advanced order types example
+    #
+    # small examples with DRL enabled to show the learning process and configuration
+    "small_learning_1": {"scenario": "example_02a", "study_case": "base"},
+    "small_learning_2": {"scenario": "example_02b", "study_case": "base"},
```
kim-mskw, line 61: I would suggest sorting the Rl exampels into the ones representing the cass in your study @nick-harder and then other RL exampels.

examples/examples.py
```diff
@@ -16,21 +16,22 @@
 os.makedirs("./examples/local_db", exist_ok=True)
 
 availabe_examples = {
+    # small examples for easier understanding of different features and configurations
```
kim-mskw, line 19: Please also change this in the example_simulations.rst file

kim-mskw (changes requested):

examples/examples.py
nick-harder: they were never here in the dictionary. I have just added these comments so if someone sees those files and doesn't see them in this dict, they know what they are

examples/examples.py
nick-harder: but isn't this already the case? or what do you mean?

examples/examples.py
nick-harder: good point, I have now completely reworked that file to include all examples!

examples/examples.py
nick-harder: I have now added a clear reference to the cases from the publication in the rst description

kim-mskw: pushed my thought

assume/scenario/loader_csv.py
```diff
@@ -54,6 +58,8 @@ def load_file(
     df = None
 
     if file_name in config:
```
maurerle, line 60: having a file_name = None would differ from having it unset.
Is this wanted?

I would suggest to just check: `if config.get(file_name)` instead?

assume/scenario/loader_csv.py
```diff
@@ -372,7 +378,7 @@ def read_units(
         bidding_strategies = {
             key.split("bidding_")[1]: unit_params[key]
             for key in unit_params.keys()
-            if key.startswith("bidding_")
+            if key.startswith("bidding_") and unit_params[key] != 0
```
maurerle, line 381: I think you generally do not want to have empty values as a given strategy here - so None and empty string as well, so I would just check `and unit_params[key]`

```suggestion
            if key.startswith("bidding_") and unit_params[key]
```

maurerle (commented):
I have some small remarks :)

And I updated the docs to have proper footnotes

assume/scenario/loader_csv.py
nick-harder: The idea is that I can have one set of inputs, but disable, for example, loading storage units, so in my config, I would say storage_units: null

maurerle: sounds good

maurerle (approved):

PR #400, opened 2024-08-07
Readd CLI docs

maurerle:
readds the cli arg docs

also updates pyyaml-include to be compatible with latest version

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,7 +11,7 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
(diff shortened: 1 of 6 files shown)


maurerle:
Looks fine:
https://assume--400.org.readthedocs.build/en/400/command_line_interface.html

I also adapted the docs building instructions to do the same as readthedocs so that the relative paths as compatible, as long as this is not fixed in sphinx-argparse - see https://github.com/sphinx-doc/sphinx-argparse/issues/61

nick-harder (approved):

PR #403, opened 2024-08-08
Manual strategy for interactive market simulation

maurerle:
This adds a simple strategy which waits for the user to give a volume and price interactively.

At the last project meeting, I discussed such an idea and had in mind to contribute this "poor mans simulation" strategy, while it would be much cooler to have a web interface to interact with this in the future.

I added some docs on how this can be used as well.
While not particularly useful at the moment, I think this is also a good showcase of how flexible the framework is and can also be used to debug behavior, by inserting the wanted bids directly to the market.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ sphinx-apidoc -o docs/source -Fa assume
 To create and serve the documentation locally, use:
 
 ```bash
-cd docs/source && python -m sphinx . ../build && cd ../.. && python -m http.server --directory docs/build/html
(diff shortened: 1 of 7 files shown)


nick-harder (approved):
@maurerle nice idea! this semester I have written the following code for such a game in my lectures, it has a web based interface to submit bids for students and a teacher view. if we could couple this with assume, this would be really cool
https://github.com/INATECH-CIG/merit_order_game

PR #404, opened 2024-08-16
add grid plotting to file if postgis is not available in the database

maurerle:
@Manish-Khanra  asked me about this functionality, so I create a short draft how this could be done if docker is not available.

The output path should be adjusted though, and I don't like having imports in the middle of the file :D


nick-harder:
no activity

PR #406, opened 2024-08-29
Fixup ruff and manual strategies docs

maurerle:
ruff 0.5.6 introduced support for ipynb linting, so this is fixed as well:
https://github.com/astral-sh/ruff/releases

I cherry-picked the commit which fixes this and added the manual strategies to the docs


mthede:
I ran `pre-commit autoupdate` to check for other updates on the pre-commit hooks. That showed that not only ruff but also the other hooks can be updated. 
grafik

Maybe we can adjust the .pre-commit-config.yaml file to keep up with the newer versions? Or are there other dependencies and we shouldn't change the pre-commit config?

maurerle:
I did not know about that handy `pre-commit autoupdater` - very nice!
For now, there is now reason to pin anything (if so, I would typically expect a comment nearby with a link to the issue why pinning is needed), so I just bumped the dependencies.

mthede (approved):

PR #382, opened 2024-07-01
Lstm learning to main

kim-mskw:
# Pull Request

## Related Issue
Closes #270

## Description
Enable the LSTM policies instead of the fully connected NN for RL in the MATD3 algorithm 

## Testing
Created a new environment and ran the examples and notebooks

## Checklist
Please check all applicable items:

- [x] Code changes are sufficiently documented (docstrings, inline comments, `doc` folder updates)
- [x] New unit tests added for new features or bug fixes
- [x] Existing tests pass with the changes
- [x] Reinforcement learning examples are operational (for DRL-related changes)
- [x] Code tested with both local and Docker databases
- [x] Code follows project style guidelines and best practices
- [ ] Changes are backwards compatible, or deprecation notices added
- [x] New dependencies added to `pyproject.toml`
- [x] A note for the release notes `doc/release_notes.rst` of the upcoming release is included
- [x] Consent to release this PR's code under the GNU Affero General Public License v3.0

## Additional Notes (if applicable)
[Any additional information, concerns, or areas you want reviewers to focus on]

## Screenshots (if applicable)
[Add screenshots to demonstrate visual changes]

PR #410, opened 2024-09-10
use inheritance between RL strategies to remove code duplication of RLStrategy

maurerle:
Even though I am a little late to add it to #382 - I refactored small parts to be less redundant.

* let RLAdvancedOrderStrategy inherit from RLStrategy, so that there is less duplicate code (the default obs_dim was 51 instead of 50 and unique_obs_dim=3 instead of 2, but I don't think we are using these defaults effectively. So this PR should not change any functionality at all.
Duplicate functions were
    * constructor
    * load_actor_params
    * get_actions
* fix some typos
* remove hard coding of available actor architectures


kim-mskw (commented):
That amkes sense. The code was indeed redundant. I will check everything again.

kim-mskw:
learning_with_complex_bids is not functional, I am looking into it

kim-mskw (approved):

PR #414, opened 2024-09-12
add first 3 jupyter notebooks to CI runs

maurerle:
This adds the first 3 less complex jupyter notebooks to the CI runs.

The other notebooks are running to long for them to be convenient, so I left them out.


maurerle:
I had to change
`!git clone https://github.com/assume-framework/assume.git`
to 
`!git clone https://github.com/assume-framework/assume.git assume-repo`

so that Python does not try to import from the checked out `assume` folder on `from assume import World`, but other than that, it at least works for the first 3 notebooks to be tested in CI (the others are a little to complex or runtime intensive)

PR #416, opened 2024-09-16
example_01c did not work, as it includes agent with complex bidding

maurerle:
* fix in initialization of complex_clearing without grid
* give better error feedback in marker clearing
* log result saving frequency at start of simulation


nick-harder:
@maurerle thanks, I was also looking into this probem, but it is still not working, specifically the small_with_BB_and_LB case. trying to figure out the problem

PR #417, opened 2024-09-16
Fix storages

maurerle:
This includes various fixes to the current storage calculation:

* remove ramping constraints if non are set
* if price forecast favors discharge, but max discharge is zero, set a bid for charging (instead of a bid with a volume of 0)
* make sure that all `charge` properties are negative and all `discharge` properties are positive
* refactor calculate_ramp_charge/calculate_ramp_discharge in base.py

I discussed with @kim-mskw about changing this, so that all max_charge/discharge properties are positive, but I dropped this idea as I had a hard time chasing errors that arose from that.

Some tests had to be adjusted, I am quite sure that the ramping (which is actually not really used for now) works the way as intended now.

Background:

When adding a storage_unit to example_01a, one did have an hour in which no bid was set at all.
I tried to find that. As the market returns the rejected bids as well, this was not an issue of the market but of the bidding strategy. This is fixed by the second bullet.

This PR is not yet ready to be merged, Documentation is still missing

### DIFF
--- a/assume/strategies/flexable_storage.py
+++ b/assume/strategies/flexable_storage.py
@@ -130,7 +130,8 @@ class flexableEOMStorage(BaseStrategy):
 
             # if price is higher than average price, discharge
             # if price is lower than average price, charge
-            if price_forecast[start] >= average_price:
+            # if price forecast favors discharge, but max discharge is zero, set a bid for charging
+            if price_forecast[start] >= average_price and max_power_discharge[start]:
                 price = average_price / unit.efficiency_discharge
                 bid_quantity = max_power_discharge[start]
             else:
--- a/assume/units/storage.py
+++ b/assume/units/storage.py
@@ -130,18 +130,18 @@ class Storage(SupportsMinMaxCharge):
         # if ramp_up_charge == 0, the ramp_up_charge is set to the max_power_charge
         # else the ramp_up_charge is set to the negative value of the ramp_up_charge
         self.ramp_up_charge = (
-            self.max_power_charge if ramp_up_charge is None else -abs(ramp_up_charge)
+            self.max_power_charge - self.max_power_discharge if ramp_up_charge is None else -abs(ramp_up_charge)
         )
         self.ramp_down_charge = (
-            self.min_power_charge
+            self.max_power_charge - self.max_power_discharge
             if ramp_down_charge is None
             else -abs(ramp_down_charge)
         )
         self.ramp_up_discharge = (
-            self.max_power_discharge if ramp_up_discharge is None else ramp_up_discharge
+            self.max_power_discharge - self.max_power_charge if ramp_up_discharge is None else ramp_up_discharge
         )
         self.ramp_down_discharge = (
-            self.min_power_discharge
+            self.max_power_discharge - self.max_power_charge
             if ramp_down_discharge is None
             else ramp_down_discharge
         )
--- a/assume/units/storage.py
+++ b/assume/units/storage.py
@@ -123,7 +123,7 @@ class Storage(SupportsMinMaxCharge):
         self.emission_factor = emission_factor
 
         # The ramp up/down rate of charging/discharging the storage unit.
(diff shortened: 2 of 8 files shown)


assume/strategies/flexable_storage.py
```diff
@@ -130,7 +130,8 @@ def calculate_bids(
 
             # if price is higher than average price, discharge
             # if price is lower than average price, charge
-            if price_forecast[start] >= average_price:
+            # if price forecast favors discharge, but max discharge is zero, set a bid for charging
```
kim-mskw, line 133: Where is this implemented? Or is this outdated?

maurerle: This is implemented through the change in the line below:

`if price_forecast[start] >= average_price and max_power_discharge[start]:`

By adding the second clause, we only set a discharge bid, if we do have something to discharge ;)

assume/strategies/flexable_storage.py
```diff
@@ -130,7 +130,8 @@ def calculate_bids(
...
+            if price_forecast[start] >= average_price and max_power_discharge[start]:
```
kim-mskw, line 134: Why not do nothing in this case?

assume/units/storage.py
```diff
@@ -97,14 +97,10 @@ def __init__(
...
-            max_power_charge if max_power_charge <= 0 else -max_power_charge
-        )
-        self.min_power_charge = (
-            min_power_charge if min_power_charge <= 0 else -min_power_charge
-        )
-        self.max_power_discharge = max_power_discharge
-        self.min_power_discharge = min_power_discharge
+        self.max_power_charge = -abs(max_power_charge)
+        self.min_power_charge = -abs(min_power_charge)
+        self.max_power_discharge = abs(max_power_discharge)
```
kim-mskw, line 102: in base.py you say these values are positive / negative, why here now abs?

kim-mskw (commented):
> This is implemented through the change in the line below:

`if price_forecast[start] >= average_price and max_power_discharge[start]:`

By adding the second clause, we only set a discharge bid, if we do have something to discharge ;) 

Yeah saw that as well ;) Hence deleted the comment :)

assume/strategies/flexable_storage.py
maurerle: Because we would waste a bidding possibility without any benefit.

As we can only bid either positive or negative, we always try to bid on what is more likely (by looking at the mean).
But if we can not bid the most likely, due to our storage constraints, it is best to bid the other direction instead of not doing anything?

assume/units/storage.py
maurerle: This enforces this hardly (due to the minus in front of the abs), to provide the proper behavior even in case of misconfigurations.

assume/strategies/flexable_storage.py
kim-mskw: But this would enforce the following behaviour in some cases:
We have a very high price of 300 €/MWh, and yet we cannot discharge due to technical constraints. We will bid for charging for 300 €/MWh.
Since we also place a price bid at the level of average_price...

I would opt for doing nothing in these cases.

assume/units/storage.py
kim-mskw: True!

kim-mskw:
For further discussions: This is the way it is implemented now

image

assume/strategies/flexable_storage.py
maurerle: Given our average is 150€/MWh and the efficiency is 0.8
We expect a price which is higher than average (300€/MWh).
We would like to discharge (and set a price of 150€/0.8 = 187.5€), but we can't.

So we should instead set a bid for charging, but do respect the average and efficiency as well:
We set 150*0.8=120€/MWh.

This is fine and not harmful, because we still rely on the average price and do not set `price_forecast[start] / efficiency` (I think you expected it that way)

kim-mskw (approved):

PR #431, opened 2024-10-08
Fix pypsa version

nick-harder:
The lates pypsa release is not compatible with our framework and causes problems.

-fix the version to the 0.30.3 before we update the framework 
-update the notebook 7 to use the correct version of pypsa

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -15,7 +15,7 @@ dependencies:
 - SQLAlchemy
 - matplotlib
 - pyomo
-- pypsa>=0.26.3
+- pypsa<=0.30.3
 - pyyaml
 
 - pip:
--- a/examples/notebooks/07_interoperability_example.ipynb
+++ b/examples/notebooks/07_interoperability_example.ipynb
@@ -54,7 +54,7 @@
     "# When running on your local machine, you can simply run pip install assume-framework[optimization]\n",
     "!pip install pyomo\n",
     "!apt-get install -y -qq glpk-utils\n",
-    "!pip install pypsa<=0.30.3\n",
+    "!pip install pypsa==0.30.3\n",
     "\n",
     "!pip install assume-framework"
    ]
--- a/examples/notebooks/07_interoperability_example.ipynb
+++ b/examples/notebooks/07_interoperability_example.ipynb
@@ -49,6 +49,13 @@
    },
    "outputs": [],
    "source": [
+    "# Install Pyomo, GLPK solver, and PyPSA\n",
+    "# This is needed only when running on Google Colab\n",
+    "# When running on your local machine, you can simply run pip install assume-framework[optimization]\n",
+    "!pip install pyomo\n",
+    "!apt-get install -y -qq glpk-utils\n",
+    "!pip install pypsa<=0.30.3\n",
+    "\n",
     "!pip install assume-framework"
    ]
   },
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ learning = [
 optimization = [
     "glpk >=0.4.7",
     "pyomo >=6.6.1",
-    "pypsa >=0.26.3",
+    "pypsa <=0.30.3",
 ]
 oeds = [
     "demandlib >=0.1.9",


maurerle (approved):
This is fine.

The warnings I am getting on 0.31.0 are:

```
DeprecatedWarning: import_components_from_dataframe is deprecated as of 0.29 and will be removed in 1.0. Use `n.add` instead. E.g. `n.add(class_name, df.index, **df)`.
  network.import_components_from_dataframe(buses, "Bus")
[assume/common/grid_utils.py:270](assume/common/grid_utils.py:270): DeprecatedWarning: import_components_from_dataframe is deprecated as of 0.29 and will be removed in 1.0. Use `n.add` instead. E.g. `n.add(class_name, df.index, **df)`.
  network.import_components_from_dataframe(lines, "Line")
[assume/common/grid_utils.py:87](assume/common/grid_utils.py:87): DeprecatedWarning: madd is deprecated as of 0.31 and will be removed in 1.0. Use `n.add` as a drop-in replacement instead.
  network.madd(
[assume/common/grid_utils.py:96](assume/common/grid_utils.py:96): DeprecatedWarning: madd is deprecated as of 0.31 and will be removed in 1.0. Use `n.add` as a drop-in replacement instead.
  network.madd(
[assume/common/grid_utils.py:108](assume/common/grid_utils.py:108): DeprecatedWarning: madd is deprecated as of 0.31 and will be removed in 1.0. Use `n.add` as a drop-in replacement instead.
  network.madd(
[assume/common/grid_utils.py:121](assume/common/grid_utils.py:121): DeprecatedWarning: madd is deprecated as of 0.31 and will be removed in 1.0. Use `n.add` as a drop-in replacement instead.
  network.madd(
````

switching to network.add gives `unhashable type DataFrame` so we have to look into this soon.

mthede:
Hi there, I'm currently working with pypsa in assume and because I was running into issues with the fixed version, I tried switching to the latest one (v0.32.0). For me, it worked without any problems. 

I also tested the tutorials 7, 10 and 11 in colab with an upgraded pypsa version. 
- 7: runs successfully
- 10: unsure, because running the simulation (Step 7.4) is currently not operational (neither in colab nor locally) - I will create an issue for that
- 11: missing installs in colab, but it's only function definitions anyway. no issues with func definitions locally with pypsa v0.32.0. 

I think in tutorials 10 and 11 `.madd()` can be changed to `.add()` and only argument "names" has to be adjusted to "name" as already done in grid_utils.py.

What do you think? Is there anything else to check?

maurerle:
Sounds good @mthede !
No I think that everything else should work. I think that the current pypsa behavior should be compatible to older versions (<v0.32) as well, otherwise we need to bump the minimum pypsa version in the pyproject.toml as well

mthede:
Okay great, I'm on it and will create a PR. 
To be sure: by "bumping the minimum pypsa version", do you mean setting it to "pypsa >=0.32.0"? Or back to the one (0.26.something) from before?

maurerle:
Currently we are at `pypsa <= 0.30.3` - I missed that, I thought we were already at >=0.32.

If everything works on new and older versions, it should be 
`pypsa >=0.26.3`
if it only works on newer versions, it should probably be `pypsa >=0.30.3` which should be save either way..?

mthede:
one issue I experienced with pypsa v0.30.3 was when loading an example with redispatch:

> in add_buses
>     network.add("Bus", buses.index, **buses)
>   File "[...]\Lib\site-packages\pypsa\components.py", line 923, in add
>     new_df.at[name, k] = typ(v) if typ != "geometry" else v
>                          ^^^^^^
>   File "[...]\Lib\site-packages\pandas\core\series.py", line 247, in wrapper
>     raise TypeError(f"cannot convert the series to {converter}")

[as far as I understood because bus["v_nom"] was a Series of float values which was supposed to convert to (single?) float and therefore throws the TypeError. but that must've been either a bug in pypsa or a (default) configuration error of network.buses? I still don't understand why it only happenend with loading a redispatch example with the assume pypsa loader and not other market_mechanism...]

I upgraded to v0.32.0 and the issue was resolved. 

not sure, if this is a relevant and reproducable problem, but if yes I'd say that we need to set pypsa>=0.32.0 ?

mthede:
Ah, I also tried example ["small_with_redispatch" ](https://github.com/assume-framework/assume/blob/2dbb529c999d7dc28676873160722b6df6611e6b/examples/examples.py#L43) with pypsa v0.30.3: it causes the same error as above!
It's running smoothly in v0.32.0.

so we need to bump the minimum version to v0.32.0?

maurerle:
Yes sounds good :)

PR #443, opened 2024-10-09
Include Pyomo as requirement, switch to HIGHS, curation of pyproject toml and dependencies

maurerle:
* Update reuse format
* improve pypi urls in pyproject.toml for better project description
* update dependency versions, as pyomo is now compatible with numpy 2
* remove unused dependencies yaml-include and glpk


nick-harder:
@maurerle what solver would be used by default instead of glpk? Maybe we do as pypsa and use the new and nice hyggs?

maurerle:
we are using glpk as default, but we do not need the pyglpk bindings for that, as pyomo has its own interface to glpk :)

nick-harder:
> we are using glpk as default, but we do not need the pyglpk bindings for that, as pyomo has its own interface to glpk :)

yes, I understand that, pyomo has the interface, but we need some solver so people can run optimization parts of ASSUME. I believe simply installing pyomo doesn't install any solver with it, or am I wrong?

maurerle:
pyomo does bring the solver interface and detects which driver is installed.

Typically you are only `apt install glpk-utils` or `apt install coinor-cbc` away from installing a solver which works with pyomo/assume (or `brew install glpk`/`brew install cbc` on mac).

Or on windows:
* https://winglpk.sourceforge.net/
* download and extract binaries from https://github.com/coin-or/Cbc/releases
* and the Gurobi install (login required): https://www.gurobi.com/downloads/gurobi-software/

We could give hints to that in the Install docs, but this is highly system specific.

pyproject.toml
```diff
@@ -32,31 +32,29 @@ dependencies = [
...
     "pyyaml >=6.0",
     "pyyaml-include >=2.2a",
+    "pyomo >=6.8.0",
 ]
 
 [project.optional-dependencies]
 learning = [
     "torch >=2.0.1",
 ]
 optimization = [
```
nick-harder, line 50: at this point we might as well rename the optimization to network since it only installs pypsa extra, what do you think?

assume/markets/clearing_algorithms/complex_clearing_dmas.py
```diff
@@ -27,7 +27,7 @@
 
 logger = logging.getLogger(__name__)
 
-SOLVERS = ["glpk", "cbc", "gurobi", "cplex"]
+SOLVERS = ["appsi_highs", "glpk", "cbc", "gurobi", "cplex"]
```
nick-harder, line 30: maybe we should stick to the same order of solvers for now among files before we have a global variable for that?

assume/strategies/dmas_powerplant.py
```diff
@@ -32,7 +32,7 @@
 logger = logging.getLogger(__name__)
 
 
-def get_solver_factory(solvers_str=["glpk", "cbc", "gurobi", "cplex"]):
+def get_solver_factory(solvers_str=["appsi_highs", "glpk", "cbc", "gurobi", "cplex"]):
```
nick-harder, line 35: also here, maybe use [highs, gurobi, glpk, the rest]

nick-harder (changes requested):

nick-harder:
@maurerle I have now switched to HIGHS by default and removed GLPK completely including git workflows. To supress the outputs from HIGHS I included solver options, but I still had to include getlogger to warning level for pyomo. I still have one problem though, namely constant one line prints from HIGHS when used with pypsa for example in small_redispatch example. I couldn't figure out how to solve that. Any suggestions?

nick-harder:
@maurerle I managed to remove the annoying logging from highs when used with pypsa. Could you please take a general look at the code and merge it if it is fine with you?

nick-harder (approved):

PR #449, opened 2024-10-17
Add residential componets and convert DSTs to classes

nick-harder:
# Pull Request

## Description
The DST components before were defined as functions, which is not very practical for the modularity perspective of the framework. This commit converts them all into classes for better reusability and future improvement. Also, new residential DST components from @HafnerMichael are refined and added to the world. DST components now also include improved docstrings for easier navigation and usability, as well as provide the description of the underlying optimization.

## Changes Proposed
-convert DST to classes for modularity and future usability
-add residential DST components like PV, EV, Heat Pump, Boiler
-create generic storage class and classes such as EV or hydrogen storage inherit from it
-add new tests for the new components
-fix old tests for the new structure
-refactor code for better readibility
-improve docstrings for DST components
-add dst_componnets to the world so that new components can be easily added to the world
-adjust notebook 10 to account for new structure

## Testing
All tests including new tests pass. All simulations work.

## Checklist


assume/units/dst_components.py
```diff
@@ -2,782 +2,1592 @@
...
-    @model_part.Constraint(time_steps)
-    def min_down_time_dri_constraint(b, t):
-        if t == 0:
-            return pyo.Constraint.Skip
+            @model_part.Constraint(self.time_steps)
+            def availability_pv_constraint(b, t):
+                """
+                Ensures the PV operates only during available periods.
+                """
+                return b.power[t] <= self.availability_profile[t] * b.max_power
```
HafnerMichael, line 684: Why not `==` to force power to be the available one? Or do you think about a controlled PV system where the power can be regulated?

assume/common/base.py
```diff
@@ -691,6 +691,43 @@ def calculate_ramp_charge(
         return power_charge
 
 
+class BaseDSTComponent:
```
maurerle, line 694: This behaves fundamentally different than the other BaseUnit types.
I would rather move this to dsm_load_shift.py?

assume/units/dsm_load_shift.py
```diff
@@ -6,6 +6,44 @@
...
+                                        the value is a dictionary of parameters for the respective technology.
+                                        Each technology is mapped to a corresponding class in `demand_side_components`.
+
+        The method:
+        ----------
+        - Looks up the corresponding class for each technology in `demand_side_components`.
+        - Instantiates the class by passing the required parameters.
+        - Adds the resulting block to the model under the `dsm_blocks` attribute.
+        """
+        from assume.units import demand_side_components
```
maurerle, line 28: This shadows a circular import, we should move the dictionary of available demand_side_components to something like `assume.units.optimization` ?

maurerle (commented):
The model_part classes still do not have any state and only consist of one static method, which was why I initially converted them to functions in #390.

It is fine for me to change this to classes.

In general, I have the feeling that the Steelplant misses a little generalization.
The DSMFlex class looks like one could just add different DSM units to it, but the actual implementation in steelplant exactly expects one electrolyser, one eaf, one dri_plant and an optional hydrogen_storage.

The DSMFlex class also expects to have exactly one electrolyser, eaf and dri_plant and as well shadows a circular import by having an import in the middle of the file.

I would like to go into a direction of having `assume.units.optimization` in which one can create an optimization module which as well integrates with multiple units of the same type as well and does also give working results if for example only an electrolyser is present.

The tests also only include prices of 1 and therefore do not check actual incentives and edge cases.

But thats just general input towards the DST units, at the current state I think thats fine.

But I think that this is something

assume/units/dst_components.py
```diff
@@ -2,782 +2,1592 @@
...
+            self.time_steps,
+            within=pyo.NonNegativeReals,
+            bounds=(0, self.max_power_charge),
+            doc="Charging power at each time step",
+        )
+        model_part.discharge = pyo.Var(
+            self.time_steps,
+            within=pyo.NonNegativeReals,
+            bounds=(0, self.max_power_discharge),
+            doc="Discharging power at each time step",
```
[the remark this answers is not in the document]
nick-harder: if everyhting is configured correctly, namely the efficiencies and prices, this won't be happening. but if something is configured wrong, like charging is and discharging loses no energy, in that case it can happen. But I don't see it worth the transition to a MILP just to fix someone's bad setup :-D

assume/units/dst_components.py
nick-harder: if we use == we don't allow any flexibility in the system. using this form the household can also deliver flexibility in the future by curtailing PV

examples/notebooks/assume-repo
maurerle, line 1: This repository should not be added ;)

HafnerMichael (approved):
Looks fine to me, all base functionalities got implemented

nick-harder:
@HafnerMichael @maurerle @Manish-Khanra I have fixed several wrong behaviors in the DST components, the updated tests covering a wide functionality checks are now also commited. Everything looks correct now. I have also added operational_cost variables to all residential DSTs except for those storage based, since it can use different prices. It would be great if someone could also take a look at this again. @maurerle regarding the modularity discussion for the DST components, we would need to discuss that with Manish in more details to be able to change the way components are defined and used. Something for the next time.

maurerle (approved):

assume/units/dst_components.py
```diff
@@ -2,782 +2,1593 @@
...
+        def max_power_pv_constraint(b, t):
+            """
+            Ensures the power output of the PV unit does not exceed the maximum power limit.
+            """
+            return b.power[t] <= b.max_power
+
+        # Operating costs
+        @model_block.Constraint(self.time_steps)
+        def operating_cost_constraint_rule(b, t):
+            return b.operating_cost[t] == b.power[t] * model.electricity_price[t]
```
HafnerMichael, line 665: In your comment you mentioned that you did not implement cost constraints for the storage based components (regarding different prices) what if the power out price also differs? Or did you implement it to serve as a base functionality and if it differs it needs to be handled wherever this component gets used?

assume/units/dst_components.py
```diff
@@ -2,782 +2,1593 @@
...
-    model_part.max_capacity = pyo.Param(initialize=max_capacity)
-    model_part.min_capacity = pyo.Param(initialize=min_capacity)
-    model_part.initial_soc = pyo.Param(initialize=initial_soc)
-    model_part.storage_loss_rate = pyo.Param(initialize=storage_loss_rate)
-    model_part.charge_loss_rate = pyo.Param(initialize=charge_loss_rate)
-    model_part.discharge_loss_rate = pyo.Param(initialize=discharge_loss_rate)
-
-    # define variables
-    model_part.soc = pyo.Var(time_steps, within=pyo.NonNegativeReals)
-    model_part.uniformity_indicator = pyo.Var(time_steps, within=pyo.Binary)
```
Manish-Khanra, line 606: @nick-harder @HafnerMichael We need model_part.uniformity_indicator = pyo.Var(time_steps, within=pyo.Binary) and the corresponding constraints (which were removed) to logically enforce that storage cannot both charge and discharge simultaneously. The binary variable uniformity_indicator ensures that this behavior is prevented. Allowing the storage unit to charge and discharge simultaneously is not physically feasible and contradicts real-world operational constraints.

assume/units/steel_plant.py
```diff
@@ -17,29 +17,13 @@
...
-    create_electric_arc_furnance,
-    create_electrolyser,
-    create_hydrogen_storage,
-)
 
 SOLVERS = ["gurobi", "glpk", "cbc", "cplex"]
 
 logger = logging.getLogger(__name__)
 
-# Mapping of component type identifiers to their respective classes
```
Manish-Khanra, line 32: @nick-harder The original idea was to use this dictionary of technologies to be unique to each plant type or demand-side agent (e.g., only technologies compatible with a steel plant, cement plant, or buildings). The idea is to exclude technologies that aren't relevant to the specific agent types, ensuring that only compatible technologies are modeled for each.

assume/units/steel_plant.py
```diff
@@ -162,30 +146,12 @@ def switch_to_flex(self, instance):
 
         return instance
 
-    def initialize_components(self, components: dict[str, dict]):
```
Manish-Khanra, line 165: @nick-harder Why the initialize_components function is removed? Won't it still be necessary to properly initialize the components of the steel plant? This function ensures that the relevant components are added to the Pyomo model based on the dictionary of technologies (dst_components).

Manish-Khanra (changes requested):

assume/units/dst_components.py
nick-harder: @HafnerMichael yes, this is the diea, that you specify which prices to use in the global energy balance constraint. and the whole system will optimize to that.

assume/units/dst_components.py
```diff
@@ -2,782 +2,1593 @@
...
+            ramp_down=self.ramp_down,
+            time_steps=self.time_steps,
+        )
 
-    # Ramp-up constraint
-    @model_part.Constraint(time_steps)
-    def ramp_up_constraint(b, t):
-        """
-        Limits the rate at which the power input to the electrolyser can increase.
+        # Define additional variables and constraints for startup/shutdown and operational status
```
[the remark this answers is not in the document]
nick-harder: @Manish-Khanra yes, I saw that, but the operation was not producing correct results, so these additional binaries were required. But if you noticed, the binaries and the corresponding constraints are created only when the parameters are actually specified such as min_downtime_steps > 1. And by default they are set to 1, so these constraints are not created and the model is still LP. I hope this is fine with your vision

assume/units/dst_components.py
nick-harder: @Manish-Khanra I would disagree. When the storage is configured properly, namely the charging and discharging efficiencies are specified, it would not charge and discharge at the same time as this would not be optimal behavior. Also, even if this behavior would occur for a battery storage, this would mean a simple financial flow and not physical and the final energy delivered or purchased would be the sum of both flows. If you would like, we could implement this constraint for the DRI and hydrogen storages as additional constraints

assume/units/steel_plant.py
nick-harder: @Manish-Khanra I moved this function to the common class since it is where it belongs. But you are right, we need to have a check inside the steel plant itself for the components provided in the component list, if it makes sense or not.

nick-harder: @Manish-Khanra I have now included required_technologies and optional_technologies to the steel plant so that the components are now checked to containt all required ones and not contain any extra ones.

assume/units/dst_components.py
nick-harder: @Manish-Khanra I have now added the additional constraint to the DRI and hydrogen storage technologies.

Manish-Khanra (approved):

assume/units/dst_components.py
```diff
@@ -2,782 +2,1617 @@
...
+        # Operating costs
+        @model_block.Constraint(self.time_steps)
+        def operating_cost_constraint_rule(b, t):
+            return b.operating_cost[t] == b.power_in[t] * model.electricity_price[t]
+
+        # Ramp-up constraint and ramp-down constraints
+        add_ramping_constraints(
+            model_block=model_block,
+            ramp_up=self.ramp_up,
+            ramp_down=self.ramp_down,
```
HafnerMichael, line 130: @nick-harder just saw it now during household adaption. Is it on purpose to use the variables self.ramp_* instead of the defined pyo.Params? If not those function parameter are not needed since they are already in the block.
I also saw also other constraints all over the file, where just the variable was used.

assume/units/dst_components.py
```diff
@@ -2,782 +2,1617 @@
...
+                - (1 / b.efficiency_discharge) * b.discharge[t]
+                - b.storage_loss_rate * prev_soc
+            )
+
+        # Apply ramp-up constraints if ramp_up is specified
+        @model_block.Constraint(self.time_steps)
+        def charge_ramp_up_constraint(b, t):
+            if t == self.time_steps.at(1):
+                return b.charge[t] <= self.ramp_up
+            return b.charge[t] - b.charge[t - 1] <= self.ramp_up
```
HafnerMichael, line 510: Like here

assume/units/dst_components.py
nick-harder: @HafnerMichael oh yeah true, just noticed that myself. At the beginning I didn't have the class variables, so I used the model ones. And also the idea of having the class variables is just in case we will have some new functions added to the dst classes in the future, for example to change some consraints to adapt the operating regime. And I didn't remove the model parameters. What do you think would be the better approach to avoid having duplicates? @Manish-Khanra

PR #453, opened 2024-10-22
Update mango to 2.x

maurerle:
With the release of mango-agents 2.x a lot of the async requirements are less strict, so we can use the new syntax which makes it easier to refer to agents and create simulations.

* one does not need to have a running container when creating the simulation anymore
* Agents can be created without a container now as well. This means that the agent does not have an aid until it is added to a container (where we can set `suggested_aid` now).
* we do not need the async wrapper functions anymore


maurerle:
I added the release notes.

My tests include:
1. having run pytest
2. running `assume -s example_01a -c tiny -db "postgresql+psycopg2://assume:assume@localhost:5432/assume"`
3. look at dashboard to see that units report values, and metadata is correctly sent to the output role
4. running `assume -s example_02a -c tiny -db "postgresql+psycopg2://assume:assume@localhost:5432/assume"` to make sure learning works
5. taking a look at the learning dashboard to see progress
6. running `assume -s example_03 -db "postgresql+psycopg2://assume:assume@localhost:5432/assume"`
7. running my amiris benchmark again
8. running the oeds integration again

Looks fine :)

maurerle:
I also offer myself to merge the changes into other branches to fix merge conflicts :laughing:

nick-harder (approved):

PR #459, opened 2024-10-23
Modified steel plant

nick-harder:
A few adjustments to the way the steel plant is defined and the DSTm components are formulated:
-use uniformly model_blick.Pram referencing
-use electrolyser as optional component in steel plant
-adjust process for steel plant when electrolyser is given
-use zero hydrogen price if electrolyser is given and get hydrogen price if not


assume/units/dst_components.py
Manish-Khanra, line 1: CO2 emission constraint included

assume/units/steel_plant.py
Manish-Khanra, line 1: Previously, `hydrogen_price` was not constructed if the electrolyser was absent, causing an AttributeError during for the as ewhen `fuel_type= 'both'`. The issue has been fixed!

assume/units/steel_plant.py
```diff
@@ -136,75 +144,94 @@ def __init__(
...
+                        self.model.dsm_blocks["electrolyser"].hydrogen_out[t]
+                        + self.model.dsm_blocks["hydrogen_storage"].discharge[t]
+                        == self.model.dsm_blocks["dri_plant"].hydrogen_in[t]
+                        + self.model.dsm_blocks["hydrogen_storage"].charge[t]
+                    )
+                else:
+                    return (
+                        self.model.dsm_blocks["electrolyser"].hydrogen_out[t]
+                        == self.model.dsm_blocks["dri_plant"].hydrogen_in[t]
+                    )
```
Manish-Khanra, line 235: The equation ensuring hydrogen flow from grid to the dri plant, when electrolyser is not the part of the system, was missing. The issue has been fixed!

assume/units/dsm_load_shift.py
Manish-Khanra, line 1: added conditional statement for the same when electrolyser is not the part of the system

Manish-Khanra (approved):
After fixing some issues, the steel plant has been thoroughly tested with different fuel types. Everything works fine now!

Manish-Khanra:
@nick-harder natural gas co2 and lime co2 factors moved from model to dst_components. These parameters are now the part of industrial dsm units.

PR #478, opened 2024-11-09
Clear the simulation in the db on startup

HafnerMichael:
# Pull Request

## Related Issue
Closes #477 

## Description
Since the db assignment got changed, the clearing did not work anymore.

## Changes Proposed
- Moved the scenario clearing in the `on_ready() `method
-
-

## Testing
[Describe the testing you've done, including any specific test cases or scenarios]

## Checklist
Please check all applicable items:

- [ ] Code changes are sufficiently documented (docstrings, inline comments, `doc` folder updates)
- [ ] New unit tests added for new features or bug fixes
- [ ] Existing tests pass with the changes
- [ ] Reinforcement learning examples are operational (for DRL-related changes)
- [ ] Code tested with both local and Docker databases
- [ ] Code follows project style guidelines and best practices
- [ ] Changes are backwards compatible, or deprecation notices added
- [ ] New dependencies added to `pyproject.toml`
- [ ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included
- [ ] Consent to release this PR's code under the GNU Affero General Public License v3.0

## Additional Notes (if applicable)
[Any additional information, concerns, or areas you want reviewers to focus on]

## Screenshots (if applicable)
[Add screenshots to demonstrate visual changes]

PR #479, opened 2024-11-10
create patch release for version 0.4.3

maurerle:
This should be merged on monday.
This patch release fixes removing previous runs from the database.

see #478

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -13,8 +13,18 @@ Upcoming Release
   The features in this section are not released yet, but will be part of the next release! To use the features already you have to install the main branch,
   e.g. ``pip install git+https://github.com/assume-framework/assume``
 
+v0.4.3 - (11th November 2024)
+===========================================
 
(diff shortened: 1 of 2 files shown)


nick-harder:
@maurerle should this be a whole new release or just a post1 release?

maurerle:
Should be a new release. That's what patch releases are for. By that, the former release should at least have been an update to 0.5.0 at least, as it did include features:
https://semver.org/

But I think that our glorious userbase will forgive our past inconsistent version schema :D

nick-harder:
> Should be a new release. That's what patch releases are for. By that, the former release should at least have been an update to 0.5.0 at least, as it did include features: https://semver.org/
> 
> But I think that our glorious userbase will forgive our past inconsistent version schema :D

ohh good to know, thanks! yeah I was thinking if it should have been 0.5 or not, but there were not that many new features to justify it somehow.

nick-harder (approved):

PR #483, opened 2024-11-15
Pandas wrapper

nick-harder:
# Pull Request

## Related Issue
Closes #[issue number] (if applicable)

## Description
As suggested by @maurerle this improves performance of all simulations by a factor of 2 up to 4, by replacing pandas actions with numpy where possible.

For this, we are using a wrapper object FastDatetimeSeries which wraps a numpy array so that we can access it using typical datetime accessors

## Changes Proposed
- Shift to using special classes of FastIndex and FastSeries
- Adjust the rest of the code to make use of this new class

## Testing
[Describe the testing you've done, including any specific test cases or scenarios]

## Checklist


nick-harder:
@maurerle all tests pass except for one in unit_operator def test_get_actual_dispatch. I don't understand why it doesn't work, could you please take a look into it?

maurerle:
Somehow the other tests for RL are not yet working @nick-harder

nick-harder:
@maurerle thanks for the fix, I have also fixed the RL tests now

assume/common/forecasts.py
```diff
@@ -82,6 +84,21 @@ def get_price(self, fuel_type: str) -> pd.Series:
...
+
+        Returns:
+            FastSeries: The converted FastSeries.
+        """
+        if isinstance(value, pd.Series):
+            value = value.values  # Use the values as an array for consistency
+        return FastSeries(index=self.index, value=value, name=name)
+
 
 class CsvForecaster(Forecaster):
```
kim-mskw, line 103: Could that be general? So from df to fastseries dictionary? Does not need to be related to forecaster.

assume/strategies/advanced_orders.py
```diff
@@ -138,7 +138,7 @@ def calculate_bids(
 
             # Flex-bid price formulation
             if op_time <= -unit.min_down_time or op_time > 0:
-                bid_quantity_flex = max_power[start] - bid_quantity_inflex
+                bid_quantity_flex = max_pwr - bid_quantity_inflex
```
kim-mskw, line 141: Why shorten the variable name? I feel this is not as self explanatory

kim-mskw:
I do not find it too confusing or inconvenient; it looks rather arbitrary. However, contrary to what @nick-harder  said in the development meeting, I would change all the Args Docsting to explicitly state that it is a fast series. If I encounter unexpected behaviour because I think it is a series, the information that it is not should be as close to me as possible.

assume/strategies/advanced_orders.py
nick-harder: max_power is an array, we cannot reuse the name again. we would need to rename max_power to max_power_array in the first place or use a different name here

assume/common/forecasts.py
```diff
@@ -82,6 +84,21 @@ def get_price(self, fuel_type: str) -> pd.Series:
 
         return self[f"fuel_price_{fuel_type}"]
 
+    def _to_fast_series(self, value, name: str) -> FastSeries:
```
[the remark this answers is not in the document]
kim-mskw: No, I meant to ask if they could be in utils. I see that this is currently just used in the forecaster. Yet this function seems very general, and it could be helpful in various locations in the future. If it is not, please indicate why.

assume/strategies/advanced_orders.py
nick-harder: I mean this is just one way of iterating over a list. we can of course agree to rename the lists and iterate over them and call the variables something else. Please make a suggestion what makes more sense from your point of view

assume/common/forecasts.py
nick-harder: replaced with a method from FastSeries itself

nick-harder:
@maurerle I am working on tests and updates, please wait

assume/common/units_operator.py
```diff
@@ -248,7 +248,9 @@ def handle_data_request(self, content: DataRequestMessage, meta: MetaDict) -> No
 
         data = []
         try:
-            data = self.units[unit].outputs[metric_type][start:end]
+            data = (
+                self.units[unit].outputs[metric_type].as_pd_series(start=start, end=end)
+            )
```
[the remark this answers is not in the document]
nick-harder: this is an internal process of ASSUME, so hopefully this is fine like this since changing it would need to call other functions from FAstSeries, which wouldn't make the situation clearer

assume/strategies/advanced_orders.py
nick-harder: I have changed the naming everywhere to max_power_values and max power in the iteration. this should avoid any confusions

nick-harder:
@kim-mskw @maurerle I have tested all non learning examples and everythings works fine, all data is also being properly written to the database and other outputs. I have also replaced and removed pandas from the code as much as possible and renamed the typing hints and other things. Please take another look at the commit

nick-harder:
@maurerle I have been testing now with the largest example we have for 2019 with two EOM and CRM markets, and the speed up is crazy, I didn't expect that. The difference is around 4.2 times. We should really all put a bit more effort and merge this as soon as we can after proper testing, this performance increase is even more what we expected...

running on pandas_wrapper
telegram-cloud-document-2-5332300261325037747

running on main
telegram-cloud-document-2-5332300261325038181

nick-harder:
@kim-mskw I have tested the RL examples we have. Some things had to be fixed byt everything is working and the learning in small 1 and 2 is also like 2 to 3 times faster

maurerle (approved):

kim-mskw:
> the review! @kim-mskw would you have time to take another look at the PR if everything is good from your side and looks fine?

Of course :) on it!

assume/scenario/loader_csv.py
```diff
@@ -563,14 +564,14 @@ def setup_world(
...
-    index = scenario_data["index"]
     powerplant_units = scenario_data["powerplant_units"]
     storage_units = scenario_data["storage_units"]
     demand_units = scenario_data["demand_units"]
     dsm_units = scenario_data["dsm_units"]
     forecaster = scenario_data["forecaster"]
 
-    save_frequency_hours = config.get("save_frequency_hours", 48)
+    # save every thousand steps by default to free up memory
+    save_frequency_hours = config.get("save_frequency_hours", 1000)
```
kim-mskw, line 574: Did we check the writing of dfs in the output role with a high save_frequency?

nick-harder: @maurerle thanks for the analysis! But 350mb compared to 500 mb, so 150mb extra is a low price to price for a significant speed up. In my opinion this change to longer periods is justified. I don't think any more or less modern system would run out of memory because of this.

maurerle: On my machine it does not give a significant speed up by reducing this param. How much is it on yours?
Typically, the save_dfs function should not block anyway.

The problem is, that this increases linearly, I just stopped the simulation after some time.

nick-harder: @maurerle try activating the saving to csv files, this will change the timing for you

nick-harder: @maurerle also as of now on pandas_wrapper the config for example_03 has the save frequency as null, so it is never saved

maurerle: @nick-harder CSV does only save at the end as `save_frequency_hours is disabled due to CSV export being enabled. Data will be stored in the CSV files at the end of the simulation.`

Which gives above graphs as well when setting `save_frequency_hours: 96` in the yaml.
So I wonder how you see a speedup with this setting

maurerle: Which we did as `unit_dispatch` changes in the CSV files when the accepted_* fields are added for example.
so eventually we should fix this as I do not really like to have linearly increasing memory consumption for a simulation tool :)

Saving csv is indeed a little slower, when the above mentioned handler to switch back to None is removed.

tests/test_units_operator.py
```diff
@@ -243,24 +244,28 @@ async def test_get_actual_dispatch(units_operator: UnitsOperator):
...
     clock.set_time(clock.time + 3600)
-    last = clock.time - 3600
 
     market_dispatch, unit_dfs = units_operator.get_actual_dispatch("energy", last)
-    assert unit_dfs[0].index[0].timestamp() == clock.time
-    assert len(unit_dfs[0]) == 1
+    assert datetime2timestamp(unit_dfs[0]["time"][0]) == last
+    assert datetime2timestamp(unit_dfs[0]["time"][1]) == clock.time
+    assert len(unit_dfs[0]["time"]) == 2
     assert len(market_dispatch) == 0
```
maurerle, line 271: Why did the tests change here to write two timestamps?

We should have seen this, as this caused #539

PR #491, opened 2024-11-21
fix tqdm output breaking redispatch simulation on windows

maurerle:
starting with linopy 0.4.3 it is possible to suppress the tqdm log output of linopy

This caused an issue seen by @Manish-Khanra as tqdm on windows did not like the plumbing of the output and error stream which is done by us to suppress all stdout output from the network optimize call.

I would leave this as a draft until https://github.com/PyPSA/linopy/pull/375 is merged and released


nick-harder (approved):

maurerle:
The upstream issue of this output redirection just got fixed in tqdm :D
https://github.com/tqdm/tqdm/issues/1537#issuecomment-2462116548

But I am happy that we removed the need of output redirection in assume by fixing the need for it in linopy :blush:

PR #492, opened 2024-11-21
fix missing database check when doing RL without a given database

maurerle:
This fixes commands like:

`assume -s example_02a -c tiny` which did throw: `AttributeError: 'NoneType' object has no attribute 'begin'` before


kim-mskw:
Interesting I'll review it @nick-harder

kim-mskw (approved):

PR #505, opened 2024-11-29
Fix a huge slowdown in flexABLE startegies

nick-harder:
# Pull Request

## Related Issue
Closes #[issue number] (if applicable)

## Description
After the latest commit with the FastSeries I have noticed a huge slow down for long and large simulations. Here I took the effort to find the reasons and came upon multiple places. At the end the problem is now fixed and the speed up for large simulations compared to current main is 4x. This results in a total speed up of 10x compared to the last stable release fr large simulations and 8x for small simulations.

## Changes Proposed
-change how average operation time is calculated
-implement static methods for loc and at methods in fast series
-do not create FastIndex constantly, instead use the slice
-for demand unit, compute the volume for the specific index instead of the whole index
-revise calculate_reward functions for flexable startegies to remove unrequired parameters and calculations

## Testing
[Describe the testing you've done, including any specific test cases or scenarios]

## Checklist


assume/common/base.py
```diff
@@ -139,25 +142,51 @@ def set_dispatch_plan(
         Iterates through the orderbook, adding the accepted volumes to the corresponding time slots
         in the dispatch plan. It then calculates the cashflow and the reward for the bidding strategies.
 
+        Additionally, updates the average operation and downtime dynamically.
```
maurerle, line 145: As this is something very different, I would like to have this part in a separate function, to maintain the right scope.

The downtime was removed - so should be removed from comment as well

Probably it would also be better to call the function then in `execute_current_dispatch`, to not have the same problem with multi market bidding as well..?

maurerle (commented):
I don't like to always calculate the average if it is only used if one uses this exact bidding strategy, but that's fine, I guess.
I made a different approach in #506 but it does not calculate the correct values so its not useful.
It's in an interesting parameter anyway.

Looks good!

assume/common/base.py
nick-harder: @maurerle removed it from base into calculate_reward, which is fitting

maurerle (approved):

assume/strategies/flexable.py
```diff
@@ -601,3 +604,24 @@ def get_specific_revenue(
...
+        end: The end index of the slice being updated (inclusive).
+    """
+    # Get the current slice of outputs
+    current_slice = unit.outputs[product_type].loc[start:end]
+
+    # Increment total operation time for operating periods in the slice
+    unit.total_op_time += (current_slice > 0).sum()
+
+    # Update the average operation time
+    total_periods = len(unit.index[:end]) + 1  # Total periods up to and including 'end'
```
maurerle, line 626: as the index of `end` is the count from zero towards it, we could just use the index, though this is more declarative what we are doing here and stays compatible to pandas Series

PR #508, opened 2024-12-03
Min max scaling for observation space

kim-mskw:
# Pull Request

## Description
To Include more robust observation space scaling, a min-max scaling is proposed instead of the formerly introduced max scaling. 

## Changes Proposed
- min/max scaling in learning strategies and advanced learning strategies
- not in tutorials, as we want to stick to an easy scaling introduction and not over-engineer it there
- changes of actual scaling values are minor

## Testing
With example 02a tiny

## Checklist


assume/strategies/learning_strategies.py
```diff
@@ -1006,19 +1035,30 @@ def create_observation(
...
         ):
-            scaled_res_load_forecast = (
-                unit.forecaster[f"residual_load_{market_id}"].loc[start:]
-                / scaling_factor_res_load
+            scaled_res_load_forecast = min_max_scale(
+                unit.forecaster[f"residual_load_{market_id}"].loc[start:],
+                lower_scaling_factor_res_load,
+                upper_scaling_factor_res_load,
             )
             scaled_res_load_forecast = np.concatenate(
```
kim-mskw, line 1063: Why no scaling in this part?

kim-mskw: I confusingly marked these lines, but I meant the ones below

assume/strategies/learning_advanced_orders.py
```diff
@@ -264,17 +265,34 @@ def create_observation(
         # =============================================================================
         # 1.1 Get the Observations, which are the basis of the action decision
         # =============================================================================
-        scaling_factor_res_load = self.max_demand
 
+        # defines bounds of observation space
+        # stays here as it is unit specific, and different forecasts might apply for different units
+        # different handling would require an extra unit loop at learning role intiliazation and unit specific max/min values
+        # further forecasts might change during the simulation if advanced forecasting is used
+        self.max_market_price = max(unit.forecaster[f"price_{market_id}"])
```
nick-harder, line 273: this can be done in __init__ so it is done only once and not during every call of the function

assume/strategies/learning_advanced_orders.py
```diff
@@ -264,17 +265,34 @@ def create_observation(
...
+        self.max_residual = max(unit.forecaster[f"residualy_load_{market_id}"])
+        self.min_residual = min(unit.forecaster[f"residual_load_{market_id}"])
+
+        # residual load forecast
+        upper_scaling_factor_res_load = self.max_residual
+        lower_scaling_factor_res_load = self.min_residual
+        
         # price forecast
-        scaling_factor_price = self.max_bid_price
+        upper_scaling_factor_price = self.max_market_price
```
nick-harder, line 283: why assign new variable here if we can use self.values directly?

assume/strategies/learning_advanced_orders.py
```diff
@@ -320,18 +342,30 @@ def create_observation(
         current_costs = unit.calculate_marginal_cost(start, current_volume)
 
         # scale unit outputs
-        scaled_max_power = current_volume / scaling_factor_total_capacity
-        scaled_marginal_cost = current_costs / scaling_factor_marginal_cost
+        scaled_max_power = min_max_scale(
```
nick-harder, line 345: same here, should be calculated only once in init

docs/source/release_notes.rst
```diff
@@ -36,6 +36,8 @@ Upcoming Release
 - **Outputs Role Performance Optimization:** Output role handles dict data directly and only converts to DataFrame on Database write.
 - **Overall Performance Optimization:** The overall performance of the framework has been improved by a factor of 5x to 12x
   depending on the size of the simulation (number of units, markets, and time steps).
+- **Learning Opservation Space Scaling:** Instead of the formerly used max sclaing of the observation space, we added a min-max scaling to the observation space.
```
nick-harder, line 39:
```suggestion
- **Learning Observation Space Scaling:** Instead of the formerly used max scaling of the observation space, we added a min-max scaling to the observation space.
```

assume/strategies/learning_strategies.py
nick-harder: oh, this is a mistake, this should be fixed

assume/strategies/learning_strategies.py
```diff
@@ -1006,19 +1035,30 @@ def create_observation(
...
         # 1.1 Get the Observations, which are the basis of the action decision
         # =============================================================================
-        # scaling factors for the observations
-        scaling_factor_res_load = self.max_demand
-        scaling_factor_price = self.max_bid_price
+        # defines bounds of observation space
+        # stays here as it is unit specific, and different forecasts might apply for different units
+        # different handling would require an extra unit loop at learning role intiliazation and unit specific max/min values
+        # further forecasts might change during the simulation if advanced forecasting is used
+        self.max_market_price = max(unit.forecaster[f"price_{market_id}"])
```
nick-harder, line 1042: we should also here compute these things in the init directly to save time. Also no need to assign values from self like upper_scaling_factor_res_load = self.max_residual since we can use directly the self. values

assume/strategies/learning_advanced_orders.py
```diff
@@ -10,6 +10,7 @@
 from assume.common.base import SupportsMinMax
 from assume.common.market_objects import MarketConfig, Orderbook, Product
 from assume.common.utils import get_products_index
+from assume.reinforcement_learning.learning_utils import min_max_scale
```
nick-harder, line 13: I don't see this as really a learning util, but just a simple util

nick-harder (changes requested):
good suggestion and a nice catch with the mistake not scaling some values. I have left some comments to improve the performance and how things are handled

assume/strategies/learning_advanced_orders.py
kim-mskw: I disagree for the reasons stated in the lengthly comment

docs/source/release_notes.rst
kim-mskw: I thought the new precommit would fix something like this?

assume/strategies/learning_strategies.py
kim-mskw: I disagree for the reasons stated in the lengthly comment

assume/strategies/learning_advanced_orders.py
kim-mskw: We did the same beforehand to make this crucial part of the learning easily changeable for new users; compare the RL tutorial. We could change it if you think it would be easier to understand.

assume/strategies/learning_advanced_orders.py
nick-harder: @kim-mskw what lengthy comment? Don't see it

kim-mskw: I argue that the market price max might be different for different units in the future, hence it should be unit specific as well 

# stays here as it is unit specific, and different forecasts might apply for different units
# different handling would require an extra unit loop at learning role intiliazation and unit specific max/min values
# further forecasts might change during the simulation if advanced forecasting is used

kim-mskw: For the sake of performance, we decided to sacrifice the accommodation of future developments. Hence, we move the scaling up to the init, where it is only called once.

kim-mskw:
I changed the setup to accommodate the changes proposed by @nick-harder.

@AndreasEppler, Somehow, the scaling does not output the intended values between 0 and 1, but the residual load forecast is still 200 something. Can you take a look at that? Also, please adjust the RL tutorials to handle this change properly.

docs/source/release_notes.rst
```diff
@@ -38,6 +38,8 @@ v0.5.0 - (10th December 2024)
...
```
nick-harder, line 41: since there has been a new release already, these should be moved up to upcoming release

nick-harder (changes requested):
@AndreasEppler there are some inconsistencies in regards of changes to where what is done. Please check everything once again

nick-harder (approved):

PR #509, opened 2024-12-04
various improvements for pypsa and amiris loader

maurerle:
We did have some problems with the Amiris and PyPSA loaders, which are about to be fixed in this PR.

This also fixes the deprecation warning in PyPSA.

When running a distributed simulation using MQTT, it helps to `set_tcp_nodelay=true` in the MQTT config to improve performance a lot.


maurerle:
This shows the actual power flows on the grid dashboard:
We should implement flows for redispatch and naive solution as well

https://github.com/user-attachments/assets/0a864832-31f1-40f5-ad56-a24d483bf925

(Sequence done with `grafanimate --scenario=demo.py:assume --output=./animations --use-panel-events --headless` and a demo.py file)

maurerle:
this is ready for review now :)

nick-harder (approved):

PR #515, opened 2024-12-09
delete nodal_pricing as complex_clearing is superior

maurerle:
This removes the nodal_pricing clearing as it does not solve correctly for different priced loads and does not work with the pypsa loader


assume/markets/clearing_algorithms/__init__.py
```diff
@@ -20,11 +20,10 @@
 
 # try importing pypsa if it is installed
 try:
-    from .nodal_pricing import NodalMarketRole
     from .redispatch import RedispatchMarketRole
 
     clearing_mechanisms["redispatch"] = RedispatchMarketRole
-    clearing_mechanisms["nodal"] = NodalMarketRole
+    clearing_mechanisms["nodal"] = ComplexClearingRole
```
nick-harder, line 26: I would completely remove nodal, as we don't really have nodal. the complex clearing is also not 100% nodal clearing as it ignores physical laws and is just a transportation problem

nick-harder (changes requested):

nick-harder (approved):

Issue #517, opened 2024-12-10
ambigous default episode count


maurerle:
the default for episode/eval_episode is set to 1 in `load_scenario_folder()` , although docs say "Defaults to 0."

_Originally posted by @mthede in https://github.com/assume-framework/assume/issues/512#issuecomment-2531064464_

nick-harder:
fixed in #519

Issue #512, opened 2024-12-09
world.output_role.db = None leads to attribute error in DRL examples


AndreasEppler:
Issue found by @mthede 
When running 04_reinforcement_learning_algorithm_example.ipynb or 09_example_Sim_and_xRL.ipynb, the following issue emerges:
![image](https://github.com/user-attachments/assets/ca775f09-6c32-4f1c-9a94-c00ecc743529)
The issue occurs when calling world.output_role.delete_simular_runs() or world.output_role.get_sum_reward()
This appears to be due to a recent change in the WriteOutput class, where db is now initialized as None.
![image](https://github.com/user-attachments/assets/dbca055a-c1d3-4865-a263-6d879f5a23c7)
Adding the if-clause
```
if world.output_role.db is not None:
        world.output_role.delete_similar_runs()
``` 
to the run_learning function has been successfully found to be a quick fix for the issue that obviously does not address the problem at its root.

maurerle:
Hi, I doubt that this error comes from an uninitialized db_uri - if you do not provide a DB_URI, there should not be any output_role - not a `has no attribute 'delete_similar_runs'`.

Did you check if this occurs on main as well?
I have the feeling, that the two mentioned examples are not working that well.
I don't have any folder `inputs/tutorial_09` locally for example..?

I fixed some issues for the output role in the attached PR though.

AndreasEppler:
Hi, thanks for looking into it.
The `inputs/tutorial_09` folder seems to be created and used in `10_DSU_and_flexibility.ipynb`.
`09_example_Sim_and_xRL.ipynb` uses the `inputs/tutorial_08` folder created by executing `08_market_zone_coupling.ipynb`. The execution of notebook 08 is done at the beginning of notebook 09, however, the line was currently only executed in google colab. I'll fix this and create a PR.

maurerle:
I think I understand now:

I renamed this function in https://github.com/assume-framework/assume/commit/d21a4e566d2c7c6c65555d6b76ae14c64a2e8683

So one must use the matching assume version from main today.
Or use the new release which will be available soon.

AndreasEppler:
I don't think the renaming of the function is what is causing the issue, as I have today's main assume version. It seems to be due to self.db being initialized as None instead of db_engine. Does thismake sense?

maurerle:
There is no way an object of type `WriteOutput` would not have the method `delete_similar_runs` with current main..

[...]

PR #519, opened 2024-12-10
Fix delete db scenario

maurerle:
Another attempt to fix the issue in 
https://github.com/assume-framework/assume/issues/512#issuecomment-2531064464

the first two commits are valid anyway, the other two commits are used to reproduce the mentioned issue (unsuccessfully on my side though)

The first two commits do:
* improve handling of assumeException (when the ovrwrite strategies question in learning is answered with a "n")
* do not access self.db before initializing the eninge in `on_ready`

Eventually this helps


examples/examples.py
```diff
@@ -107,7 +107,7 @@
     data_format = "local_db"  # "local_db" or "timescale"
 
     # select the example to run from the available examples above
-    example = "small"
+    example = "small_learning_1"
```
nick-harder, line 110: this should not change?

assume/scenario/loader_csv.py
```diff
@@ -728,7 +728,7 @@ def load_scenario_folder(
     study_case: str,
     perform_evaluation: bool = False,
     terminate_learning: bool = False,
-    episode: int = 1,
+    episode: int = 0,
     eval_episode: int = 1,
```
nick-harder, line 732: should eval_episode be 0 as well?

nick-harder (changes requested):

examples/examples.py
maurerle: > the other two commits are used to reproduce the mentioned issue

Yes, I will revert those, once we know whats up with #512

assume/scenario/loader_csv.py
maurerle: I didn't look into this change really, just applied it by recommendation in #512 but it did not help me reproduce an issue

nick-harder:
this also closes #517

nick-harder (approved):

PR #521, opened 2024-12-11
Fixes the issue that a warning was raised when there were no accepted bids

nick-harder:
Fixes the issue that a warning was raised when there were no accepted… bids. This is normal behavior and should not raise a warning, as this can naturally happen. The warning is raised only of both accepted and rejected are empty


maurerle (commented):
If this is what we want to warn about, we should do this at the beginning of the clear_market function and just look at  `self.all_orders`

maurerle (approved):
Looks fine to me

PR #524, opened 2025-01-02
Fix storage behavior

maurerle:
running: `assume -s example_01c -c dam_with_complex_opt_clearing -db "postgresql+psycopg2://assume:assume@localhost:5432/assume"`

previous behavior:

image

new behavior:

image

Description:

Storages were using the current SoC before, which led to always using the initial value to calculate discharge possibility.
This is fixed, by using the get_soc_before function as intended.

Also fixes debug printing of FastSeries (... ellipses was at the wrong place) and cuts of data if a timeseries index was given in FastIndex

### DIFF
--- a/assume/common/base.py
+++ b/assume/common/base.py
@@ -522,7 +522,7 @@ class SupportsMinMaxCharge(BaseUnit):
         Returns:
             float: The SoC before the given datetime.
         """
-        if dt - self.index.freq <= self.index[0]:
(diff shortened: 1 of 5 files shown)


nick-harder:
@maurerle thanks! is this ready for review?

nick-harder (approved):

nick-harder:
@maurerle please also add a release note so we can keep track of updates, thanks!

PR #531, opened 2025-01-14
dashboards: various fixes - merit order and technology plot

### LINKED ISSUE #322: Add merit order plot for a specific hour
Something like this:
image

but also show the demand like that:

https://www.epexspot.com/en/market-data?market_area=DE-LU&trading_date=2024-03-12&delivery_date=2024-03-13&underlying_year=&modality=Auction&sub_modality=DayAhead&technology=&product=60&data_mode=aggregated&period=&production_period=

image

This should also be available in the Grafana dashboard

maurerle:
* add merit order plot to overview
* add dispatch by technology to overview
* fix unmoveable panels from dashboards

closes #322 

## Screenshots

image

### DIFF
--- a/docker_configs/dashboard-definitions/ASSUME Comparison.json
+++ b/docker_configs/dashboard-definitions/ASSUME Comparison.json
@@ -348,8 +348,6 @@
           "sort": "none"
         }
       },
-      "repeat": "market",
-      "repeatDirection": "h",
       "targets": [
         {
           "datasource": {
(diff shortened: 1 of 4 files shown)


maurerle:
@nick-harder @kim-mskw I added correct colorization of the technology plot and fixed the title in the merit order plot.

I also adjusted the learning dashboard to not use transform actions but use SQL instead, so that the calculation happens on the server instead of the browser and hopefully does work a lot smoother.

nick-harder:
@maurerle thank you!

PR #536, opened 2025-01-21
Tensorboard integration

AndreasEppler:
# Pull Request

## Related Issue
No related issue

## Description
TensorBoard Implementation for visualizing the RL Training process without Grafana as Grafana is too slow for large simulations with many units.
Allows to observe daily average of `train` and `eval` params for each unit. 
For visuals, see screenshots below. 

### Changes to store `rl_critic_params` in db

- matd3.py: Returning `critic_losses` and `learning_rate` in `update_policy()` function
- learning_role.py: sending `learning_rate` and `critic_losses` to Output role through added `write_rl_critic_params_to_output()` function
- Outputs.py: handles new output_message from `learning_role`, merges new `rl_critic_params` (`learning_rate`and `critic_losses` plus `initial_exploration` parameter) to `rl_params` and stores them in db
- world.py: Initializes new `learning_role_agent`

### Changes to implement TensorBoard

- learning_utils.py: Added `TensorBoardLogger` class that writes tensorboard log files when called through `update_tensorboard()` function
- loader_csv.py: Handles TensorBoard logging Directory overwrite, calls `update_tensorboard` after each world.run()
- world.py: Initializes TensorBoard logger 
- ASSUME_Leaning.json: Removed old learning plots from Grafana Dashboard
- README.md: added Installation tutorial
- tensorboard_info.py: Docu file for displaying TensorBoard documentation in TensorBoard.

## Testing
- Extensive testing of small_learning_1, 2 & 3
- Used yappi to measure runtimes of the implementation: 
- For small_learning_3 (10 RL units), TB_writer consumes 0.8% of time, while write_dfs in Output role takes 0.1% longer. (see screenshots below)

## Checklist

## Additional Notes

Since the Meeting last Wednesday, the following changes were implemented as discussed:
- TensorBoard is now located in learning_utils instead of Output role
- Metrics are now aggregated on a daily Basis to Speed up TB writer
- Evaluation metrics are now also uploaded and displayed
- Different runs can now all be displayed simultaneously in the Dashboard (see screenshots) and can be filtered using regexes

### Visual example of TensorBoard:

TB visual example 2c
TB visual example 2a

### tuna runtime measurements:

Time spent in writer:

tuna TB 2c writer

Time spent in Output.store_ds:

TensorBoard integration:

tuna TB 2c output

Without TensorBoard integration

tuna main 2c output


assume/reinforcement_learning/learning_utils.py
```diff
@@ -153,3 +163,140 @@ def func(_):
         return val
 
     return func
+
```
maurerle, line 166: I would rather put this in `assume/reinforcement_learning/tensorboard_logger.py`

eventually next to a `tensorboard_info.py` or just integrate the text into the same file then..? :)

docs/tensorboard_info.py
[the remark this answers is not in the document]
AndreasEppler: Ah, that's good to know 👍🏻 I just implemented your suggestions.

assume/world.py
```diff
@@ -233,15 +243,16 @@ def setup(
             # self.clock_agent.stopped.add_done_callback(stop)
             self.container.register(self.clock_agent, suggested_aid="clock_agent")
         else:
-            self.setup_learning()
+            #TODO: which ID should be used here?
```
kim-mskw, line 246: Is that resolved @AndreasEppler?

kim-mskw:
Running example small_learnin_1 throws the following error: @AndreasEppler 

(assume-framework) PS C:\Users\tg3533\Documents\Code\assume> & C:/Users/tg3533/AppData/Local/miniconda3/envs/assume-framework/python.exe c:/Users/tg3533/Documents/Code/assume/examples/examples.py
Traceback (most recent call last):
  File "c:\Users\tg3533\Documents\Code\assume\examples\examples.py", line 9, in <module>
    from assume import World
  File "C:\Users\tg3533\Documents\Code\assume\assume\__init__.py", line 8, in <module>
    from assume.scenario.loader_csv import (
  File "C:\Users\tg3533\Documents\Code\assume\assume\scenario\loader_csv.py", line 29, in <module>
    from assume.world import World
  File "C:\Users\tg3533\Documents\Code\assume\assume\world.py", line 308
    tensorboard_path = f"{self.learning_config.get("trained_policies_save_path", "logs")}/TB",
                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: f-string: unmatched '('

kim-mskw:
> Running example small_learnin_1 throws the following error: @AndreasEppler
> 
> (assume-framework) PS C:\Users\tg3533\Documents\Code\assume> & C:/Users/tg3533/AppData/Local/miniconda3/envs/assume-framework/python.exe c:/Users/tg3533/Documents/Code/assume/examples/examples.py Traceback (most recent call last): File "c:\Users\tg3533\Documents\Code\assume\examples\examples.py", line 9, in from assume import World File "C:\Users\tg3533\Documents\Code\assume\assume__init__.py", line 8, in from assume.scenario.loader_csv import ( File "C:\Users\tg3533\Documents\Code\assume\assume\scenario\loader_csv.py", line 29, in from assume.world import World File "C:\Users\tg3533\Documents\Code\assume\assume\world.py", line 308 tensorboard_path = f"{self.learning_config.get("trained_policies_save_path", "logs")}/TB", ^^^^^^^^^^^^^^^^^^^^^^^^^^ SyntaxError: f-string: unmatched '('

Can you reproduce that?

AndreasEppler:
> > Running example small_learnin_1 throws the following error: @AndreasEppler
> > (assume-framework) PS C:\Users\tg3533\Documents\Code\assume> & C:/Users/tg3533/AppData/Local/miniconda3/envs/assume-framework/python.exe c:/Users/tg3533/Documents/Code/assume/examples/examples.py Traceback (most recent call last): File "c:\Users\tg3533\Documents\Code\assume\examples\examples.py", line 9, in from assume import World File "C:\Users\tg3533\Documents\Code\assume\assume__init__.py", line 8, in from assume.scenario.loader_csv import ( File "C:\Users\tg3533\Documents\Code\assume\assume\scenario\loader_csv.py", line 29, in from assume.world import World File "C:\Users\tg3533\Documents\Code\assume\assume\world.py", line 308 tensorboard_path = f"{self.learning_config.get("trained_policies_save_path", "logs")}/TB", ^^^^^^^^^^^^^^^^^^^^^^^^^^ SyntaxError: f-string: unmatched '('
> 
> Can you reproduce that?

Huh no, I can't. When is the error being thrown? At the beginning or the end of the training? And is your config file unchanged, especially `trained_policies_save_path`?

assume/world.py
AndreasEppler: Yes, resolved it  commit 8402de82469632cac8c806ed6151789f3a9bfa20.

nick-harder:
@AndreasEppler @kim-mskw should i resolve the merge conflicts or you will take care of it?

AndreasEppler:
> @AndreasEppler @kim-mskw should i resolve the merge conflicts or you will take care of it?

I can resolve them but I'm currently on a train so I have to wait until I've got stable internet connection, which is approximately in about an hour. If you don't want to wait, you can also do it as it's only small conflicts.

assume/world.py
```diff
@@ -267,6 +283,37 @@ def setup_learning(self) -> None:
...
+                    "output_agent_addr": self.output_agent_addr,
+                    "train_start": self.start,
+                    "train_end": self.end,
+                    "freq": self.forecaster.index.freq,
+                }
+            )
+
+            self.learning_operators[id] = learning_role_agent
+
+            self.tensor_board_logger = TensorBoardLogger(
```
nick-harder, line 308: @AndreasEppler why make the tensor board a seperate part of world if it could simply be an object inside the learning role and everything could be handled within it?

assume/world.py
```diff
@@ -267,6 +283,37 @@ def setup_learning(self) -> None:
             )
             rl_agent.suspendable_tasks = False
 
+            id = "Learning_1"
+            # add Learning_1 to the list of learning operators if not already existing
+            if self.learning_operators.get(id):
+                raise ValueError(f"LearningOperator {id} already exists")
+
+            learning_role_agent = RoleAgent()
```
nick-harder, line 291: I don't understand the reason of all these extra lines of code. LEarning Role was already a Role, it could receive and send messages, Why we need this extra?

assume/world.py
```diff
@@ -66,18 +66,21 @@ class World:
         market_operators (dict[str, mango.RoleAgent]): The market operators for the world instance.
         markets (dict[str, MarketConfig]): The markets for the world instance.
         unit_operators (dict[str, UnitsOperator]): The unit operators for the world instance.
+        learning_operators (dict[str, RoleAgent]): The learning operators for the world instance.
```
nick-harder, line 69: for now we don't support multiple learning operators, and we don't have a concrete idea on how we would implement this in ASSUME. So I believe this should not be here for now. maybe in future developments

assume/world.py
```diff
@@ -66,18 +66,21 @@ class World:
...
         unit_types (dict[str, BaseUnit]): The unit types for the world instance.
         bidding_strategies (dict[str, type[BaseStrategy]]): The bidding strategies for the world instance.
         clearing_mechanisms (dict[str, MarketRole]): The clearing mechanisms for the world instance.
         addresses (list[str]): The addresses for the world instance.
         loop (asyncio.AbstractEventLoop): The event loop for the world instance.
+        tensor_board_logger (TensorBoardLogger): The tensor board logger for the world instance.
```
nick-harder, line 75: as said below, I believe a better place for the tensorboard is in the learning role itself not to confuse regular users. Also, I don't see any benefits of it being in the world? Learning role has all access we need

assume/common/outputs.py
```diff
@@ -61,6 +62,7 @@ def __init__(
         export_csv_path: str = "",
         save_frequency_hours: int = None,
         learning_mode: bool = False,
+        episodes_collecting_initial_experience: int = 0,
```
nick-harder, line 65: there is no need to drag the episodes_collecting_initial_experience variable through the whole simulation, the learning role can simply include this into the message when sending the rl_params, since this information is included in the learning strategy itself. This info is also included in the learning role and can be also sent along other parameters

assume/reinforcement_learning/learning_role.py
```diff
@@ -34,6 +35,7 @@ class Learning(Role):
     def __init__(
         self,
         learning_config: LearningConfig,
+        episodes_collecting_initial_experience: int,
```
nick-harder, line 38: why is this added here? this is defined below in init from the config

assume/reinforcement_learning/learning_role.py
```diff
@@ -387,3 +392,60 @@ def compare_and_save_policies(self, metrics: dict) -> bool:
...
+        )
+
+        output_list = [
+            {
+                "unit": u_id,
+                "critic_loss": params["loss"],
+                "total_grad_norm": params["total_grad_norm"],
+                "max_grad_norm": params["max_grad_norm"],
+                "learning_rate": learning_rate,
+                "datetime": self.datetime
```
nick-harder, line 433: can you please explain why we need the datetime for non time series data here?

assume/reinforcement_learning/learning_role.py
```diff
@@ -307,7 +311,8 @@ def update_policy(self) -> None:
             This method is typically scheduled to run periodically during training to continuously improve the agent's policy.
         """
         if self.episodes_done >= self.episodes_collecting_initial_experience:
-            self.rl_algorithm.update_policy()
+            learning_rate, unit_params = self.rl_algorithm.update_policy()
+            self.write_rl_critic_params_to_output(learning_rate, unit_params)
```
nick-harder, line 315: there is no need to extract this values here and forward them into the function again. This can move directly to the end of policy update, so that the values are sent to the function before the update_policy function ends

assume/reinforcement_learning/algorithms/matd3.py
```diff
@@ -596,3 +621,5 @@ def update_policy(self):
                 # Perform batch-wise Polyak update (NO LOOPS)
                 polyak_update(all_critic_params, all_target_critic_params, self.tau)
                 polyak_update(all_actor_params, all_target_actor_params, self.tau)
+
+        return learning_rate, unit_params
```
nick-harder, line 625: here, instead of returning the values, just call the learning_role function directly and send the data

assume/common/outputs.py
```diff
@@ -440,6 +460,22 @@ async def store_dfs(self):
...
+            "rl_params" in self.write_buffers
+            and "rl_critic_params" in self.write_buffers
+        ):
+            df1 = pd.DataFrame(self.write_buffers["rl_params"])
+            df2 = pd.DataFrame(self.write_buffers["rl_critic_params"])
+            merged_df = pd.merge(df1, df2, how="outer")
+            merged_list = merged_df.to_dict("records")
+            self.write_buffers["rl_params"] = merged_list
+            del self.write_buffers["rl_critic_params"]
+        # if only rl_critic_params are present, rename them to rl_params
```
nick-harder, line 474: this comment doesn't make sense, the following statements are outside of the if else condition

assume/reinforcement_learning/learning_role.py
```diff
@@ -387,3 +392,60 @@ def compare_and_save_policies(self, metrics: dict) -> bool:
...
+        """
+        db_addr = self.context.data.get("output_agent_addr")
+
+        # Initialize datetime the first time the function is called
+        if self.datetime is None:
+            self.datetime = pd.to_datetime(self.context.data.get("train_start"))
+
+        freq_timedelta = pd.Timedelta(self.context.data.get("freq"))
+        # This is for padding the output list when the gradient steps are not identical to the train frequency
+        time_steps = int(
```
nick-harder, line 422: this can be calculated once in the init function since this doesn't change during the simulation

nick-harder (changes requested):
@AndreasEppler thanks for the nice work! I didn't check the tensor logger itself yet, for now I only looked into the integration. And I have a couple of comments. One main question is why keep the tensorboard as part of world and not simply inside the learning role? I don't see any benefits of it being a part of the world, only more confusion later on. The learning role has all the info we need anyway

assume/common/outputs.py
AndreasEppler: I do not understand your remark, to be honest. The following statements are:
# if only rl_critic_params are present, rename them to rl_params
 elif "rl_critic_params" in self.write_buffers:
            self.write_buffers["rl_params"] = self.write_buffers["rl_critic_params"]
            del self.write_buffers["rl_critic_params"]
which is part of the `if`-`elif` structure. Can you please elaborate on what you mean? 
I'll make the comments more precise though.

assume/reinforcement_learning/learning_role.py
AndreasEppler: As far as I understand mango, this cannot be calculated in the init as it uses context data. As the context data is updated after initializing the learning_role, the context data is not available in the init yet. I now created a new init_logging function where all of this is calculated only once at the beginning instead.

assume/reinforcement_learning/learning_role.py
AndreasEppler: Yes, the datetime is necessary for the later merge of the `rl_params` and `rl_critic_params` in the output_role. We need the `rl_critic_params` to contain datetime information in order to store all of the learning_data combined in the database.

assume/world.py
AndreasEppler: Makes sense. I just moved the tensorboard logger to the learning_role.

assume/world.py
AndreasEppler: You're right, they are indeed unnecessary. I removed them and am using the learning_role itself now.

assume/world.py
AndreasEppler: Good question, moved it to learning_role. 👍🏻

assume/reinforcement_learning/learning_role.py
AndreasEppler: I added this as I initialized `episodes_collecting_inital_experience` in world in order to hand it to the output_role. I did not want this code:
```python
self.episodes_collecting_initial_experience = max(
            learning_config.get("episodes_collecting_initial_experience", 5), 1
        )
```
to exist twice in the codebase (both in world and learning_role) as it could lead to different fallback values existing if only one of the fallback values was changed in the future. Therefore I decided to hand the parameter over to learning_role as an argument instead and remove the above lines from learning_role.

assume/reinforcement_learning/learning_role.py
kim-mskw: If we wanted to change that long-term, we would need different learning parameters, which we potentially wanted to tackle in another PR.

assume/common/outputs.py
nick-harder: @AndreasEppler so this means during the initial exploration we are collecting info only from the learning strategies themselves, right? And this info is collected by the rl_unit_operator in the write_to_learning_role() method. And since the bidding strategies also contain the information about the initial experiences, we can pack that info into the message as well. So we will have all the things coming from a single place, what do you think?

nick-harder: @AndreasEppler solved

nick-harder:
@AndreasEppler @kim-mskw I have tried the tensorboard, everything works fine, but I believe we still need to work on how the things are logged into the tensorboard, since in it's current form it is fine for a single agent of very small number of agents. But using it for a simulation with like 20 30 agents already makes the tensorboard impossible to use and to find anything in it. Maybe we need to drop the per agent metrics from the tensorboard to make it usable, and refer to the grafana for detailed per agent view. If we remove the average plots from grafana, it works perfectly. Also, I find the idea of each metric being logged as a seperate run a bit unconventional, in my opinion everything under one simulation_id should be treated as one run. So we can also compare different runs to each other. Right now it is hard to see anything. Let's discuss this in detail tomorrow during the call

nick-harder:
I have created a branch with my suggested changes here https://github.com/assume-framework/assume/tree/only-average-in-tensorboard

AndreasEppler:
As Kim is still sick, we won't be in tomorrow's dev call @nick-harder. But we can talk about it at the next opportunity.

assume/reinforcement_learning/learning_role.py
```diff
@@ -137,6 +141,17 @@ def __init__(
         # list of avg_changes
         self.avg_rewards = []
 
+        # tensor board logging
+        self.tensor_board_path = (
```
kim-mskw, line 145: Was this not supposed to be in a different folder now?

kim-mskw: I think this is redundant and can be deleted? Please check @AndreasEppler

kim-mskw: Additionally the tutorials still have an unnecessary remove of ensorboard directory as save_path still?

kim-mskw (approved):
See last comment, otherwise approved

assume/reinforcement_learning/learning_role.py
AndreasEppler: Good catch, removed it 👍🏻

nick-harder (approved):

PR #544, opened 2025-02-10
Improve competetive behavior

nick-harder:
# Pull Request

## Description
Until now large scale MADRL simulations didn't produce desired results. I have reevaluated my approach and introduced some changes to the learning process to improve.

## Changes Proposed
-  Introduce randomized policy update instead of updating all policies at once to prevent cooperative behavior and training.
- Refactor reward function to obstruct overprofits to bring the players closer to reality where obvious market manipulation would be punished.
- added smaller fixes to how continue learning is treated to enable learning more sophisticated policies from initial fast ones

## Testing
Performed the simulation for a month of 2019 to get working policies faster. Reused those policies as starting policies for the full 2019 simulation with quick update.

Some metrics for 2019 for the simulation in hourly resolution:
RMSE: 10.22 EUR/MWh
MAE: 6.52 EUR/MWh
SMAPE: 21%
MBD: 0.8 EUR/MWh

## Checklist

Some plots for the 2019 simulation
daily_average_2019
hourly_07_2019
hourly_12_2019


assume/reinforcement_learning/algorithms/matd3.py
```diff
@@ -386,6 +385,9 @@ def update_policy(self):
         strategies = list(self.learning_role.rl_strats.values())
         n_rl_agents = len(strategies)
 
+        # Randomly select a subset of strategies to update
+        selected_strategies = random.sample(strategies, k=int(n_rl_agents / 3))
```
maurerle, line 389: is this fixed to use a third of the agents?
Why?
should this be configurable?

examples/inputs/example_03a/config.yaml
```diff
@@ -7,29 +7,29 @@ base_case_2019:
...
+    device: cuda:0
+    # action_noise_schedule: linear
     noise_sigma: 0.1
     noise_scale: 1
     noise_dt: 1
     validation_episodes_interval: 5
-    early_stopping_steps: 10
-    early_stopping_threshold: 0.05
+    # early_stopping_steps: 10
+    # early_stopping_threshold: 0.05
```
maurerle, line 32: i would either remove this or keep it set with a default value?

Should we keep the cuda:0?

maurerle (approved):
looks fine for me in total.
I left some comments you can address

assume/reinforcement_learning/algorithms/matd3.py
nick-harder: @maurerle this is something we will discuss during our dev meeting this week

examples/inputs/example_03a/config.yaml
nick-harder: @maurerle yeah it kind of doesn't make sense to run such large MADRL runs on a cpu, therefore the cuda device. if none is available, it will use cpu anyway. this is just to show people that cuda is prefered here. Idk if this makes sense

nick-harder:
@maurerle I saw some issues with how we were loading everything twice when in learning mode, especially with the learned strategies being loaded twice, as it caused sometimes weird behavior when I wanted to continue learning. Could you please take another look at the PR if you are fine with it?

assume/world.py
```diff
@@ -116,18 +116,33 @@ def __init__(
...
+                    logger.error(
+                        "Could not connect to %s, trying again (%d/%d)",
+                        database_uri,
+                        attempts,
+                        max_attempts,
+                    )
                     # log error if not connection refused
                     if not e.code == "e3q8":
                         logger.error("%s", e)
                     time.sleep(2)
```
maurerle, line 138:
```suggestion
                    time.sleep(2**attempts)
```
This increases total wait time to 1 minute in total which should be more suitable in some scenarios
(2 + 4 + 8 + 16 + 32)

.gitignore
```diff
@@ -142,6 +142,7 @@ assume-db
 learned_strategies
 examples/outputs
 assume-repo
+tensorboard
```
maurerle, line 145: wouldn't it be better to leave this change in the tensorboard PR?

nick-harder:
@maurerle thanks! So I assume you are also fine with the way how the scenario data is now loaded into the world, and how learning is executed now?

assume/scenario/loader_csv.py
```diff
@@ -1025,11 +1015,16 @@ def run_learning(
 
     world.reset()
 
+    # Set 'trained_policies_load_path' to None in order to load the most recent policies,
+    # especially if previous strategies were loaded from an external source.
+    # This is useful when continuing from a previous learning session.
+    world.scenario_data["config"]["learning_config"]["trained_policies_load_path"] = (
```
kim-mskw, line 1021: I wonder here, why do we not use the best policies instead of the most recent ones?

assume/strategies/learning_strategies.py
```diff
@@ -539,77 +539,87 @@ def calculate_reward(
...
+
+        # account for only 10% of the profit if it is positive to not overstimulate the egant
+        # to manipulate the market
+        profit = min(profit, 0.1 * abs(profit))
+
+        # calculate opportunity cost
+        # as the loss of income we have because we are not running at full power
+        opportunity_cost = (
+            (market_clearing_price - marginal_cost)
+            * (offered_volume_total - accepted_volume_total)
```
kim-mskw, line 599: Why not stick here to the actual dispatch and the max_power? Seems more robust in case we have some differences between the offered volume and the actual dispatch? Also the Opportunity costs are not bound by the offered volume, but the max_power that could have been offered.

assume/strategies/learning_strategies.py
```diff
@@ -539,77 +539,87 @@ def calculate_reward(
...
+            * duration
+        )
+
+        # if our opportunity costs are negative, we did not miss an opportunity to earn money and we set them to 0
+        opportunity_cost = max(opportunity_cost, 0)
 
-        profit = profit - costs
+        # set regret_scale as 0.1 if dispatched above min. power, 0.5 if not dispatched at all
```
kim-mskw, line 606: Could you give an explanation here why you make the difference?

assume/strategies/learning_strategies.py
```diff
@@ -539,77 +539,87 @@ def calculate_reward(
...
+            and unit.outputs[product_type].at[start - unit.index.freq] != 0
         ):
-            costs += unit.hot_start_cost / 2
+            operational_cost += unit.hot_start_cost / 2
+
+        profit = income - operational_cost
+
+        # account for only 10% of the profit if it is positive to not overstimulate the egant
+        # to manipulate the market
+        profit = min(profit, 0.1 * abs(profit))
```
kim-mskw, line 593: This distinction between the negative and positive profit is worrying me a bit. It seems like you induce risk aversity durch die Hintertür (cannot think of a nice english translation). My first thought was: Your pushing the agent to just cover its marginal costs, by weighing the negative profits that much higher. It seem natrual to me that they rather converge to their mc bidding. Whcih of course makes the learning more stable but also primes the results beforehand. I think this might be a case of too much reward tuning, depending on what we want.

examples/inputs/example_02c/config.yaml
```diff
@@ -14,22 +14,22 @@ base:
...
     gamma: 0.99
     device: cpu
     action_noise_schedule: linear
     noise_sigma: 0.1
     noise_scale: 1
     noise_dt: 1
-    validation_episodes_interval: 5
-    early_stopping_steps: 10
-    early_stopping_threshold: 0.05
+    # validation_episodes_interval: 5
```
kim-mskw, line 30: Out of curiosity, why did you comment out the validation interval?

assume/scenario/loader_csv.py
nick-harder: yeah true, this was the default behavior before, but it would be now a good point to change that

assume/strategies/learning_strategies.py
nick-harder: this makes them more sensitive to losses, but doesn't prevent from getting the increased profit and using it to the full potential, if present.

examples/inputs/example_02c/config.yaml
nick-harder: oh, the configs are a bit of a mess, I need to check them out again before merge, as this was me just playing around.

assume/strategies/learning_strategies.py
nick-harder: I believe we should keep it to the offered and accepted volumes in this reward function, as if we were to expand this to multiple markets, and there would be a seperate reward calculation for each market, this logic would be correct. With only a single market available, this formulation equals the max power and current dispatch logic.

assume/strategies/learning_strategies.py
nick-harder: I believe punishing power plants, which just happened to be on the edge of the merit order the same way as those, who didn't make the cut at all, is not fair and therefore we should distinguish between them

assume/strategies/learning_strategies.py
kim-mskw: True, but weighing every positive profit with just 10% of the actual profit is quite a strong assumption, which puts very much weight on avoiding losses in comparison to making a profit. Consequently, of course every unit rather bids their mc instead of anything else. 

For trying to find emergent bidding startegies or novel equilibria in zero mc systems this seem to be a very strong nudging. It makes sense if you want to verify the historic results of course.

assume/strategies/learning_strategies.py
kim-mskw: In a multi-market setting the opportunity costs should still be calculated based on the question what if I would offer my entire capacity in this market and not the other. Hence, the upper bound should still be max_power in my opinion.

assume/strategies/learning_strategies.py
nick-harder: for zero mc systems this has absolutely no effect since their profits cannot be negative as they don't have any marginal costs

assume/strategies/learning_strategies.py
nick-harder: yes, agree, changed

PR #545, opened 2025-02-13
Improves built-in forecast calculations

nick-harder:
Currently used methods for calculating built-in forecasts are limited and sometimes even wrong
- instead of using only the marginal cost for the first time step, now it uses time varying marginal costs to calculate simple merit order based market clearing price, so the prices change if the fuel price is also a time series
- extracting VRE technologies for residual load calculation is now more robust taking all power plants with wind and solar in the technology, so any combinations like wind_onshore or Wind Offshore will work fine. This has already caused some issues, as this is not documented anywhere.

### DIFF
--- a/assume/common/forecasts.py
+++ b/assume/common/forecasts.py
@@ -96,6 +96,12 @@ class CsvForecaster(Forecaster):
     Methods are included to retrieve forecasts for specific columns, availability of units,
     and prices of fuel types, returning the corresponding timeseries as pandas Series.
 
+    Notes:
+    - Some built-in forecasts are calculated at the beginning of the simulation, such as price forecast and residual load forecast.
+    - Price forecast is calculated for energy-only markets using a merit order approach.
+    - Residual load forecast is calculated by subtracting the total available power from variable renewable energy power
+      plants from the overall demand forecast. Only power plants containing 'wind' or 'solar' in their technology column are considered VRE power plants.
+
     Args:
         index (pd.Series): The index of the forecasts.
         powerplants_units (pd.DataFrame): A DataFrame containing information about power plants.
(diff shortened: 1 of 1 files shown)


assume/common/forecasts.py
```diff
@@ -359,39 +365,61 @@ def calculate_market_price_forecast(self, market_id):
...
+            sorted_units = mc_t.sort_values().index
+            sorted_mc = mc_t.loc[sorted_units]
+            sorted_power = power_t.loc[sorted_units]
+
+            # Compute the cumulative sum of available power in the sorted order.
+            cumsum_power = sorted_power.cumsum()
+
+            # Find the first unit where the cumulative available power meets or exceeds demand.
+            matching_units = cumsum_power[cumsum_power >= demand_t]
+            if not matching_units.empty:
```
maurerle, line 415:
```suggestion
            if matching_units.empty:
```

I would turn this around to avoid the double negation in the else case.
And but the price=1000 case at first

assume/common/forecasts.py
```diff
@@ -359,39 +365,61 @@ def calculate_market_price_forecast(self, market_id):
...
-        for col in sorted_columns[::-1]:
-            # find times which can still be provided with this technology
-            # and cheaper once
-            cheaper = cumsum_power[col] > sum_demand
-            # set the price of this technology as the forecast price
-            # for these times
-            price_forecast.loc[cheaper] = marginal_costs[col].loc[cheaper]
-            # repeat with the next cheaper technology
+        # 6. Initialize the price forecast series.
+        price_forecast = pd.Series(index=self.index, data=np.nan)
```
maurerle, line 396: Why should this be nan?

maurerle (approved):
Makes sense to calculate the merit order for every hour with the correct prices.
Thanks!

assume/common/forecasts.py
nick-harder: true, fixed

nick-harder:
@maurerle this actually also had some pretty significant effect on the results, in big part because of how the storages reacted
image

PR #546, opened 2025-02-17
Adjust tensorboard display

nick-harder:
Restructure show tensorboard is displayed and what is logged


AndreasEppler:
tests/test_tensorboard_logger.py no longer works so I'm currently fixing that before we can merge the branches.

PR #551, opened 2025-02-21
Improve market participation

maurerle:
This makes sure, that at least one unit has a bidding strategy for a market, before deciding to participate on that market.

Respect the `eligible_obligations_lambda` set in the `MarketConfig` to only bid on markets where the UnitsOperator fulfills the requirements.

It is possible to set this as string.

User documentation is currently missing.


assume/common/market_objects.py
```diff
@@ -279,3 +279,39 @@ class MetaDict(TypedDict):
 
 contract_type = Callable[[Agent, Agent], None]
 market_contract_type = Callable[[Agent, Agent, list], None]
+
+
+def only_renewables(unit):
+    """check that technology is renewable"""
+    return unit["technology"] in [
```
nick-harder, line 286: it would be better if this technologies were documented somewhere, or at least the check would not check exact matching, but at least allow some variation like Solar or wind onshore, and so on. You can check how this is done in the residual load calculation function in the forecaster. Otherwise this functionality is very limited and works only for those, who wrote it and knows in detail.

nick-harder (changes requested):

nick-harder (approved):

Issue #556, opened 2025-02-26
Returning wrong date from FastIndex for start_idx


mthede:
When running a simulation with (naive) storage units after pulling the latest commits from main, I wondered why the storage units only charge but no longer discharge. Additionally, the soc didn't increase, although the market dispatch and actual dispatch was negative.

I pinpointed the issue to the [storage.execute_dispatch_function() ](https://github.com/assume-framework/assume/blob/b53a650aff757a7dec4010bdfdddf3001b5f9d6a/assume/units/storage.py#L179). It was returning a DateTimeIndex for t+1 instead of t as would be expected.

Passing the "round_up=False" argument also to the start_idx like for the stop_idx fixing #550 in the FastIndex fast_pandas.py solved the issue and storage units were successfully discharging. However, I'm not 100% if it is correct in terms of the specific market clearning timestep (t) and the dispatch timestep (t+1).
![Image](https://github.com/user-attachments/assets/e3b7aff4-f35e-4ca2-8942-218d03de3a82)

@maurerle Would this be the intended behavior?

nick-harder:
@maurerle I believe this is something you have introduced recently, could you please take a look into this behavior?

Issue #549, opened 2025-02-20
Documentation of bidding strategies is missing


nick-harder:
We have identified, that the bidding strategies are not described in the documentation. We also don't have a list of available bidding strategies and how are they named to use them in the input files. We need to improve this ASAP @kim-mskw @maurerle @gugrimm @Manish-Khanra. We should also change the names of the bidding strategies since all these flexable, dmas and so on are apparently very confusing to people. We need to identify a common structure for naming the bidding strategies, maybe something like "unit_market_specific", for example powerplant_eom_naive or storage_eom if no specific info is provided.

tiernan-buckley-ufr:
@nick-harder I'll look into it

maurerle:
We should also add information about the new UnitsOperatorStrategies from #641

The UnitsOperator now can have a UnitsOperatorStrategy in `portfolio_strategies` per market.
If a market has an operator strategy, it is used for all biddings on this market.
Otherwise, all units bid independtly from each other using the `IndependentBiddingStrategy`.

PR #557, opened 2025-02-26
Fix storage strategies

nick-harder:
FlexABLE storage strategies were acting weird when used with a day ahead bidding. The problem was that the SOC was updated only later during execute_current_dispatch and was not reflected for the next bidding period, where the storage units assumed again an initial soc of 0. This PR adds an soc update part to the set_dispatch function when used by storage units. This also refactors the flexable storage strategy to make viable bids on the day ahead market compared to before.

### DIFF
--- a/assume/common/base.py
+++ b/assume/common/base.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
 from collections import defaultdict
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import TypedDict
 
 import numpy as np
@@ -171,6 +171,44 @@ class BaseUnit:
             self.outputs[f"{product_type}_accepted_price"].loc[start:end_excl] = (
                 accepted_price
             )
+
+        # also update the SOC if the unit is a storage unit
+        if isinstance(self, SupportsMinMaxCharge):
+            start = orderbook[0]["start_time"]
+            end = orderbook[-1]["end_time"]
+            time_delta = self.index.freq / timedelta(hours=1)
+
+            for t in self.index[start:end]:
+                next_t = t + self.index.freq
+                # continue if it is the last time step
+                if next_t not in self.index:
+                    continue
+                current_power = self.outputs["energy"].at[t]
+
+                # calculate the change in state of charge
+                delta_soc = 0
+                soc = self.outputs["soc"].at[t]
+
+                # discharging
+                if current_power > 0:
+                    max_soc_discharge = self.calculate_soc_max_discharge(soc)
+
+                    if current_power > max_soc_discharge:
+                        current_power = max_soc_discharge
+
+                    delta_soc = -current_power * time_delta / self.efficiency_discharge
+
+                # charging
+                elif current_power < 0:
+                    max_soc_charge = self.calculate_soc_max_charge(soc)
+
+                    if current_power < max_soc_charge:
+                        current_power = max_soc_charge
+
+                    delta_soc = -current_power * time_delta * self.efficiency_charge
+
+                # update the values of the state of charge and the energy
+                self.outputs["soc"].at[next_t] = soc + delta_soc
 
     def calculate_cashflow_and_reward(
         self,
@@ -538,23 +576,6 @@ class SupportsMinMaxCharge(BaseUnit):
         Returns:
             tuple[np.array, np.array]: The min and max discharging power for the given time period.
         """
-
-    def get_soc_before(self, dt: datetime) -> float:
-        """
-        Returns the State of Charge (SoC) before the given datetime.
-        If datetime is before the start of the index, the initial SoC is returned.
-        The SoC is a float between 0 and 1.
-
-        Args:
-            dt (datetime.datetime): The current datetime.
-
-        Returns:
-            float: The SoC before the given datetime.
-        """
-        if dt - self.index.freq < self.index[0]:
-            return self.initial_soc
-        else:
-            return self.outputs["soc"].at[dt - self.index.freq]
 
     def calculate_ramp_discharge(
         self,
--- a/assume/strategies/dmas_storage.py
+++ b/assume/strategies/dmas_storage.py
@@ -144,7 +144,7 @@ class DmasStorageStrategy(BaseStrategy):
         ]
 
         self.model.vol_con = pyo.ConstraintList()
-        v0 = unit.get_soc_before(start)
+        v0 = unit.outputs["soc"].at[start - unit.index.freq]
 
         for t in time_range:
             if t == 0:
--- a/assume/strategies/flexable_storage.py
+++ b/assume/strategies/flexable_storage.py
@@ -61,17 +61,17 @@ class flexableEOMStorage(BaseStrategy):
         # Calculate bids
         # =============================================================================
         # save a theoretic SOC to calculate the ramping
-        theoretic_SOC = unit.outputs["soc"].at[product_tuples[0][0]]
+        start = product_tuples[0][0]
+        theoretic_SOC = unit.outputs["soc"].at[start]
+        previous_power = unit.get_output_before(start)
+
+        current_power = unit.outputs["energy"].at[start]
+        current_power_discharge = max(current_power, 0)
+        current_power_charge = min(current_power, 0)
 
         bids = []
         for product in product_tuples:
             start, end = product[0], product[1]
-
-            previous_power = unit.get_output_before(start)
-
-            current_power = unit.outputs["energy"].at[start]
-            current_power_discharge = max(current_power, 0)
-            current_power_charge = min(current_power, 0)
 
             # calculate min and max power for charging and discharging
             min_power_charge, max_power_charge = unit.calculate_min_max_charge(
--- a/assume/strategies/flexable_storage.py
+++ b/assume/strategies/flexable_storage.py
@@ -58,47 +58,28 @@ class flexableEOMStorage(BaseStrategy):
         """
 
         # =============================================================================
-        # Storage Unit is either charging, discharging, or off
-        # =============================================================================
-        start = product_tuples[0][0]
-        end_all = product_tuples[-1][1]
-
-        previous_power = unit.get_output_before(start)
-
-        # save a theoretic SOC to calculate the ramping
-        theoretic_SOC = unit.get_soc_before(start)
-
-        # calculate min and max power for charging and discharging
-        min_power_charge_values, max_power_charge_values = (
-            unit.calculate_min_max_charge(start, end_all)
-        )
-        min_power_discharge_values, max_power_discharge_values = (
-            unit.calculate_min_max_discharge(start, end_all, soc=theoretic_SOC)
-        )
-
-        # =============================================================================
         # Calculate bids
         # =============================================================================
-        bids = []
+        # save a theoretic SOC to calculate the ramping
+        theoretic_SOC = unit.outputs["soc"].at[product_tuples[0][0]]
 
-        for (
-            product,
-            max_power_discharge,
-            min_power_discharge,
-            max_power_charge,
-            min_power_charge,
-        ) in zip(
-            product_tuples,
-            max_power_discharge_values,
-            min_power_discharge_values,
-            max_power_charge_values,
-            min_power_charge_values,
-        ):
+        bids = []
+        for product in product_tuples:
             start, end = product[0], product[1]
+
+            previous_power = unit.get_output_before(start)
 
             current_power = unit.outputs["energy"].at[start]
             current_power_discharge = max(current_power, 0)
             current_power_charge = min(current_power, 0)
+
+            # calculate min and max power for charging and discharging
+            min_power_charge, max_power_charge = unit.calculate_min_max_charge(
+                start, end, soc=theoretic_SOC
+            )
+            min_power_discharge, max_power_discharge = unit.calculate_min_max_discharge(
+                start, end, soc=theoretic_SOC
+            )
 
             # Calculate ramping constraints using helper function
             max_power_discharge = unit.calculate_ramp_discharge(
@@ -176,8 +157,6 @@ class flexableEOMStorage(BaseStrategy):
 
             theoretic_SOC += delta_soc
             previous_power = bid_quantity + current_power
-
-        bids = self.remove_empty_bids(bids)
 
         return bids
 
@@ -267,7 +246,7 @@ class flexablePosCRMStorage(BaseStrategy):
         end = product_tuples[-1][1]
 
         previous_power = unit.get_output_before(start)
-        theoretic_SOC = unit.get_soc_before(start)
+        theoretic_SOC = unit.outputs["soc"].at[start]
 
         _, max_power_discharge_values = unit.calculate_min_max_discharge(
             start, end, soc=theoretic_SOC
@@ -400,7 +379,7 @@ class flexableNegCRMStorage(BaseStrategy):
 
         previous_power = unit.get_output_before(start)
 
-        theoretic_SOC = unit.get_soc_before(start)
+        theoretic_SOC = unit.outputs["soc"].at[start]
 
         _, max_power_charge_values = unit.calculate_min_max_charge(start, end)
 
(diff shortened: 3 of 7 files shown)


assume/units/storage.py
```diff
@@ -368,11 +372,17 @@ def calculate_min_max_discharge(
...
+            soc = self.outputs["soc"].at[start]
+        max_soc_discharge = self.calculate_soc_max_discharge(soc)
         max_power_discharge = max_power_discharge.clip(max=max_soc_discharge)
 
+        # convert to single values if only one value is left
+        if len(min_power_discharge) == 1:
+            min_power_discharge = min_power_discharge[0]
+        if len(max_power_discharge) == 1:
+            max_power_discharge = max_power_discharge[0]
+
```
maurerle, line 385: this does not give a clear return type - requiring to test this if the return is float or array..?

Even though I understand it is tempting, I would rather keep this to return a list with one element in these cases..?

assume/common/base.py
```diff
@@ -172,6 +172,44 @@ def set_dispatch_plan(
                 accepted_price
             )
 
+        # also update the SOC if the unit is a storage unit
+        if isinstance(self, SupportsMinMaxCharge):
+            start = orderbook[0]["start_time"]
+            end = orderbook[-1]["end_time"]
+            time_delta = self.index.freq / timedelta(hours=1)
+
+            for t in self.index[start:end]:
```
mthede, line 181: I'm again confused by the behavior of the FastIndex. For the first set_dispatch with DA-bidding, start="2019-01-02 00:00:00" and end="2019-01-03 00:00:00": Iterating through the index gives 25 t's. I'd expect 24, because the last dispatch should set the soc for "2019-01-03 00:00:00" instead of "2019-01-03 01:00:00", right? alt text

mthede: grafik

nick-harder: @mthede yes you are right good catch, the end should be set as start_time of the last bid. I have fixed it now, please check it again

mthede: I'm a bit biased because the index caused unexpected behavior in my own scenarios #556

tests/test_storage.py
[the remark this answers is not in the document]
nick-harder: what clutter do you mean? 

It would be really great if you could add such a test

assume/common/base.py
```diff
@@ -518,44 +559,27 @@ def calculate_min_max_charge(
         Args:
             start (datetime.datetime): The start time of the dispatch.
             end (datetime.datetime): The end time of the dispatch.
-            product_type (str, optional): The product type of the unit. Defaults to "energy".
+            soc (float, optional): The current state-of-charge. Defaults to None.
```
maurerle, line 562: should be changed in the method signature as well..?

assume/strategies/flexable_storage.py
```diff
@@ -58,47 +58,37 @@ def calculate_bids(
...
 
         # =============================================================================
-        # Storage Unit is either charging, discharging, or off
+        # Calculate bids
         # =============================================================================
+        # save a theoretic SOC to calculate the ramping
         start = product_tuples[0][0]
-        end_all = product_tuples[-1][1]
-
+        theoretic_SOC = unit.outputs["soc"].at[start]
```
maurerle, line 65: is this correct to be the soc at start and not the one before?

assume/strategies/flexable_storage.py
```diff
@@ -267,7 +255,7 @@ def calculate_bids(
         end = product_tuples[-1][1]
 
         previous_power = unit.get_output_before(start)
-        theoretic_SOC = unit.get_soc_before(start)
+        theoretic_SOC = unit.outputs["soc"].at[start]
```
maurerle, line 258: same here? We want to set a bid for `start`, therefore taking the previous soc right?

But the SoC for `start` might already be adjusted due to a different market bidding here.

To calculate the min_max_discharge, we therefore need to include the soc_before (what we came from), the soc at `start` (which might already have changed), and check if our requested delta is in the allowed change of this.

Prior, we only used the soc before, now we only use the one at `start` and check if the individual changes of one market do not exceed ramping constraints..

If this does work both ways, without a test failing, we need to add tests first

assume/strategies/learning_strategies.py
```diff
@@ -816,11 +816,8 @@ def calculate_bids(
         else:
             bid_direction = "ignore"
 
-        _, max_discharge = unit.calculate_min_max_discharge(start, end_all)
-        _, max_charge = unit.calculate_min_max_charge(start, end_all)
-
-        bid_quantity_supply = max_discharge[0]
-        bid_quantity_demand = max_charge[0]
+        _, bid_quantity_supply = unit.calculate_min_max_discharge(start, end_all)
+        _, bid_quantity_demand = unit.calculate_min_max_charge(start, end_all)
```
maurerle, line 820: why does this change here?
We reverted having varying return values, right? Or has this been left on purpose? :)

tests/test_storage.py
maurerle, line 1: I would revert the changes to this file? As functionaly, nothing should have changed..? Except for removing the product_type param?

tests/test_storage.py
```diff
@@ -329,7 +353,10 @@ def test_set_dispatch_plan(mock_market_config, storage_unit):
     storage_unit.outputs["soc"][start] = 0.5 * storage_unit.max_soc
 
     bids = strategy.calculate_bids(storage_unit, mc, product_tuples=product_tuples)
-    assert len(bids) == 0
+    assert len(bids) == 1
```
maurerle, line 356: This does actually change - why?

assume/common/base.py
mthede: yes, you're right! I missed that one...

assume/strategies/flexable_storage.py
mthede: Mhmm, I'm with you. Especially because in `set_dispatch_plan()` we're currently only updating the `soc` at `next_t` and leave `t` unchanged. 

I'm wondering: what is the general "definition" of `unit.outputs["soc"].at[start]`? 
- Is it the remaining SoC at `start` after selling volumes at different markets? 
- Or is it the SoC at the beginning of `start` before market participating? And the updated SoC is available at the next timestep?

maurerle: The currently (hopefully consistently) used definition is the prior one.
That's why `get_soc_before` was a thing.

So `unit.outputs["soc"].at[start]` is the soc remaining after the time slot.
Seems like this should be tested and documented a little better :/

mthede: Okay, that's good to know. 

About documentation: the unit.outputs[...] is an attribute of each unit. But most of the changes to the values are happening due to bidding strategies and when initializing the unit, it's not really known which outputs are going to be stored. So the outputs-dict is not documented as an attribute of the unit. What would you recommend to document and how? 
A) Document the dict with all possible series and short descriptions in the BaseUnit, although they might not be relevant for the specific unit? (E.g. "soc" for powerplants)
B) I also thought about a table with the series name in the first column (e.g. "energy", "soc", ...), a short description/definition in the second column and the next columns the unit classes and bidding strategies for which this output is relevant. But that might be a pain to keep up-to-date... 

#549 is already onto the issue of missing bidding strategy/clearing mechanism documentation. We should just make sure not to have duplicate, inconsistent documentation.

assume/strategies/flexable_storage.py
nick-harder: @maurerle I believe we need to sit and decide the notion we will be following here. Usually the system dynamics of storage units are described as 
`SOC_{t+1} = SOC_{t} + P_ch_{t} - P_dis_{t}`

And this is the notion I am trying to follow here. So the set dispatch updates the soc for the next time step. and hence here in the bidding strategy I take the soc for the current time step as it should reflect the changes from before. 

Do you agree with such logic?

maurerle: So you would implement this exactly opposite to the discussion below.
Which is `SOC_{t} = SOC_{t-1} + P_ch_{t} - P_dis_{t}` - so the SoC at the end of the timeslot, as currently implemented.
Therefore in `t` one only needs to adjust P_ch_{t}, P_dis_{t} and SOC_{t}.

We can change this behavior of course, but it should be the same everywhere.

assume/strategies/flexable_storage.py
```diff
@@ -177,8 +167,6 @@ def calculate_bids(
             theoretic_SOC += delta_soc
             previous_power = bid_quantity + current_power
 
-        bids = self.remove_empty_bids(bids)
-
```
maurerle: why are we removing this here?
I readded it, as this change is unrelated..?

assume/strategies/flexable_storage.py
maurerle: Well, seems that the unit.outputs[soc] behavior has been changed by @nick-harder in this PR so the second definition:

> Or is it the SoC at the beginning of start before market participating? And the updated SoC is available at the next timestep?

is the correct one now.

tests/test_storage.py
maurerle: It changed due to the removal of clearing empty bids.
It is readded and should be fine.

maurerle (approved):
I added a test for the multi-hour bidding, which checks that the soc is calculated and set correctly.

This PR changes the way the SoC is defined in ASSUME, so this should be noted in the next release.

assume/strategies/flexable_storage.py
nick-harder: @maurerle this was removed since when bids were not sent, the SOC is only updated until the last sent bid and not afterwards. Did you check for such behavior?

assume/strategies/dmas_storage.py
```diff
@@ -144,7 +144,7 @@ def build_model(self, unit: SupportsMinMaxCharge, start: datetime, hour_count: i
         ]
 
         self.model.vol_con = pyo.ConstraintList()
-        v0 = unit.get_soc_before(start)
+        v0 = unit.outputs["soc"].at[start - unit.index.freq]
```
mthede, line 147: in flexable.py the 'unit.get_soc_before(start)' call was changed to 'unit.outputs["soc"].at[start]'. why here to the soc at t-1?

tests/test_storage.py
```diff
@@ -133,7 +133,10 @@ def test_soc_constraint(storage_unit):
     storage_unit.outputs["soc"][start - timedelta(hours=1)] = (
         0.05 * storage_unit.max_soc
     )
-    assert storage_unit.get_soc_before(start) == 0.05 * storage_unit.max_soc
+    assert (
```
mthede, line 136: What are we testing here for? Do we still need it after removing the get_soc_before() method?

assume/strategies/flexable_storage.py
kim-mskw: This behavior occurs due to line 353 in units_operator.py 

` start = timestamp2datetime(last + 1)`

Was shifting this by one second intended? @maurerle, that line seems to be from you.

tests/test_storage.py
```diff
@@ -389,6 +392,54 @@ def test_set_dispatch_plan(mock_market_config, storage_unit):
...
+    start = datetime(2022, 1, 1, 0)
+    for i in range(3):
+        s = datetime(2022, 1, 1, i)
+        end = datetime(2022, 1, 1, i + 1)
+        product_tuples.append((s, end, None))
+
+    mc = mock_market_config
+    strategy = flexableEOMStorage()
+
+    storage_unit.outputs["energy"][start] = 100
```
mthede, line 410: Following the logic ´SOC_{t+1} = SOC_{t} + P_ch_{t} - P_dis_{t}´, don't we need to adjust the next soc too when setting the start energy (as already sold volume)?
I'd suggest calling storage_unit.execute_current_dispatch(start, end), which adjusts the soc for the whole time period according to the existing dispatch to have a consistent starting state.

tests/test_storage.py
```diff
@@ -389,6 +392,54 @@ def test_set_dispatch_plan(mock_market_config, storage_unit):
...
+    storage_unit.outputs["soc"][start] = 0.5 * storage_unit.max_soc
+
+    bids = strategy.calculate_bids(storage_unit, mc, product_tuples=product_tuples)
+    assert len(bids) == 2
+    # the second and third hour have are charging 60 MW
```
mthede, line 415: Generally: I understand the bidding example setting like that:
- price forecast is always 50, average price also 50. "price_forecast[start] >= average_price" favors discharge for all product starts, right?
- max_discharge = 100
- ramp_down_charge = -50
- ramp_down_discharge = 50
- ramp_up_charge = -60
- ramp_up_discharge = 60

why are we then charging in the second and third product start? with still almost 400MWh of remaining SoC, couldn't and shouldn't we continue discharging? no previous volumes were sold in second and third hour. no ramping constraints are violated, if we continue discharging at 100 MW. maybe I'm just confused with the ramp_down/up definitions, although I feel like the documentation is not really up to date

tests/test_storage.py
```diff
@@ -389,6 +392,54 @@ def test_set_dispatch_plan(mock_market_config, storage_unit):
...
+    bids[0]["accepted_volume"] = -60
+    bids[0]["accepted_price"] = 45
+    assert bids[1]["start_time"] == datetime(2022, 1, 1, 2)
+    assert bids[1]["volume"] == -60
+    bids[1]["accepted_volume"] = -60
+    bids[1]["accepted_price"] = 45
+
+    # now dispatch full discharge
+    storage_unit.set_dispatch_plan(mc, bids)
+    storage_unit.execute_current_dispatch(start, end)
```
mthede, line 427: After calling "set_dispatch_plan()", you also call "execute_current_dispatch()". This overwrites the set dispatches before comparing the energy and soc values. Shouldn't we remove the "execute_current_dispatch()" call to check if the set dispatch leads to correctly updated socs before actually dispatching?

assume/strategies/flexable_storage.py
```diff
@@ -58,47 +58,37 @@ def calculate_bids(
...
+        # save a theoretic SOC to calculate the ramping
         start = product_tuples[0][0]
-        end_all = product_tuples[-1][1]
-
+        theoretic_SOC = unit.outputs["soc"].at[start]
         previous_power = unit.get_output_before(start)
 
-        # save a theoretic SOC to calculate the ramping
-        theoretic_SOC = unit.get_soc_before(start)
+        current_power = unit.outputs["energy"].at[start]
```
mthede, line 68: why has this been moved out of the product loop? it now assumes the current_power from the first product start for every other product as well?

tests/test_storage.py
kim-mskw: I agree. It is adjusted

assume/strategies/flexable_storage.py
kim-mskw: I agree it should be moved again in th loop.

kim-mskw (changes requested):
As per my and @mthede comments I will change some things before merging.

maurerle:
I agree with the suggested changes. Do not have time tomorrow. Thursday should be fine

tests/test_storage.py
kim-mskw: We chnaged the test to account for that

kim-mskw (approved):

assume/common/units_operator.py
```diff
@@ -350,6 +350,7 @@ def get_actual_dispatch(
             tuple[list[tuple[datetime, float, str, str]], list[dict]]: market_dispatch and unit_dispatch dataframes
         """
         now = timestamp2datetime(self.context.current_timestamp)
+        # add one second to exculde the frist time stamp, because it is already executed in the last step
```
maurerle, line 353: Thanks :)

assume/strategies/flexable_storage.py
maurerle: If this happens at the start or end of a clearing duration, this probably does not work.

I do not really like to have the constraint for storages to require bidding for each individual hour, otherwise the SoC is wrong..

So this is more a quickfix..

maurerle: I thought a little about it and did not find a good solution yet, so I will extract this into a separate issue we fix in the following days, as removing the "remove_empty_bids" does not fix it for other storage strategies..

PR #559, opened 2025-03-06
549 documentation of bidding strategies and clearing mechanisms is missing

### LINKED ISSUE #549: Documentation of bidding strategies is missing

tiernan-buckley:
# Pull Request

## Related Issue
Closes #549 

## Description
Adds Bidding Strategy description to documentation, renames bidding strategies

## Changes Proposed
- Add Bidding Strategy description to documentation
- Remove "flexable" from the naming of bidding strategies

## Checklist

### DIFF
--- a/docs/source/bidding_strategies.rst
+++ b/docs/source/bidding_strategies.rst
@@ -65,6 +65,10 @@ Naive
 
                                                                        A demand unit can realise a "price-inelastic" demand bid by setting
                                                                        the bid price very high and volume equalling demand at the timepoint.
+  elastic_demand            demand                  EOM                This bidding strategy is formulated for a demand unit to realise a "price-elastic" demand bid,
+                                                                       approximating a marginal utility curve.
+                                                                       
+                                                                       The elasticity can be set to "linear" or "isoelastic".
   naive_profile             powerplant,             EOM, CRM_pos,      Similar to :code:`naive`, however it is a block bid for 24 hours to
                             Energy-supplying        CRM_neg            simulate a bid for the Day-Ahead market, where bid price is set to the marginal cost
                             units (e.g. Fuel Cell)                     at the starting timepoint.


nick-harder:
@maurerle hey Florian, it would be great if you could also look into your DMAS strategies and find a suiting category for them as well as provide a description. We decided to go away from model names in the strategies to more descriptive naming. So we dropped flexable from the name. it would be great to do the same for DMAS.

maurerle (changes requested):
As I read #549 I thought that this is only about the names used in the `bidding_strategies` dictionary, and not about renaming about everything bidding related.
Still fine, though the current state of the PR is hard to review.

It includes:
* removal of tests
* removal of strategies
* renaming of all strategies, which are breaking changes as nothing works as before

Clarifying the names of course makes things more readable.
Though I don't see how `LearningEOMStorage` is better than `StorageRLStrategy`

Having Strategy as a suffix makes clear what this class is and makes it hard to confuse it with a Storage unit.
An RLStrategy was the general name for all learning strategies.
If this should clarify the usage in EOM, than EOMStorageRLStrategy would be a better fit as it is descriptive what this is about.

I would also suggest removing "Standard" as a strategy prefix, as it does not add meaning?
Abbreviating NegativeCRM to NCRM makes things harder to read for new people from my perspective.
If users think that this is an improvement, I am fine with it of course.

In general, it would be good to keep the old names of the strategies in the `bidding_strategies` dictionary for 1 or 2 releases, so that people can migrate more smoothly.

examples/inputs/example_01c/storage_units.csv
```diff
@@ -1,2 +1,2 @@
 name,technology,bidding_EOM,bidding_CRM_pos,bidding_CRM_neg,max_power_charge,max_power_discharge,efficiency_charge,efficiency_discharge,min_soc,max_soc,additional_cost_charge,additional_cost_discharge,natural_inflow,unit_operator
-Storage 1,PSPP,flexable_eom_storage,flexable_pos_crm_storage,flexable_neg_crm_storage,1000.0,992.0,0.86,0.9,0.0,6076.0,0.28,0.28,0.0,Operator 1
+Storage 1,PSPP,eom_storage,pcrm_storage,ncrm_storage,1000.0,992.0,0.86,0.9,0.0,6076.0,0.28,0.28,0.0,Operator 1
```
maurerle, line 2: currently, this should be prefixed with standard right?

nick-harder: @maurerle this is still a draft, it will be fixed later

nick-harder:
@maurerle yes, this PR indeed changes a lot of things, but I believe it is needed since many things are too confusing, or don't have any additional functionality. 

1) Keeping Strategy in the name makes sense, agree
2) LearningEOMStorage gives me directly the info that this is a learning strategy for energy only market for a storage unit, while RLStorageStrategy doesn't contain the info about the designated market. As we develop more methods and add new markets, this will lead to confusion.
3) Agree about the negative and positive in CRM, we should keep neg and pos in the name to make it easier. 
4) The naming notion is Type_Market_Unit, where Type is the bidding strategy type like naive, learning, and unit like power plant and storage if the bidding strategy is specifically defined for that unit type. Do you propose the notion Market_Unit_Type? Our idea was to have the type at the beginning to differentiate between the four types we have now: naive, standard, learning, misc

gugrimm:
I find the naming StrategyType_Market_Unit convincing.
I agree with @maurerle that "Standard" does not add meaning. But have no better idea.

After reading the docu, I still have questions:
- In the table describing the 5 bidding strategy methodologies: Does Naive "tend" to bid marginal cost or does it bid marginal cost?
- In the Naive section: Is the Strategy with bidding_strategy_id "naive" applicable for all units? It states powerplants and demand, but is it applicable for dsm or storage? The naming without unit suggests all units can have this strategy. But if this is the case, we should also state what it does for storage and dms, not only for powerplants and demand.
- In the Naive section: Is the Strategy with bidding_strategy_id "naive" applicable for all markets? I guess not, I think it does not work for the redispatch market for example, users should us naive_redispatch for this. But as no market name appears in the bidding_strategy_id, one could think it does work for redispatch market. I think we should clarify this in the text "not applicable for redispatch market, use naive_redispatch or naive_redispatch_dsm instead." Or did I get sth wrong?

nick-harder:
@gugrimm could you please take another look at the documentation and improve the description, where needed?

nick-harder:
@maurerle could you please also add a description for your DMAS strategies to the docs, so we can conclude this PR? And if you have time to take a other look at the PR, it would be really appreciated

PR #567, opened 2025-03-27
Fix Notebook 9: xRL

### LINKED ISSUE #561: Not returning observations in notebook 9 (09_example_Sim_and_xRL)
When running the notebook in either google colab or locally, I cannot run the notebook.

**Colab** 
I ran the notebook and this is the error I received on the cell where we are running an example case:

(The cell where the error occurs)
Image

(The error message)
Image

I think this error is caused by Google using an outdated version of the _assume_ package.

**Locally** 
I installed the _assume_ package and updated it to the latest version. I originally got the same error as the error that I got in colab but I manually changed the files that were installed for the package. (The changes that I did were taking the files from the _assume_ Git Hub and replacing the files that were installed when I installed the package.) Afterwards, I tried to run the notebook and it gave the following error message:  

Image

I tried to look for the function that would create this path and save the observations as a JSON file, but I couldn't find it.

I would appreciate some advice on overcoming this issue.

nick-harder:
Fix the notebook 9 not working anymore due to updated packages and changes syntax.
- update versions
- update code
- test and apply pre-commit

Closes #561


examples/notebooks/09_example_Sim_and_xRL.ipynb
```diff
@@ -890,8 +875,8 @@
    "outputs": [],
    "source": [
     "!pip install matplotlib\n",
-    "!pip install shap==0.42.1\n",
-    "!pip install scikit-learn==1.3.0"
+    "!pip install shap==0.47.1\n",
+    "!pip install scikit-learn==1.6.1"
```
maurerle, line 879: why is it required to have exact versions here? wouldn't be >= better here?

maurerle (approved):

examples/notebooks/09_example_Sim_and_xRL.ipynb
nick-harder: there is version 1.6.2 which breaks the notebook, hence the specific versions. This helps us also to unsure that these notebooks work longer and do not require so much maintenance.

PR #568, opened 2025-04-02
Improve database handling

nick-harder:
## Changes Proposed
- add rl_meta to store simulation IDs and episodes to speed up grafana dashboards when having many large simulation
- adjust variables queries in Grafana learning dashboard to make use of the rl_meta table and speed up querying
- add a function to list and delete specific simulations from the database to avoid deleting complete database


maurerle (approved):
So this adds new functionality to delete simulations using:

```python
maintenance = DatabaseMaintenance(db_uri=db_uri)
# 1. Retrieve unique simulation IDs:
unique_ids = maintenance.get_unique_simulation_ids()
print("Unique simulation IDs:", unique_ids)
# 2. Delete specific simulations:
maintenance.delete_simulations(["example_01", "example_02"])
# 3. Delete all simulations except a few:
maintenance.delete_all_simulations(exclude=["example_01"])
```

We should add this to the docs as well..?
Currently, this is never called, but I do like to have the availability of this.

PR #569, opened 2025-04-03
Improves documentation for the outputs handling and overall structure

nick-harder:
## Changes Proposed
- enable configuring buffer size for the outputs buffer
- add documentation for the outputs including methods and tables with their columns
- restructure the index for better grouping
- move reply buffer inside RL section for better readability
- change representation of redispatch and DSM units for nicer documentation
- correct level errors and general bugs in documentation
- other small fixes


nick-harder:
@maurerle is pointed out before, here is a PR with the documentation for the outputs class with the table for the outputs structure as well as the new class for managing simulations

docs/source/demand_side_agent.rst
```diff
@@ -38,63 +39,57 @@ Bidding and Optimization
...
 
 Industrial Agent: Steel Plant
-=============================
+-----------------------------
 
 The **Industrial Agent** represents facilities such as steel plants, where key components like electrolysers, DRI plants, hydrogen storage, and electric arc furnaces are modeled. These components form a process flow that optimizes electricity consumption based on the agent’s objectives.
 
-Attributes
-----------
+### Attributes
```
maurerle, line 46: This is not markdown, looks like AI generated conversions gone wrong.

See here for available headlines in RST: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

docs/source/learning_algorithm.rst
```diff
@@ -104,3 +104,6 @@ Other available options for the "policy" include Long-Short-Term Memory (LSTMs).
 Note, that the specific implementation of each network architecture is defined in the corresponding classes in the codebase. You can refer to the implementation of each architecture for more details on how they are implemented.
 
 [2] Y. Ye, D. Qiu, J. Li and G. Strbac, "Multi-Period and Multi-Spatial Equilibrium Analysis in Imperfect Electricity Markets: A Novel Multi-Agent Deep Reinforcement Learning Approach," in IEEE Access, vol. 7, pp. 130515-130529, 2019, doi: 10.1109/ACCESS.2019.2940005.
+
+
+.. include:: buffers.rst
```
maurerle, line 109: Do you think we should have this as an include appended? 
We haven't done this yet, and had each page on its own or as a sub page of existing ones, but we can introduce it here of course.

See https://assume--569.org.readthedocs.build/en/569/learning_algorithm.html

docs/source/outputs.rst
```diff
@@ -0,0 +1,172 @@
...
+
+The ``WriteOutput`` class is responsible for:
+
+- Collecting simulation output messages.
+- Buffering data until either a time-based interval or a memory limit is reached.
+- Converting raw data into structured pandas DataFrames.
+- Storing data into a database and/or exporting it as CSV files.
+- Managing performance trade-offs by allowing configuration of buffer sizes and storage intervals.
+
+Constructor Parameters
```
maurerle, line 23: Having API docs here as well, makes things redundant and harder to maintain.
I would rather just link to the API docs of the WriteOutput class, which currently is here:
https://assume--569.org.readthedocs.build/en/569/assume.common.html#module-assume.common.outputs

maurerle (changes requested):

docs/source/learning_algorithm.rst
nick-harder: @maurerle I don't see why not. we can keep the documentation in separate files, but so that structure is better, I believe we can use include. we do it in latex, why not here as well. Is there a reason against it?

nick-harder:
@maurerle thanks for the comments, good ones, now fixed

docs/source/demand_side_agent.rst
```diff
@@ -32,18 +53,20 @@ Each component (e.g., Electrolyser, Heat Pump, DRI Plant) is modeled in detail w
...
+For full references see :mod:`assume.units.dst_components`.
+
 Bidding and Optimization
 ------------------------
 
 Once the DSU is defined, it can participate in the electricity market by optimizing its operation to either minimize costs or maximize flexibility. The bidding strategy depends on the agent type and market conditions.
 
 Industrial Agent: Steel Plant
-=============================
+-----------------------------
```
maurerle, line 64: most of this is still Attributes which look like they duplicate the docstrings.. But well, it was the case before as well..

docs/source/learning_algorithm.rst
maurerle: It's not a problem, I just liked to know, that to edit the learning_algorithm docs, I have to look into the `learning_algorithm.rst`, but now this is a little more ambigous.

Following the rational: if it is too long for one rst source file - it might be too long for one docs file as well..?

That's what is different from latex right? :)

docs/source/buffers.rst
```diff
@@ -2,9 +2,9 @@
 ..
 .. SPDX-License-Identifier: AGPL-3.0-or-later
 
-###############################
-Buffers
-###############################
+#############
+Reply Buffer
```
maurerle, line 6: Isn't it a replay buffer?

maurerle (approved):
Good to have the documentation on this improved!
Though I mentioned some concerns you can address if you like :)

docs/source/buffers.rst
nick-harder: hahaha funny mistake

PR #574, opened 2025-04-14
Enable changing number of learning agents in continue_learning mode

nick-harder:
# Pull Request

## Description
This PR allows us to change the number of learning agents between training runs while still reusing previously trained critics in `continue_learning` mode.

Useful for scenarios like training power plants first and later adding storage units or adjusting the number of agents.

## Changes Proposed
- Add weight transfer logic to adapt critic weights when agent count changes
- Perform architecture check before loading critics; if mismatched, safely transfer compatible weights
- Transfer logic copies:
  - Shared observation weights
  - Unique per-agent obs and action weights for overlapping agents only
- Add `strategy.loaded` flag to mark strategies loaded from file
- In `continue_learning`, disable initial exploration only for loaded strategies
- Move common attributes of learning strategies into base class

## Testing
Tested on RL cases by changing the number of units in the simulation

## Checklist

### DIFF
--- a/assume/reinforcement_learning/algorithms/matd3.py
+++ b/assume/reinforcement_learning/algorithms/matd3.py
@@ -10,7 +10,11 @@ from torch.nn import functional as F
 from torch.optim import AdamW
 
 from assume.reinforcement_learning.algorithms.base_algorithm import RLAlgorithm
-from assume.reinforcement_learning.learning_utils import polyak_update
+from assume.reinforcement_learning.learning_utils import (
+    infer_n_agents,
+    polyak_update,
+    transfer_weights,
+)
 from assume.reinforcement_learning.neural_network_architecture import CriticTD3
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


nick-harder:
@kim-mskw ready for review

kim-mskw (commented):
Cool approach! Does it make sense to copy the weights of other agents to the newly added agents? Just so that they have a warm start kind of?

assume/reinforcement_learning/algorithms/matd3.py
```diff
@@ -227,24 +335,21 @@ def create_actors(self) -> None:
...
         """
 
-        obs_dim_list = []
-        act_dim_list = []
-
         for strategy in self.learning_role.rl_strats.values():
             strategy.actor = self.actor_architecture_class(
-                obs_dim=strategy.obs_dim,
-                act_dim=strategy.act_dim,
+                obs_dim=self.obs_dim,
```
kim-mskw, line 340: Why did you move that to the algorithm instead of the startegy? Could it not bee that they differ?

assume/reinforcement_learning/algorithms/matd3.py
```diff
@@ -257,20 +362,7 @@ def create_actors(self) -> None:
                 ),  # 1=100% of simulation remaining, uses learning_rate from config as starting point
             )
 
-            obs_dim_list.append(strategy.obs_dim)
-            act_dim_list.append(strategy.act_dim)
-
-        if len(set(obs_dim_list)) > 1:
-            raise ValueError(
-                "All observation dimensions must be the same for all RL agents"
```
kim-mskw, line 265: Same here, then we would still need that

assume/reinforcement_learning/algorithms/matd3.py
nick-harder: @kim-mskw For MATD3, it is not allowed for the strategies to differ in our current implementation. Previously, we had a check that raised an error if there was a dimension mismatch, but this check occurred only after the strategies had already been created, which did not make sense. Now, the check is performed before the creation of the strategies, and the error is raised earlier if necessary. If everything is correct, the centralized values are used for the strategies to ensure robustness.

assume/reinforcement_learning/algorithms/matd3.py
nick-harder: @kim-mskw this is now done in a separate function self.check_policy_dimensions()

nick-harder:
@kim-mskw This is not an ideal solution, but it works quite well. The primary use case is to first train the power plants individually and then later activate the storage units. I tested this in a simulation with 10 power plants and 4 storage units, and it performed well. The entire approach relies on the fact that, in ASSUME, we always add power plants first, followed by storage units. Essentially, the weights are transferred for the units that have already been trained, while the new connections for the newly added agents (storage units) remain as initialized. 

In my example, when adding the 4 storage units, the part of the new critic that covers the power plants is exactly the same as the old critic, and the newly added connections in the first layer are newly initialized. Since the deeper layers in our architecture do not depend on the number of agents (as long as the number of agents doesn’t change significantly), we keep those layers unchanged to provide a good starting point for the agents.

image

assume/reinforcement_learning/algorithms/matd3.py
kim-mskw: Fair, okay thanks for the clarification.

kim-mskw:
> @kim-mskw This is not an ideal solution, but it works quite well. The primary use case is to first train the power plants individually and then later activate the storage units. I tested this in a simulation with 10 power plants and 4 storage units, and it performed well. The entire approach relies on the fact that, in ASSUME, we always add power plants first, followed by storage units. Essentially, the weights are transferred for the units that have already been trained, while the new connections for the newly added agents (storage units) remain as initialized.
> 
> In my example, when adding the 4 storage units, the part of the new critic that covers the power plants is exactly the same as the old critic, and the newly added connections in the first layer are newly initialized. Since the deeper layers in our architecture do not depend on the number of agents (as long as the number of agents doesn’t change significantly), we keep those layers unchanged to provide a good starting point for the agents.
> 
> image

Oka understood. I mainly asked, because if you consecutively add powerplants, it might make sense to copy the weight for newly added unique obs from old unique_obs, but if you add storages, the weights are most likely not helpful. So then I get that. Then also throwing out the entire initial layer makes sense as we add these conceptually different unqiue _obs.

nick-harder:
@kim-mskw for the newly added units, the weights are not copied; they are kept as initialized. The orange is only for power plants from scratch, and the orange continues learning with storage units added

nick-harder:
@kim-mskw I have now made the method more robust. Now a map for the order of the units is stored when saving the critics. This map is loaded when loading critics, and it checks the length and the order. It now uses this map for proper weight transfer, so a unit can be deleted, added in the middle or the beginning of the units file, and the weight transfer logic will be robust.

kim-mskw (approved):
made some renaming and added/changed comments and docu, but otherwise looks good!

PR #575, opened 2025-04-16
Add elastic demand

nick-harder:
# Pull Request

## Description
This PR introduces an elastic demand bidding strategy by @gugrimm.

## Changes Proposed
- Developed a new bidding strategy
- Adjusted how an empty demand_df is treated. Now it is filled out with zeros and logs a warning instead of raising an error
- Added extra parameters to the demand unit
- Fixed how actual dispatch for the demand unit is calculated. Now it uses the output["energy"] instead of simply relying on the volume as it was before. The volume might not be fully fulfilled, and this would lead to wrong data logging. Now this is fixed.

## Testing
Added tests for the new bidding strategy. Also added checks inside the demand unit in case elasticity is provided.

## Checklist


kim-mskw (commented):
Can you elabortae a bit on the find_first_block_bid() function? How can Power * price be a volume? 
Also with first you mean highest yes?

nick-harder:
@gugrimm could you please add a better explanation for the way the first bid is derived as Kim suggested?

gugrimm:
The thing with the units is: this is not physics and units do not necessarly need to fit. It took me a while to understand, but what we are looking is a relation between quantity and price, a behavioral relationship based on some economic theory. As far as i understand, in this case units do not match dimensions, unlike in physics where units of an equation always fit. In our case the behavioral relationship must be enriched with knowledge "of the x and y axis" to make sense in real world...

gugrimm (dismissed):

kim-mskw:
Thanks for the clarification. I think I got it now. It works like this right? 
image

kim-mskw (changes requested):
Looks good! Two small things.
Should we maybe rename it to flexible_demand? As the other strategy is called inflexible, or we switch it to inelastic? 
And can you maybe add documentation to this demand type?

gugrimm:
The picture you drew is exactly how it works, yes! I don't get why you call it "isometric" demand - probably a mistake? "Isoelastic" is the correct term.
In the naming I believe we should use elastic and inelastic. Both terms show this has sth to do with economics / marginal utility, or at least these are the terms used in the literature I read on elasticities. Flexible and inflexible on the other hand side refers to time-shifts (as I understand it). So our demand side agents are "flexible" in my understanding. I changed the docu accordingly.
I documented the bidding strategy in the PR #559 .

kim-mskw (approved):

PR #578, opened 2025-05-05
Fixes spelling and formatting mistakes introduces with the last PR

nick-harder:
The last PR has introduces several spelling and formatting mistakes. This fixes it


nick-harder:
@kim-mskw I believe your pre-commit is outdated, maybe you should delete and upgrade it, as it would have caught such mistakes

maurerle (approved):

.pre-commit-config.yaml
```diff
@@ -30,3 +30,7 @@ repos:
     -   id: codespell
         types_or: [python, rst, markdown]
         files: ^(assume|docs|tests)/
+        args:
+          # comma-separated list of words you want codespell to skip
+          - --ignore-words-list=DED,esy,flexable,Meger,ore
+          - --ignore-regex=https?://\S+
```
maurerle, line 36: Doesn't it work to have this here?
https://github.com/assume-framework/assume/blob/main/pyproject.toml#L152

nick-harder:
@maurerle I have adjusted the test workflow to run the full pre-commit workflow. Fine by you?

.pre-commit-config.yaml
nick-harder: @maurerle idk why, but it fails without this argument. It's stupid to keep at at two places

nick-harder: @maurerle ok fixed by pointing to tomli

maurerle: Well yes.. Since python 3.11 tomllib is part of the standard library, so it failed for python 3.10.
Looks fine

nick-harder: @maurerle yeah maybe we should already drop 3.10 at some point :-)

PR #579, opened 2025-05-05
Improve database structure

nick-harder:
# Pull Request

## Description

This PR addresses two key pain points:

* **Slow startup**: Deleting millions of rows between learning runs can take \~5 minutes before a new simulation even begins.
* **Grafana errors**: Missing tables or columns on first startup cause dashboard panels to raise errors.

We solve these by:

1. **Partitioning** all time-series tables by `simulation` so dropping a partition is an **O(1)** metadata operation, instantly clearing old data.
2. **Bootstrapping** the full schema at container init (using `REAL` types and enabling `timescaledb`/`postgis`) so Grafana never sees missing objects.
3. **Bulk writes** via a single `COPY … FROM STDIN` per table inside one transaction, replacing thousands of INSERTs for multi× speedup.

These changes bring **a speed-up of around 25 to 30%** to the large non-learning simulations by increasing the write speed. And this allows to start new learning simulations much faster without the wait before the simulation. **This is true only when using TimescaleDB and not the local DB.**

_Performance:_ Example 'base_case_2019' for a full year, current time: 14:22, new time: 10:22.

I also fixe a small issue with the unit operator still submitting tensors when not in learning mode, which can cause issues in the future as the behavior of tensors within an optimization algorithm is unpredictable.

## Changes Proposed

* **Docker-init schema** (`docker_configs/db-init/assume_schema.sql`): pre-creates every table with `REAL` precision and required extensions, eliminating Grafana errors on first use.
* **LIST-partitioned tables**: `market_meta`, `market_dispatch`, `unit_dispatch`, `rl_params`, `grid_flows`, `kpis` now use `PARTITION BY LIST (simulation)`, and per-run partitions are created/dropped for instant deletes.
* **Bulk COPY writes**: Refactored `store_dfs` to stream DataFrames via `COPY … FROM STDIN` in a single transaction, cutting write overhead by an order of magnitude.
* **Dynamic table creation**: Catch `NoSuchTableError` to auto-bootstrap new output tables with `df.head(0).to_sql()`, so no manual schema changes are ever needed.

## Testing
Tested using timescaleDB and localdb, all code work fine. Grafana is also operational.

## Checklist


maurerle:
Very cool that you are looking to improve the performance.

We currently are not using some crucial features of timescaledb - which are the hypertables - which is basically what you try to achieve, but partition by time.
Furthermore I would rather investigate having an index on the simulation_id.
So I don't think this PR is the best long term solution.

I am currently quite sparse on time, but would try to file a PR with a proper solution in the next weeks..

nick-harder:
@maurerle hey, thanks! I considered these hypertables but because they are using the time chunks I decided against that idea. Also the main thing here is to introduce a proper database schema with proper keys and rules so we can see if something is broken and being logged somehow twice and so on. Now all entries are unique and it is actively controlled by the database if some double values are added, which should not be the case. Also, I have introduced the indexing for different tables, which should improve the speed of the dashboards. Please take a look at the schema when you have time.

PR #583, opened 2025-05-13
Prepare for release and update version and release notes

nick-harder:
So now I believe is a good time to release the version with all the fixes and new things we have.


kim-mskw:
@nick-harder, maybe we can quickly finish the documentation PR #549 beforehand?

nick-harder:
@kim-mskw Since that one changes too many things, I would prefer to keep it for a major release 0.6.0, which we would introduce later. What do you think?

kim-mskw:
Okay fair! That does make sense. It would be great if we could quickly implement the fix proposed in #585 though. It is not a major bug, but it would be nice to have it fixed.

maurerle:
@kim-mskw let's go? :)

nick-harder:
@maurerle I believe Kim is very busy, so let's do it ourselves?

PR #585, opened 2025-05-15
fix overwriting clearing price of rejected orders

### LINKED ISSUE #584: Pay-as-clear mechanism with multiple products overwrites MCP of rejected orders
If multiple products are available (e.g. "duration": "1h", "count": 24, resembling DAM-bidding), [order["accepted_price"] is overwritten by the clearing price of the last product for all rejected orders](https://github.com/assume-framework/assume/blob/c7c132cb248eee6c7ac408cd11fe110c1e1f89db/assume/markets/clearing_algorithms/simple.py#L169C17-L169C54). 

This leads to incorrect values shown in the dashboard and they are also later on used for reward calculations. This may affect opportunity cost calculations for learning strategies. 

I'd suggest introducing a product-specific dict like [`accepted_product_orders` ](https://github.com/assume-framework/assume/blob/c7c132cb248eee6c7ac408cd11fe110c1e1f89db/assume/markets/clearing_algorithms/simple.py#L161) and adding them one at a time to the final orderbook.

mthede:
# Pull Request

## Related Issue
Closes #584

## Description
Handle rejected orders for each product seperately to avoid overwriting the clearing price.

This is required to have a reward in RL for rejected bids as well

## Changes Proposed
- Introduce product-specific dict for rejected orders

## Testing
Tested it in the context of current WIP on day-ahead market bidding for learning strategies on example 02a. Checked that all orders of the same product start have the same accepted price. 

## Checklist

### DIFF
--- a/assume/markets/base_market.py
+++ b/assume/markets/base_market.py
@@ -651,7 +651,19 @@ class MarketRole(MarketMechanism, Role):
                     }
                 else:
                     order["accepted_volume"] = 0.0
-                    order["accepted_price"] = 0.0
+                    # TODO entry is generally not needed
+                    # it is of interest to have a reward in RL for rejected bids
+                    # the matching of meta to product by start_time is sufficient in most cases
+                    # but might not be correct if multiple orders with same start (but different end or zone)
+                    # exist - in these cases the rejected_bids should be set in the clearing itself
+                    order["accepted_price"] = next(
+                        (
+                            product["price"]
+                            for product in market_meta
+                            if product.get("product_start") == order["start_time"]
+                        ),
+                        0,
+                    )
 
         self.open_auctions - set(market_products)
 


assume/markets/base_market.py
```diff
@@ -652,6 +652,16 @@ async def clear_market(self, market_products: list[MarketProduct]):
                 else:
                     order["accepted_volume"] = 0.0
                     order["accepted_price"] = market_meta[0]["price"]
+                    # TODO: shouldn't this be dependent on the product start for multiple products?
```
mthede, line 655: I'm not sure if this is fully related to the issue, but a while ago I had an issue with this line. If multiple products are available, shouldn't the set accepted price be matched to the right product_start/start_time? But maybe I didn't fully understand for which cases @nick-harder included this.

kim-mskw: To my it is also not apparent what this is supposed to do. @nick-harder do you rememberthe intention here?

kim-mskw (approved):

assume/markets/base_market.py
maurerle: This was added so that **every** bid has the fields "accepted_volume" and "accepted_price" - as a backup if the market mechanism does not implement this for the rejected bids (such as `redispatch`, `pay_as_bid` or `pay_as_clear_complex_dmas`).
The current implementation is wrong, but it would be also okay to just set the `accepted_price` to just `price` or 0..?

@mthede why did you need the accepted_price of rejected bids in the first place? :)

maurerle: So I would suggest to set the accepted_price here to 0 and remove the comments - otherwise this PR looks fine!
Thanks :)

mthede: Ahh, alright got it. I looked into the other clearing mechanisms and market meta outputs and I think it makes sense to stick to the current implementation. Actually, for multiple products, line 648 should match it correctly. I'm really not sure anymore how exactly I ran into this issue. So I'll just delete the comment and leave it as it is. Or why do you think the current implementation is wrong and you'd prefer 0 as a fallback?

mthede: For rejected bids, the reward calculation of learning agents uses the MCP (stored as `accepted_price` in the orderbook, e.g. [here](https://github.com/assume-framework/assume/blob/c7c132cb248eee6c7ac408cd11fe110c1e1f89db/assume/strategies/learning_strategies.py#L1028) and [here](https://github.com/assume-framework/assume/blob/c7c132cb248eee6c7ac408cd11fe110c1e1f89db/assume/strategies/learning_strategies.py#L555C33-L555C34)) for the opportunity costs if more capacity could've been sold. So the incorrectly calculated regret for rejected orders doesn't give the right incentive.

nick-harder:
hey @mthede, so now I am a bit lost, is this a needed fix or not? what was exactly the problem? Somehow from the code changes I don't see directly what behavior is different now

mthede:
yes, the current changes in simple.py are still necessary to ensure that the correct acceped_price (the MCP) is set for rejected orders in a setting with multiple products (e.g. 24x hourly products for DAM-bidding). otherwise the whole rejected_orders for all product times way overwritten by the clear_price of the last product.

the second discussion above was about the "fallback" for other market mechanisms where the accepted_price is not directly set in the clearing function. I was wondering, if the intented value is set correctly but removed my comments on that because I think is okay the way it is right now.

assume/markets/base_market.py
maurerle: This value should never be needed.
When calculating anything on the bids, one should only consider the accepted ones (e.g. accepted_volume != 0) the price is only given to have an accepted price at all, but it does not make any sense, as the bid is not accepted..
Using the market clearing price - or the first market_meta price for all bids does not make sense here.

I did set the value to 0 so that the inaccuracy with a useless value does not happen anymore.
Any other comments to that?
Otherwise it looks fine

maurerle: But @mthede - why are you looking for the accepted_price of rejected bids? Can't you just filter these out by using `accepted_volume != 0`?

kim-mskw: @mthede and @maurerle As discussed in the dev meeting, we need the clearing_price, and as the agent does not get the market_meta, the only way to retrieve the actual clearing_price is. Hence, we decided to put the accepted_price with not zero but indeed the clearing price.  

Also to make it consitent between complex clearing and simple clearing we decided to have the accepted price to be the market price.

PR #590, opened 2025-06-18
Real time Simulation

maurerle:
# Pull Request

This adds the option to simulate real time if the start and end of a simulation is set correctly.
The main change required for this is to set the `real_time` param in the setup of the world.

```
world.setup(
        start=start,
        end=end,
        save_frequency_hours=1,
        simulation_id=simulation_id,
        # set real_time to true here
        real_time=True,
    )
```

This PR also fixes an issue with formulas in docstrings which did not render very well in `find_first_block_bid` and a broken ref in the learning documentation.

## Testing
run `python examples/world_rt.py`
adding automatic tests for a real time simulation is not easily possible.

## Checklist


docs/source/realtime_simulation.rst
```diff
@@ -0,0 +1,35 @@
...
+        # set real_time to true here
+        real_time=True,
+    )
+
+The start of the simulation should be configured to be the next full minute.
+Note that the time must be provided in UTC as every timestamp in mango is interpreted as UTC/zulu time to be timezone agnostic.
+
+This simulation starts directly and ends after one hour.
+
+Take a look at the `world_rt.py` example for a full working real time simulation which can be run using
```
nick-harder: can this be a link to the file directly on github? otherwise it is hard to find this file.

nick-harder: Also, why not also embed the real_time argument into existing scenario_loaders? Is there some incompatibility why we cannot use normal folder loader for example with real time simulation setting? Having the real_time argument only in the setup, which is always used only internally by the loaders, which would be the most probable way for users also to load scenarios, is somehow very limited and can be seen more like just an internal feature.

examples/world_rt.py
```diff
@@ -0,0 +1,100 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
```
nick-harder: I believe this file is not really needed since we already have a world_script example, where we can simply add a comment about the real_time in the setup and how to switch it on.

nick-harder (changes requested):
I have left some comments, which may be helpful for us to keep the codebase more sleek and also more user friendly for the future

docs/source/realtime_simulation.rst
maurerle: The main difference is that we need to set the start and end according to the current time:

```python
start_tz = datetime.now() + rd(minutes=1, second=0, microsecond=0)
# start should be UTC
start = timestamp2datetime(start_tz.timestamp())
end = start + rd(hours=1)
```

Thats also why the existing `world_example.py` does not work easily.

Another option would be to add an offset somehow.
But yes, in general this is more a feature for fine-grained distributed simulations and not for your every day ASSUME simulation.

examples/world_rt.py
kim-mskw: Discussion with @maurerle, we want the script to be out-of-the-box runnable and not clutter the original world file. Yet to make sure it is the rela time simulation we move it into an extra folder similar to the distirbuted simulation, so that no one mistakes it for the actual world file

kim-mskw (approved):

Issue #525, opened 2025-01-02
SoC name is misleading as it is used with absolute values


maurerle:
When I hear SoC - I think of a Percent Value between 0 and 1 (or 0 and 100) as also given in https://en.wikipedia.org/wiki/State_of_charge

However, we went from relative SoC to absolute SoC in 

> switched from using SOC in the storage unit as relative value to absolute value to avoid confusions

#464

I'd like to reiterate if there is a better way to handle this - eventually renaming it to something else than SoC..? Or going back to relative values.

Not a big issue, but still improvable for end users.

nick-harder:
@maurerle yes true! maybe we can rename it all to absolute_SOC to avoid confusion?

PR #591, opened 2025-06-22
release notes and minor bug fix

Manish-Khanra:
- Introduced a `ThermalStorage` class that extends `GenericStorage` to support both short-term (freely cycling) and long-term (schedule-driven) storage operation. Long-term mode allows users to define a binary schedule to restrict charging and discharging to specific periods, enabling realistic modeling of industrial or seasonal thermal storage behavior.
  - To use this feature, set `storage_type` to `"long-term"` and provide a `storage_schedule_profile` (0: charging allowed, 1: discharging allowed).

**Improvements:**

- The initial SOC is now explicitly enforced as a model constraint for all storage units. This guarantees that simulations always start from the intended state, ensuring scientific reproducibility and correctness.
- **Enhanced Test Suite for Storage Units:** Comprehensive unit tests have been added for both `GenericStorage` and `ThermalStorage`, including short-term and long-term scheduling, efficiency losses, ramping, power limits, schedule adherence, and initial SOC.

**Bug Fixes:**

- Fixed an error in the schedule constraint logic for long-term storage. The model now strictly enforces that charging and discharging can only occur during their respective scheduled periods.
- Fixed a bug where the initial SOC could be ignored or set to an unintended value by the solver. The initial SOC is now always fixed at model initialization.
- Added a missing upper bound constraint to `natural_gas_in` in the `Boiler` component to ensure that the fuel input never exceeds the specified maximum power. This change prevents the solver from assigning unphysical input values and brings consistency with the handling of other fuel types.

### DIFF
--- a/assume/units/dst_components.py
+++ b/assume/units/dst_components.py
@@ -198,9 +198,9 @@ class Boiler:
         self.initial_operational_status = initial_operational_status
         self.kwargs = kwargs
 
-        if self.fuel_type not in ["electricity", "natural_gas"]:
+        if self.fuel_type not in ["electricity", "natural_gas", "hydrogen_gas"]:
             raise ValueError(
-                "Unsupported fuel_type for a boiler. Choose 'electricity' or 'natural_gas'."
+                "Unsupported fuel_type for a boiler. Choose 'electricity' or 'natural_gas' or 'hydrogen_gas'."
             )
 
     def add_to_model(
@@ -257,6 +257,11 @@ class Boiler:
                 raise ValueError(
                     "Natural gas boiler requires a natural gas price profile in the model."
                 )
+        elif self.fuel_type == "hydrogen_gas":
+            if not hasattr(model, "hydrogen_gas_price"):
+                raise ValueError(
+                    "Hydrogen gas boiler requires a hydrogen gas price profile in the model."
+                )
 
         # Define parameters
         model_block.max_power = pyo.Param(initialize=self.max_power)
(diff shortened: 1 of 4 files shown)


assume/units/dst_components.py
```diff
@@ -415,7 +466,7 @@ def __init__(
         )
         self.efficiency_charge = efficiency_charge
         self.efficiency_discharge = efficiency_discharge
-        self.initial_soc = initial_soc
+        self.initial_soc = initial_soc_frac * max_capacity
```
maurerle, line 469: Here we again introduce SoC as a real value, while it is typically used as percentage 0-1 or 0-100 in literature..
See #525 - it is fine and consistent to do this here, but I somehow do not like that it is like that currently

docs/source/release_notes.rst
```diff
@@ -12,34 +12,28 @@ Upcoming Release
...
-  variance in the data, but a note was added to explain this.
-- **Changed market price in rejected orders**: Fixed a bug where the wrong market price was written in the rejected orders, namely any auction with more than 1 product had the price of the last product written as the
-  market price instead of the price of the respective hour. This was, however, only a mistake for the rejected orders.
-
+- **Correct Schedule Enforcement in ThermalStorage:** Fixed an error in the schedule constraint logic for long-term storage. The model now strictly enforces that charging and discharging can only occur during their respective scheduled periods.
+- **Initial SOC Bug:** Fixed a bug where the initial SOC could be ignored or set to an unintended value by the solver. The initial SOC is now always fixed at model initialization.
+- **Boiler Upper Bound:** Added a missing upper bound constraint to `natural_gas_in` in the `Boiler` component to ensure that the fuel input never exceeds the specified maximum power. This change prevents the solver from assigning unphysical input values and brings consistency with the handling of other fuel types.
```
maurerle, line 33: The release notes should not remove/replace existing past release notes

Manish-Khanra: Thank you for pointing this out. The previously removed release notes have now been restored.

assume/units/dst_components.py
Manish-Khanra: accepted your changes!

docs/source/release_notes.rst
```diff
@@ -12,6 +12,29 @@ Upcoming Release
   The features in this section are not released yet, but will be part of the next release! To use the features already you have to install the main branch,
   e.g. ``pip install git+https://github.com/assume-framework/assume``
 
+0.5.4 - (23th June 2025)
```
maurerle: This is not released yet.. I would just remove the headline

maurerle (approved):
I fixed the release notes.
I did not fully understand all tests, but I think this is fine for now.

PR #595, opened 2025-07-01
Revise storage energy costs for learning

kim-mskw:
# Pull Request

## Description
This PR changes the rationale behind the calculation of energy costs for the learning unit. Previously, the energy cost represented the net economic value of stored energy, which included charging expenditures and subtracted revenues from discharging. Now, the calculation is based purely on acquisition cost, i.e., the weighted average of past charging prices.

The acquisition cost rationale only updates the stored energy cost when energy is added. Discharging does not affect the cost value of the remaining energy. This prevents the energy cost from becoming negative—a case that can occur under the economic value approach when cumulative discharge revenues exceed acquisition costs. A negative cost, however, is not a meaningful signal for decision-making: just because past revenues were high does not mean that the remaining energy should be given away below its purchase cost.

Tests with both rationales (by @kim-mskw using example 02e) show that using acquisition cost leads to SHAP values that clearly identify cost_stored_energy as an important feature. In contrast, the economic value rationale obscures this relationship, suggesting that the acquisition cost representation better supports interpretability.

## Changes Proposed
- Rename energy_cost to cost_stored_energy
- Added tests
- Switched from economic value to acquisition cost rationale

## Testing
Tested using example_02_e with two storage units and both energy cost calculation methods. Each was run over 60 learning episodes, followed by SHAP value analysis.

## Checklist

**Implementation now: Acquisitions costs**
image

**Implementation before: Economic value**
image

### DIFF
--- a/tests/test_flexable_storage_strategies.py
+++ b/tests/test_flexable_storage_strategies.py
@@ -167,7 +167,7 @@ def test_flexable_pos_crm_storage(mock_market_config, storage):
     product_tuples = [(start, end, None)]
 
     # constant price of 50
-    specific_revenue = (50 - (4 / 0.95)) * 360 / (0.36 * 1000)
+    specific_revenue = (50 - 4) * 360 / (0.36 * 1000)
 
     storage.forecaster = NaiveForecast(index, availability=1, price_forecast=50)
     bids = strategy.calculate_bids(storage, mc, product_tuples=product_tuples)
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -89,7 +89,7 @@ def test_calculate_operational_window(storage_unit):
 
     assert min_power_discharge[0] == 0
     assert max_power_discharge[0] == 100
-    assert cost_discharge == 4 / 0.95
+    assert cost_discharge == 4
 
     min_power_charge, max_power_charge = storage_unit.calculate_min_max_charge(
         start, end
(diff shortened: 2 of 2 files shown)


assume/strategies/learning_strategies.py
```diff
@@ -1032,15 +1032,25 @@ def calculate_reward(
...
         # Calculate and clip the energy cost for the start time
+        # cost_stored_energy = average volume-weighted procurement costs of the currently stored energy
         if next_soc < 1:
-            unit.outputs["energy_cost"].at[next_time] = 0
+            unit.outputs["cost_stored_energy"].at[next_time] = 0
+        elif accepted_volume < 0:
+            # increase costs of current SoC by price for buying energy
+            unit.outputs["cost_stored_energy"].at[next_time] = (
+                unit.outputs["cost_stored_energy"].at[start] * current_soc
+                - (accepted_price + marginal_cost) * accepted_volume * duration_hours
```
mthede, line 1043: The marginal costs for storage units are already efficiency corrected, whereas the accepted_price and accepted_volume are not. The charging efficiency losses are taken into account by dividing by the actual next_soc. Is this consistent? Do we need to account for the roundtrip efficiencies, so including charging losses as well?

I tried to think about how to correctly include the different costs and efficiencies, but my brain isn't braining right now 🥵☀️

kim-mskw: Is the error of not having the MC higher than that of having only the MC efficiency corrected, right? Perhaps we could simply note that they are a proxy and leave it as is? What do you think?

Issue #619, opened 2025-07-23
Error when running 10_DSU_and_flexibility


igarizio:
When running this example ([10_DSU_and_flexibility](https://github.com/assume-framework/assume/blob/main/examples/notebooks/10_DSU_and_flexibility.ipynb)) the code fails in the section "cleaning dispatch data from the simulation to feed it to the PyPSA network". The actual [line that fails is](https://github.com/assume-framework/assume/blob/main/examples/notebooks/10_DSU_and_flexibility.ipynb?short_path=9c34160#L2811) (the next cell contains the same problem):
`redisp_p_set = redisp_p_set.drop(columns=["demand_EOM1", "demand_EOM2"])`
> KeyError: "['demand_EOM1', 'demand_EOM2'] not found in axis"

I think this is just because the previous line creates an empty dataframe:
```python
redispatches = market_dispatch_total[
    market_dispatch_total["market_id"].str.contains("redispatch", case=False, na=False)
]
```
Is it possible that `market_dispatch_total["market_id"]` should be `market_dispatch_total["simulation"]`?

maurerle:
Hi @igarizio ,

I tried again today to reproduce this issue but could not do this in any of my older python envs.
Maybe reinstalling `pandas`, `numpy` and/or `pypsa` did help somehow..?

If you can still reproduce this issue, can you send a `pip freeze` output of your env and maybe try if it works in a fresh env on your side?

Thank you very much for your efforts!

igarizio:
Hi @maurerle, thanks for looking into this!

Sorry, I should have given you way more details when creating this issue (I assumed it was just a column name mix-up, but I think it's something else).
I was running this example in colab (from the link provided [here](https://assume.readthedocs.io/en/latest/examples/10_DSU_and_flexibility.html)). This is what I did: I ran the first cell, waited until it finished, restarted the session, and ran everything again. I saved the outputs [here](https://colab.research.google.com/drive/1unmegSsIabM_kWuxb-_A1cFE4UKa_sc0#scrollTo=CfQv2GupDkKa).

As I mentioned above, the error that killed the notebook was `KeyError: "['demand_EOM1', 'demand_EOM2'] not found in axis"` but I had not realized that the previous cell had a silent error (I believe from `world.run()`). Here's that error:
```
[...]

PR #620, opened 2025-07-30
add notebook 10 to CI for functional validation

maurerle:
This adds Notebook 10 to the CI


maurerle:
This is here to show that #619 works fine with the latest version installed.
It generally makes sense to have one of the later notebooks tested in CI as well.

I don't want to add everything though, to not increase the CI time that much :|

PR #621, opened 2025-08-01
improve check for available solvers

maurerle:
## Description
As we nowadays directly depend on Pyomo, we should use a helper function to check for the available pyomo solvers.
That way, we reduce rewriting the code.

## Changes Proposed
- streamline pyomo solver functions


kim-mskw:
Good suggestion for reducing redundancies!

kim-mskw:
Tested it with zonal (`small_with_zonal_clearing`) and redispatch example (`small_with_zonal_clearing`)

PR #628, opened 2025-08-13
Added case for non accepted order in dispatch plan for base unit

ChriKo97:
# Pull Request

## Description
Adds new case for not accepted orders preventing errors in set_dispatch_plan method of baseUnit

## Changes Proposed
- Currently, not accepted orders returned to unit lead to error in dispatch plan, as "accepted_price" key is not set in order object
- New case prevents this error and sets accepted price to 0.

## Testing
No testing

## Checklist

### DIFF
--- a/assume/common/base.py
+++ b/assume/common/base.py
@@ -164,7 +164,9 @@ class BaseUnit:
             self.outputs[product_type].loc[start:end_excl] += added_volume
 
             # Get the accepted price and store it in the outputs
-            if isinstance(order["accepted_price"], dict):
+            if "accepted_price" not in order:
+                accepted_price = 0
+            elif isinstance(order["accepted_price"], dict):
                 accepted_price = list(order["accepted_price"].values())
             else:
                 accepted_price = order["accepted_price"]


ChriKo97:
Hi @kim-mskw,

currently the accepted_price needs to be set in the clearing mechanism. @maurerle and I decided that it may be useful to allow to not set the accepted_price in the clearing mechanism, as it could be misleading (order is not accepted but has an accepted price). The proposed change in the pull request would allow new clearing mechanism to omit the "accepted_price" key in the returned orderbook

kim-mskw:
> Hi @kim-mskw,
> 
> currently the accepted_price needs to be set in the clearing mechanism. @maurerle and I decided that it may be useful to allow to not set the accepted_price in the clearing mechanism, as it could be misleading (order is not accepted but has an accepted price). The proposed change in the pull request would allow new clearing mechanism to omit the "accepted_price" key in the returned orderbook

I am hesitant with this change, as some of the capabilities in Assume, namely the learning, depend on the actual accepted price being set in the clearing , always, despite the non-acceptance of bids. So this would lead to some clearing algorithms not working with some bidding strategies; of course, they do not work with each other all the time. Still, the accepted volume is 0 if nothing is accepted, why the price then misleading? Rather, change the naming instead of making extra cases for different clearing algorithms?

PR #644, opened 2025-10-01
setup improvements

jrasko:
1. Specify the default database in the datasource.yml to prevent a manual setup
2. Use the new `docker compose` semantics instead of `docker-compose`

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ cd assume
 2. Start the database and Grafana using the following command:
 
 ```bash
-docker-compose up -d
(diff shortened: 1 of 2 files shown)


maurerle (approved):
Thanks!
This issue works around the problem of the dashboard showing `grafana postgres requires a default database to which to connect` when set.

We found the correct param by comparing `http://localhost:3000/api/datasources` before and after "save and test" of the datasource.

Great that works now.
And welcome to the Team @jrasko

PR #648, opened 2025-10-07
Learning refactor

### LINKED ISSUE #645: Current RL architecture incompatible with learning unit operators
The current learning architecture is incompatible with learning agents that are different from units (e.g. operators/market agents). See an overview of issues attached:
@maurerle @kim-mskw 

[TODO.txt](https://github.com/user-attachments/files/22659477/TODO.txt)

kim-mskw:
# Pull Request

## Related Issue
Closes #645 

## Description
- Units are not Mango roles (cannot send messages); the learning role has no direct access to units in a distributed setup.
- A special `_learning_unit_operator` aggregates per-update data from its units and sends a single message to the learning role.
- This pushes **learning data transport** outside the actual learning logic/strategy.
- **Single-sender coupling**: hard dependency on `_learning_unit_operator` (fragile, tight coupling).
- **No market-agent support**: learning-capable market agents can’t provide their own learning data without custom extensions to learning-market-agent.
- **No multi-sender aggregation**: the learning role can’t merge inputs from multiple unit operators (not implemented).

## Changes Proposed
- **Idea:** All learning entities/strategies (whether for market agents, unit operators, or units) should have direct access to the learning role and write data there themselves.
    - This would reduce the number of learning-related functions outside the learning context and centralize learning data handling in the learning role.
    - It would also **eliminate the need for double bookkeeping**, where all data is first written to `unit.outputs` and then collected by the operator to send to the learning role.
    - However, this would make distributed learning simulations (where different learning units run on different machines) impossible, but this feature is not currently used.

## Testing
Made several learning related tests to verify that buffer storage or updates are not fucked up. 
1. Test example_02a 
Tested this with the default config values and verified in TensorBoard that learning happens as intended. So for a single agent everything works as wanted. Great. 
    
image

2. Test example_02b 
Tested the learning behavior here as well with regard to general reward and profit direction, this looks good. Note: here we have medium competition between the agents, hence the profit will be reduced as soon as units are competing to be in the market. The resulting prices are however to low. 
image
This is a problem known to us because of the abs(profit) scaling introduced by @nick-harder. I never liked it since it is a clear reward tuning, priming learning results. So I got rid of it and commented why. 
These are the new example_02b results:
image

## Checklist

### DIFF
--- a/docs/source/learning.rst
+++ b/docs/source/learning.rst
@@ -186,3 +186,24 @@ A more detailed description is given in the dashboard itself.
 .. image:: img/Grafana_Learning_2.jpeg
     :align: center
     :width: 500px
+
+In accordance with common machine learning practices and to align with widely accepted learning standards, we also integrate **TensorBoard** to visualize
+the training process. This provides a familiar and intuitive interface for users who have experience with deep learning frameworks, where TensorBoard is
+a de-facto standard for inspecting metrics such as losses, gradients, hyperparameters, and model behavior over time.
+
+TensorBoard logs are automatically written to the directory specified in the project (here ``./tensorboard/``).
+
+To launch TensorBoard locally, run:
+
+.. code-block:: bash
+
+    tensorboard --logdir ./tensorboard --port 6006
+
+After starting the server, open the following URL in your browser:
+
+::
+
+    http://localhost:6006
+
+TensorBoard will then display dashboards for scalars, histograms, graphs, projectors, and other relevant visualizations, depending on the metrics that
+the training pipeline currently exports.
--- a/docs/source/learning_algorithm.rst
+++ b/docs/source/learning_algorithm.rst
@@ -8,7 +8,7 @@ Reinforcement Learning Algorithms
 
 In the chapter :doc:`learning` we got a general overview of how RL is implemented for a multi-agent setting in Assume.
 If you want to apply these RL algorithms to a new problem, you do not necessarily need to understand how the RL algorithms work in detail.
-All that is needed is to adapt the bidding strategies, which is covered in the tutorial.
+All that is needed is to adapt the bidding strategies, which is covered in the tutorials.
 However, for the interested reader, we will give a brief overview of the RL algorithms used in Assume.
 We start with the learning role, which is the core of the learning implementation.
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -21,12 +21,17 @@ Upcoming Release
 - **Fix tests on Windows**: One test was always failing on Windows, which is fixed so that all tests succeed on all archs
 
 **Improvements:**
+- **Application of new naming convention for bidding strategies**: [unit]_[market]_[method]_[comment] for bidding strategy keys (in snake_case) and [Unit][Market][Method][Comment]Strategy for bidding strategy classes (in PascalCase for classes)
+- **Restructured learning_role tasks**: Major learning changes that make learning application more generalizable across the framework.
 
-- **Application of new naming convention for bidding strategies: [unit]_[market]_[method]_[comment] for bidding strategy keys (in snake_case) and [Unit][Market][Method][Comment]Strategy for bidding strategy classes (in PascalCase for classes)
+  - **Simplified learning data flow:** Removed the special ``learning_unit_operator`` that previously aggregated unit data and forwarded it to the learning role. Eliminates the single-sender dependency and avoids double bookkeeping across units and operators.
+  - **Direct write access:** All learning-capable entities (units, unit operators, market agents) now write learning data directly to the learning role.
+  - **Centralized logic:** Learning-related functionality is now almost always contained within the learning role, improving maintainability.
+  - **Note:** Distributed learning across multiple machines is no longer supported, but this feature was not in active use.
 
-  **New Features:**
+
+**New Features:**
 - **Unit Operator Portfolio Strategy**: A new bidding strategy type that enables portfolio optimization, where the default is called `UnitsOperatorEnergyNaiveDirectStrategy`. This strategy simply passes through bidding decisions of individual units within a portfolio, which was the default behavior beforehand as well. Further we added 'UnitsOperatorEnergyHeuristicCournotStrategy' which allows to model bidding behavior of a portfolio of units in a day-ahead market. The strategy calculates the optimal bid price and quantity for each unit in the portfolio, taking into account markup and the production costs of the units. This enables users to simulate and analyze the impact of strategic portfolio bidding on market outcomes and unit profitability.
-
 
 0.5.5 - (13th August 2025)
 ==========================


kim-mskw:
# Status Update

## What Changed

### Improved Buffer Management in `store_to_buffer_and_update`
- Atomic snapshot creation for experience data  
- Exclusion of last timestep to handle reward delay  
- Prevention of race conditions in asynchronous processing  
- Collective transformation from GPU to CPU to avoid GPU memory pressure  
- Storage in nested dictionaries with `unit_id` and `starttime_id` to enforce consistent ordering of experiences in the buffer  

### Separation of Synchronous and Asynchronous Code
- Buffer and policy updates now run in the synchronous context  
- Output writing operations handled in the asynchronous context  
- **Un-Fixed recurring "no running event loop" errors**  

---

## What Stayed the Same
- Buffer structure  
- Policy update logic  
- Output formats

assume/common/outputs.py
```diff
@@ -266,6 +266,14 @@ def convert_rl_params(self, rl_params: list[dict]):
         df["evaluation_mode"] = self.evaluation_mode
         df["episode"] = self.episode if not self.evaluation_mode else self.eval_episode
 
+        # check for tensors and convert them to floats
+        # TODO: apply on dataframe with datetime index (column is pd.Series) doesn't work
+        # df = df.apply(convert_tensors)
+
+        # workaround: apply on each column
+        for col in df.columns:
+            df[col] = df[col].apply(convert_tensors)
```
jrasko, line 272: consider ditching the loop and applying to each column directly like this:
```suggestion
        df.apply(convert_tensors, axis=1)
```
This may increase the performance

assume/common/utils.py
```diff
@@ -607,28 +611,106 @@ def convert_tensors(data):
...
+                # Navigate to the parent of the target location
+                for p in path[:-1]:
+                    target = target[p]
+
+                # Set the final value
+                final_key = path[-1]
+                if isinstance(target, (list | pd.Series)):
+                    target[final_key] = val
+                elif isinstance(target, dict):
+                    target[final_key] = val
```
jrasko: These cases seem equivalent and may be merged:
```suggestion
                if isinstance(target, (list | pd.Series | dict)):
                    target[final_key] = val
```

assume/reinforcement_learning/learning_utils.py
```diff
@@ -188,6 +188,55 @@ def copy_layer_data(dst, src):
             dst[k].data.copy_(src[k].data)
 
 
+def transform_buffer_data(nested_dict: dict, device: th.device) -> list:
```
jrasko: This method always returns a numpy array:
```suggestion
def transform_buffer_data(nested_dict: dict, device: th.device) -> np.ndarray:
```

assume/reinforcement_learning/learning_utils.py
```diff
@@ -188,6 +188,55 @@ def copy_layer_data(dst, src):
...
+            "Error, while transforming RL data for buffer: No data found to determine feature dimension"
+        )
+
+    # Pre-allocate tensor (keep on same device as input data)
+    result = th.zeros((len(all_times), len(unit_ids), feature_dim), device=device)
+
+    # Fill tensor with values (stays on same device as input so if on GPU it stays there during filling)
+    for t, timestamp in enumerate(all_times):
+        for u, unit_id in enumerate(unit_ids):
+            values = nested_dict[timestamp].get(unit_id, [])
```
jrasko, line 231: The documentation (Line 193) implies nested_dict to be structured like this:
```
nested_dict: Dict with structure {unit_id -> {datetime -> list[tensor]}}
```
But here we access nested_dict like this: timestamp -> unti_id -> list[tensor]
So either the documentation or this code are incorrect

assume/common/outputs.py
kim-mskw: Good point unfortunately this does not work because then the output is differently. I tested and if I rember correctly it is then a series for which the convert tensor function does not mange to store the position correctly.

kim-mskw: That is also what the TODO says. although this fixes it.

assume/common/utils.py
kim-mskw: Great thanks!

assume/reinforcement_learning/learning_utils.py
kim-mskw: Good catch! I changed that in the code to avoid some loops further down the road and forgot to change it in the doc string. Thanks :)

kim-mskw:
@maurerle, could you jump in for review? I want to finish this soon since @chiara-fb is waiting for it, but I haven't heard back from @nick-harder, who wanted to do the review last Friday.

assume/reinforcement_learning/learning_role.py
```diff
@@ -139,9 +148,243 @@ def __init__(
...
+        """
+        Compute and validate validation_interval.
+
+        Returns:
+            validation_interval (int)
+        Raises:
+            ValueError if training_episodes is too small.
+        """
+        default_interval = int(learning_config.get("validation_episodes_interval", 5))
+        training_episodes = int(getattr(self, "training_episodes", 0))
```
maurerle: @kim-mskw this is called after __init__ so self.training_episodes exists and can be called.

assume/reinforcement_learning/learning_role.py
```diff
@@ -139,9 +148,243 @@ def __init__(
...
+        validation_interval = min(training_episodes, default_interval)
+
+        min_required_episodes = (
+            int(getattr(self, "episodes_collecting_initial_experience", 0))
+            + validation_interval
+        )
+
+        if training_episodes < min_required_episodes:
+            raise ValueError(
+                f"Training episodes ({training_episodes}) must be greater than the sum of initial experience episodes ({getattr(self, 'episodes_collecting_initial_experience', 0)}) and evaluation interval ({validation_interval})."
```
maurerle: same here

assume/strategies/learning_strategies.py
```diff
@@ -709,8 +726,8 @@ class EnergyLearningSingleBidStrategy(EnergyLearningStrategy, MinMaxStrategy):
     """
 
     def __init__(self, *args, **kwargs):
-        obs_dim = kwargs.pop("obs_dim", 74)
-        act_dim = kwargs.pop("act_dim", 1)
+        obs_dim = self.learning_role.obs_dim
+        act_dim = self.learning_role.act_dim
```
maurerle: This won't work if not in learning_mode

assume/common/units_operator.py
```diff
@@ -465,6 +461,9 @@ async def submit_bids(self, opening: OpeningMessage, meta: MetaDict) -> None:
             market_config=market,
             product_tuples=products,
         )
+
+        orderbook = convert_tensors(orderbook)
```
maurerle: This should only be done for learning strategies somehow - otherwise we have different behavior depending on having torch installed, which is weird.

This fails the integration test, as the dataframes here are different than expected.

We would be back to the situation that we would need `UnitsOperatorDirectStrategy` normally, but as soon as one learning strategy is added, we need `UnitsOperatorDirectStrategyWithConvert` - which is not ideal either.

Best would be to adjust the convert_tensors function to pass the test - aka not change the format if no tensors are there.

assume/strategies/learning_strategies.py
```diff
@@ -60,6 +61,10 @@ def __init__(self, *args, **kwargs):
         # sets the device of the actor network
         device = kwargs.get("device", "cpu")
         self.device = th.device(device if th.cuda.is_available() else "cpu")
+        if self.learning_mode and not self.learning_role:
+            raise AssumeException("Learning Role must be set in LearningMode")
```
maurerle, line 65: The check to fail early here was added by myself and is required, as the register method is called afterwards anyways on the self.learning_role.
So we need to adjust the tests here to add a mock learning_role when creating learningStrategies..?

This reduces testability slightly, as we have the dependency on the learning role - but I think it is better to fail early.
Alternative would be to disable registration if learning_role is not set, which could be silent unwanted behavior as well.

maurerle (changes requested):
Looks good - Some adjustments to the tests and convert_tensors is still needed to pass the tests

assume/strategies/learning_strategies.py
maurerle: I am missing some work of @mthede here? Is this postponed to a later PR?
This situation was discussed to be solved by passing the learning_config..? :)

assume/common/utils.py
```diff
@@ -712,3 +792,47 @@ def get_supported_solver(default_solver: str | None = None):
...
+    if continue_learning:
+        logger.warning(
+            f"Save path '{save_path}' exists.\n"
+            "You are in continue learning mode. New strategies may overwrite previous ones.\n"
+            "It is recommended to use a different save path to avoid unintended overwrites.\n"
+            "You can set 'trained_policies_save_path' in the config."
+        )
+        proceed = input(
+            "Do you still want to proceed with the existing save path? (y/N) "
+        )
```
jrasko: inputs are always a point of failure for tests, so we should add the possibility to skip the input using an environment variable.
If this hasn't been an issue yet, we may need to add tests for this 😄

assume/reinforcement_learning/learning_role.py
```diff
@@ -139,9 +148,242 @@ def __init__(
...
+                return None
+
+            train_freq_str = str(self.train_freq)
+            try:
+                train_freq = pd.Timedelta(train_freq_str)
+            except Exception:
+                logger.warning(
+                    f"Invalid train_freq '{train_freq_str}' — skipping adjustment."
+                )
+                return None
```
jrasko: this part may be moved out of the top level try-catch to reduce complexity
```suggestion
        if not self.learning_mode:
            return None

        train_freq_str = str(self.train_freq)
        try:
            train_freq = pd.Timedelta(train_freq_str)
        except Exception:
            logger.warning(
                f"Invalid train_freq '{train_freq_str}' — skipping adjustment."
            )
            return None
        try:
```

assume/common/utils.py
maurerle: ~~This is already surrounded by `OVERWRITE_LEARNED_STRATEGIES`..~~
But I would rather have `input` surrounded with `if not os.getenv("NON_INTERACTIVE")`..?
Eventually with a wrapper function, so that we do not have to define it multiple times..?

assume/reinforcement_learning/learning_role.py
jrasko: if there is a specific place where the top level try-catch is expecting an exception, please wrap only these lines to facilitate later debugging

assume/common/utils.py
jrasko: @maurerle When using a _NON_INTERACTIVE_ flag everywhere, we have a strict default behavior, but we might want to test how the program reacts to different inputs, using different behaviors. So I think a general _NON_INTERACTIVE_ flag might be not suited as a long-term solution

assume/reinforcement_learning/learning_role.py
maurerle: i removed the top-level try catch.
I guess this should not fail silently without returning anything.

assume/common/utils.py
maurerle: Also True - but lets handle this in a separate issue afterwards, I guess?

assume/common/outputs.py
```diff
@@ -266,6 +266,13 @@ def convert_rl_params(self, rl_params: list[dict]):
         df["evaluation_mode"] = self.evaluation_mode
         df["episode"] = self.episode if not self.evaluation_mode else self.eval_episode
 
+        # check for tensors and convert them to floats
+        # apply on dataframe with datetime index (column is pd.Series) doesn't work
+        # df = df.apply(convert_tensors)
```
nick-harder: Do we need this comments and not working code here?

assume/common/units_operator.py
nick-harder: maybe this could be part of the base market, so this check is done always ion all of the orderbooks if we have installed torch?

assume/reinforcement_learning/learning_role.py
```diff
@@ -139,9 +146,238 @@ def __init__(
...
+            cache = {
+                "obs": {t: current_obs[t] for t in timestamps_to_process},
+                "actions": {t: current_actions[t] for t in timestamps_to_process},
+                "rewards": {t: current_rewards[t] for t in timestamps_to_process},
+                "noises": {t: current_noises[t] for t in timestamps_to_process},
+                "regret": {t: current_regrets[t] for t in timestamps_to_process},
+                "profit": {t: current_profits[t] for t in timestamps_to_process},
+            }
+
+            # write data to output agent before resetting it
```
nick-harder: this comment somehow contradicts the comment above where you reset cache directly

assume/reinforcement_learning/learning_role.py
```diff
@@ -139,9 +146,238 @@ def __init__(
...
+            self.write_rl_params_to_output(cache)
```
nick-harder, line 298: maybe the whole thing with cache preparation could be in a different method? all the collection, timestamp and reset logic? to keep the store and update clean?

assume/reinforcement_learning/learning_role.py
```diff
@@ -139,9 +146,238 @@ def __init__(
...
+                return
+
+        # rewrite dict so that obs.shape == (n_rl_units, obs_dim) and sorted by keys and store in buffer
+        self.buffer.add(
+            obs=transform_buffer_data(cache["obs"], device),
+            actions=transform_buffer_data(cache["actions"], device),
+            reward=transform_buffer_data(cache["rewards"], device),
+        )
+
+        self.update_policy()
```
nick-harder: maybe we can move the contents of this update policy inside this method since it is just two lines of code? somehow now we have a sequence of three methods for the storing and update.

assume/scenario/loader_csv.py
```diff
@@ -915,20 +891,14 @@ def setup_world(
     else:
         logger.info("Adding unit operators and units")
         for company_name in set(units.keys()):
-            if company_name == "Operator-RL" and world.learning_mode:
```
nick-harder, line 918: In line 432 we are still replacing all the unit operators with Operator RL. Do we want to keep this behavior? I remember you wanted to switch to allow multiple unit operators and also let the user decide if they want to keep one or they need multiple.

assume/strategies/learning_strategies.py
```diff
@@ -366,13 +378,7 @@ class EnergyLearningStrategy(TorchLearningStrategy, MinMaxStrategy):
     """
 
     def __init__(self, *args, **kwargs):
-        obs_dim = kwargs.pop("obs_dim", 38)
-        act_dim = kwargs.pop("act_dim", 2)
-        unique_obs_dim = kwargs.pop("unique_obs_dim", 2)
```
nick-harder, line 371: why was this dropped? How is it now defined and where? I don't understand this

assume/scenario/loader_csv.py
maurerle: I think we can remove the whole `adjust_unit_operator_for_learning` function now?

assume/strategies/learning_strategies.py
maurerle: Yes, this was removed by mistake. Readded

maurerle:
There are 7 tests failing in a weird way:

```
FAILED tests/test_drl_storage_strategy.py::test_storage_rl_strategy_sell_bid - AssertionError: Expected reward 0.09000000000000001, got 0.0
FAILED tests/test_drl_storage_strategy.py::test_storage_rl_strategy_buy_bid - AssertionError: Expected reward -0.45, got 0.0
FAILED tests/test_integration_cli.py::test_cli - AssertionError: DataFrame are different
FAILED tests/test_rl_buffer.py::test_replay_buffer_add - ValueError: cannot select an axis to squeeze out which has size not equal to one
FAILED tests/test_rl_strategies.py::test_learning_strategies_parametrized[EnergyLearningStrategy-38-2-mlp-2-expected_volumes0] - assert np.float64(0.0) == 0.01
FAILED tests/test_rl_strategies.py::test_learning_strategies_parametrized[EnergyLearningStrategy-38-2-lstm-2-expected_volumes1] - assert np.float64(0.0) == 0.01
FAILED tests/test_rl_strategies.py::test_learning_strategies_parametrized[EnergyLearningSingleBidStrategy-74-1-mlp-1-expected_volumes2] - assert np.float64(0.0) == 0.01
```

assume/scenario/loader_csv.py
nick-harder: Maybe for performance purposes we can leave and keep it as optional behvaior when no portfolio optimization is used. this has impact on performance, so it would be nice to have it. especially for large simulations. But maybe warn the user and add an arg to the config?

assume/common/outputs.py
kim-mskw: made the explanation standalone and deleted old code

assume/common/units_operator.py
kim-mskw: fair, done

assume/reinforcement_learning/learning_role.py
kim-mskw: I need to make the timestamp check here and depneding on this the store to buffer and update. I do not see how this is can be solved nicely, so I keep it as it is for now

assume/strategies/learning_strategies.py
kim-mskw: okay, it is adjusted in the test. I agrre it should fail early

assume/scenario/loader_csv.py
kim-mskw: I think now that the convert tensor function is in the base_market it should not have performance issues and we can just get rid of it.

.github/pull_request_template.md
maurerle, line 1: This change should be fixed.

PR #650, opened 2025-10-17
use `np.ndarray` instead of `np.array` as type hint

jrasko:
Also add some missing type hints and correct wrong ones


maurerle (approved):
looks good. `np.array` is a function (to create a ndarray) and not a type, therefore `np.ndarray` should be used everywhere

PR #653, opened 2025-10-20
force negative demand

jrasko:
Force min_power < 0 and max_power < 0 when creating a new unit
Adjust the csv-importer adapter to transform positive power demands into negative ones
Update the example jupyter notebooks

### DIFF
--- a/assume/units/demand.py
+++ b/assume/units/demand.py
@@ -21,8 +21,8 @@ class Demand(SupportsMinMax):
         technology (str): The technology of the unit.
         bidding_strategies (dict): The bidding strategies of the unit.
         index (pandas.DatetimeIndex): The index of the unit.
-        max_power (float): The maximum power output capacity of the power plant in MW.
-        min_power (float, optional): The minimum power output capacity of the power plant in MW. Defaults to 0.0 MW.
+        max_power (float): The maximum power input (negative) capacity of the power plant in MW.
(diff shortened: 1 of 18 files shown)


assume/units/demand.py
```diff
@@ -54,16 +54,20 @@ def __init__(
...
 
+        if max_power > 0:
+            raise ValueError(f"max_power must be < 0 for unit {self.id}")
+        if min_power > 0:
+            raise ValueError(
+                f"min_power must be < 0 but is {min_power} for unit {self.id}"
+            )
+        if max_power > min_power:
+            raise ValueError(
+                f"max_power (={max_power} must be <= min_power (={min_power}) for unit {self.id}"
```
maurerle:
```suggestion
                f"{=max_power} must be <= {=min_power} for unit {self.id}"
```

This would print "maxpower=60 must be <= min_power=40"

examples/notebooks/01_minimal_manual_example.ipynb
maurerle, line 1: Please clear the notebooks before committing

maurerle (approved):
max_power is now consistently negative for demand - this is a good fix

ramping for demand is turned of instead of having weird changes (as min and max would have to be reversed)

This also fixes doc strings for demand.

Looks good to me @kim-mskw can you give it a look as well?

PR #654, opened 2025-10-21
add validations to powerplant and storage constructors

jrasko:
Check attributes when initialising powerplants and storages to prevent side effects that may lead to an unexpected behavior.

### DIFF
--- a/assume/units/storage.py
+++ b/assume/units/storage.py
@@ -64,7 +64,7 @@ class Storage(SupportsMinMaxCharge):
         min_power_charge: float = 0.0,
         min_power_discharge: float = 0.0,
         min_soc: float = 0.0,
-        initial_soc: float = 0.0,
+        initial_soc: float | None = 0.0,
         soc_tick: float = 0.01,
         efficiency_charge: float = 1,
         efficiency_discharge: float = 1,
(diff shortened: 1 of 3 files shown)


assume/units/powerplant.py
```diff
@@ -303,6 +312,7 @@ def calculate_marginal_cost(self, start: datetime, power: float) -> float:
         """
         # if marginal costs already exists, return it
         if self.marginal_cost is not None:
+            # FIXME self.marginal_cost is a float and cannot be indexed
```
jrasko: this seems like a codesmell, I can fix this if I know how it should be (list or float)

assume/units/storage.py
```diff
@@ -125,34 +135,49 @@ def __init__(
...
+
+        if ramp_up_discharge is None:
+            self.ramp_up_discharge = self.max_power_discharge - self.max_power_charge
+        elif ramp_up_discharge < 0:
+            raise ValueError(f"{ramp_up_discharge=} must be >= 0 for unit {self.id}")
+        else:
+            self.ramp_up_discharge = ramp_up_discharge
+
+        if ramp_down_discharge is None:
+            self.ramp_down_discharge = self.max_power_discharge - self.max_power_charge
```
maurerle: I think we should remove the ramp_down/ramp_up constraints if they are set as none - that way we do not introduce potentially error-prone calculations

PR #655, opened 2025-10-22
add MinMaxStrategy and MinMaxChargeStrategy to ensure compatibility between unit and strategy

jrasko:
Distinguish Strategies based on the Units that can use them.
Strategies that can be used on MinMaxUnits must now implement the MinMaxStrategy class while Strategies that are to be used by MinMaxChargeUnits needs to implement the MinMaxChargeStrategy class


maurerle (approved):
This wil break for additional strategies which are userdefined.
They need to adjust their strategies

PR #656, opened 2025-10-22
improve error message when learning without given database

maurerle:
Learning without a connected database fails with `No rewards were collected during evaluation run` when no database is configured.

Example: `assume -s example_02c`

This should provide a better warning.

A potential fix is to simply add the database `assume -s example_02c -db "postgresql://assume:assume@localhost:5432/assume"`

With this PR a more helpful hint is given in these cases


jrasko (approved):
TL;DR

Issue #635, opened 2025-09-10
Demand unit id with NaiveForcast is forced to comprise the substring 'demand'.


Graf-Wronski:
When creating a demand unit with an id like "household" and a ```NaiveForecast```, this unit will not participate in the market.

Reason to this is the code line 67 in ```assume/units/demand.py```:
    ```self.volume = -abs(self.forecaster[self.id])```

and the corresponding ```__getitem__``` function of ```NaiveForecast```

Only if "demand" appears in the id of the demand unit, the volume will be correctly set. IMO code ```line:67``` in ```assume/units/demand.py``` should read:
    ```self.volume = -abs(self.forecaster.demand)```

To reproduce:

```
from pathlib import Path

import pandas as pd

from datetime import datetime

from assume import World
from assume.common.forecasts import NaiveForecast


start = datetime(2023, 1, 13, hour=0)
end = datetime(2023, 1, 13, hour=4)
temporal_resolution = "60min"
index = pd.date_range(start, end, freq=temporal_resolution)
n_steps = len(index)

# Create ASSUME World.
db_uri = "sqlite:///local_db/assume_db.db"
world = World(database_uri=db_uri)
world.setup(
    start=start,
    end=end,
    save_frequency_hours=1,
    simulation_id="reproduce_issue",
    index=index)

world.add_unit_operator(f"demand_operator")

print(f"{"unit_id":<25}{"volume.sum()":>10}")

for unit_id in ["demand_unit_0", "household_0", "schokolademandelbrot_0", "suisse_availability"]:

    demand_forecast = NaiveForecast(index, demand=100)

    world.add_unit(
        id=unit_id,
        unit_type="demand",
        unit_operator_id=f"demand_operator",
        unit_params={
            "min_power": 0,
            "max_power": 1000,
            "bidding_strategies": {"EOM": "naive_eom"},
            "technology": "demand"},
    forecaster=demand_forecast)

    volume = world.unit_operators["demand_operator"].units[unit_id].volume
    print(f"{unit_id:<25}{volume.sum():>10}")

world.run()
``` 

<img width="477" height="168" alt="Image" src="https://github.com/user-attachments/assets/fd7afbc9-f6f8-47cd-a244-61f6f40ec41c" />

PR #660, opened 2025-10-29
Have a single forecaster per unit instead of one global forecaster

jrasko:
The global forecaster provides a way to store the input data of the units in a centralized dict.
However, this approach complicates the identification of misconfiguration and may result in simulations that run with zero values producing wrong outputs, making it difficult for users to identify the problem.

Assigning each units its own forecaster solves these issues by requiring necessary fields to be set and simplifying validations of fields on unit-level. This also solves issue #635, where the resolved forecaster field depends on the id of a unit.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -81,6 +81,7 @@ User Guide
 * :doc:`outputs`
 * :doc:`unit_operator`
 * :doc:`units`
+* :doc:`unit_forecasts`
 * :doc:`bidding_strategies`
 * :doc:`learning`
 * :doc:`learning_algorithm`
@@ -102,6 +103,7 @@ User Guide
    outputs
    unit_operator
    units
+   unit_forecasts
    bidding_strategies
    learning
    learning_algorithm
--- a/docs/source/unit_forecasts.rst
+++ b/docs/source/unit_forecasts.rst
@@ -1,0 +1,32 @@
+.. SPDX-FileCopyrightText: ASSUME Developers
+..
+.. SPDX-License-Identifier: AGPL-3.0-or-later
+
+###########
+Unit Forecasts
+###########
+
+To provide units with static input data such as availability, fuel prices or market price prediction data, each unit has an associated unit forecast.
+
+The data of forecasts is typically provided as time series data and either loaded from a csv file or calculated internally from imported data.
+
+Each unit type has typically its own forecast type:
+
+=============== ====================
+ Unit Type      Forecast Type
+=============== ====================
+ Power Plants   PowerplantForecaster
+ Storage Units  UnitForecaster
+ Demand Units   DemandForecaster
+=============== ====================
+
+Each forecast type contains specific fields relevant to the respective unit type. For example, a PowerplantForecaster may include fields such as fuel prices, while a DemandForecaster may include the expected demand profile.
+
+To update calculated forecast fields at runtime, create and run a new `ForecastInitialisation` with the updated data and then use the results to update the respective unit forecasts.
+
+For custom units, there are two options to handle forecasts:
+
+1. Add a new forecaster for the custom unit type by creating a new class that inherits from `UnitForecaster` and implements the necessary fields and methods.
+2. Use the `CustomUnitForecaster`, which allows for flexible definition of forecast fields without creating a new class.
+
+Option 2 is used when importing custom units via csv files.


assume/units/steel_plant.py
```diff
@@ -79,6 +79,10 @@ def __init__(
             location=location,
             **kwargs,
         )
+        if not isinstance(forecaster, SteelplantForecaster):
+            raise TypeError(
+                f"forecaster must be of type {SteelplantForecaster.__name__}"
+            )
```
maurerle, line 85: don't know if this is too much magic, but you could also have a ClassVariable `forecaster_type = Forecaster` in the base unit and add the check `if not isinstance(forecaster, cls.forecaster_class):` in the base class as a `@classmethod`.

That way, you only need to set `forecaster_type = SteelplantForecaster` in the class and do not need to repeat the check itself

assume/common/forecaster.py
```diff
@@ -0,0 +1,207 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import pandas as pd
+
+from assume.common.fast_pandas import FastIndex, FastSeries
+
+type ForecastIndex = FastIndex | pd.DatetimeIndex | pd.Series
+type ForecastSeries = FastSeries | list | float | pd.Series
```
maurerle: I would still use the TypeVar command from typing here..

We still have downloads from 3.10 and 3.11 and I would like to have the next release to still target these as well: https://pypistats.org/packages/assume-framework

According to https://endoflife.date/python it is still supported.

Our main functionality also works with python 3.14, but pytorch tends to be behind with this quite often: https://github.com/pytorch/pytorch/issues/156856

assume/units/steel_plant.py
jrasko: I think both implementations are a valid way, but I think I like the explicit checking better, since every dev can than understand the mechanic on first sight

assume/common/forecaster.py
jrasko: swapped to use typalias

kim-mskw:
I can review this tomorrow if needed!

jrasko:
@kim-mskw i'm eager to hear your opinion on this restrucuring :smile:

assume/scenario/loader_csv.py
```diff
@@ -549,6 +560,95 @@ def load_config_and_create_forecaster(
     buses = load_file(path=path, config=config, file_name="buses")
     lines = load_file(path=path, config=config, file_name="lines")
 
+    calculator = Calculations(
```
kim-mskw: This lookes kinda weird, that we intialize just a calculator. Does not seem intuiutive what it does. Could we rename it to Forecast_calculator? And similar the class not calculations?

assume/scenario/loader_csv.py
```diff
@@ -549,6 +560,95 @@ def load_config_and_create_forecaster(
...
+        market_configs=config["markets_config"],
+    )
+    calculator.set_forecast(forecasts_df)
+    calculator.set_forecast(demand_df)
+    calculator.set_forecast(exchanges_df)
+    calculator.set_forecast(availability, prefix="availability_")
+    calculator.calc_forecast_if_needed()
+
+    unit_forecasts: dict[str, UnitForecaster] = {}
+    market_prices, residual_loads = calculator.calculate_market_forecasts()
```
kim-mskw, line 580: Ah great so no redundnacies with regard to the market_wide_forecasts. I would add the wide for clarity in the naming if you agree. 

So if I want to update the forecasts later on during the simulation, I will need to do this again correct? Could we refractor it into another function so that it might be callable during some time in the world.run() and not being bind to the load_config_and_create_forecaster function?

assume/common/calculations.py
```diff
@@ -4,95 +4,12 @@
...
-            >>> forecaster = Forecaster(index=FastIndex(start='2020-01-01', end='2025-01-01'))
-            >>> price = forecaster.get_price('lignite')
-            >>> print(price)
-        """
-
-        return self[f"fuel_price_{fuel_type}"]
-
-
-class CsvForecaster(Forecaster):
+class Calculations:
```
kim-mskw: Why is this a calculations class? See other comment

assume/common/forecaster.py
```diff
@@ -0,0 +1,208 @@
...
+
+import pandas as pd
+
+from assume.common.fast_pandas import FastIndex, FastSeries
+
+ForecastIndex: TypeAlias = FastIndex | pd.DatetimeIndex | pd.Series
+ForecastSeries: TypeAlias = FastSeries | list | float | pd.Series
+
+
+class UnitForecaster:
```
kim-mskw, line 14: Do we expect a unit_operator that wants a forecast to have a class that differs substantially from this base class? If not I would suggest renaming it to something different than UnitForecsater. Or does a unit_opertor get the forecaster for all her units assigned?

assume/scenario/loader_amiris.py
```diff
@@ -168,15 +172,15 @@ def add_agent_to_world(
                     agent["Id"],
                     {
                         "min_power": 0,
-                        "max_power": 100000,
+                        "max_power": -100000,
```
kim-mskw, line 175: While on it, what is htis TODO @maurerle

assume/units/building.py
```diff
@@ -106,13 +111,14 @@ def __init__(
...
-        # Initialize forecasting data for various energy prices and demands
-        self.electricity_price = self.forecaster["price_EOM"]
-        self.natural_gas_price = self.forecaster["fuel_price_natural gas"]
-        self.heat_demand = self.forecaster[f"{self.id}_heat_demand"]
-        self.ev_load_profile = self.forecaster["ev_load_profile"]
-        self.battery_load_profile = self.forecaster["battery_load_profile"]
-        self.inflex_demand = self.forecaster[f"{self.id}_load_profile"]
+        self.market_id = "EOM"  # FIXME enable other markets
+
+        self.electricity_price = self.forecaster.price[self.market_id]
```
kim-mskw, line 116: So if I want to add a new unit type I will need to add a respecitve forecaster to get the data I need. Although the data is already there/made in the calculator. Correct?

assume/world.py
```diff
@@ -199,11 +199,10 @@ def setup(
             start (datetime.datetime): The start datetime for the simulation.
             end (datetime.datetime): The end datetime for the simulation.
             simulation_id (str): The unique identifier for the simulation.
-            index (pandas.Series): The index for the simulation.
+            lr_logfreq (str): Log frequency of the learning role
```
kim-mskw: This is not related to logging only. It is the simulation freuency. Why the renaming?

jrasko: actually, the world had no index, the docstring was just not up to date. Instead, it took the index from the forecaster and used it here:
```python
            self.learning_role.init_logging(
                simulation_id=self.simulation_id,
                episode=episode,
                eval_episode=eval_episode,
                db_uri=self.db_uri,
                output_agent_addr=self.output_agent_addr,
                train_start=self.start,
                freq=self.lr_logfreq, # <-------------------
            )
```
if this corresponds to the frequency of the simulation, a simple renaming of the field to `simulation_frequency` should do the trick

assume/units/building.py
jrasko: Well, there are two possible ways to implement custom type forecasters:

1. Implementing a forecaster that derives from `UnitForecaster` and passing it directly
2. Using the `CustomUnitForecaster`, which most closely corresponds to the old forecaster by allowing any fields to be set

Option 2 is also used for custom unit CSV import. If needed, we may pass calculated fields to the `CustomUnitForecaster` to avoid the need for recalculation

assume/scenario/loader_csv.py
jrasko: Hey, I think this would be a new feature which is out of scope for now.

As a remark, I think we should keep forecasters as a sink for static data, dynamic cases like this should be handled by another object (my opinion, we should discuss this in the dev group)

assume/scenario/loader_csv.py
jrasko: My first thought was the use of static methods to calculate all needed fields. Since this turned out more complciated than I first anticipated, I just used the legacy CSVForecaster mechanisms and outsourced them into a seperate object - as a technical dept that is not yet covered by this PR but may be adressed in the future.

If you find the name `ForecastCalculator` more fitting I can just rename it.

assume/common/calculations.py
jrasko: since the discussion is in the comment above, i resolve this conversation

assume/common/forecaster.py
jrasko: I'm a little confused here - as far as I understand, the unit gets a forecaster, not the unit operator - or do you mean something different?

assume/scenario/loader_amiris.py
maurerle: I did not validate the section to behave similar as in AMIRIS. So I would have to report back about this.
We generally have to retest that everything works with latest AMIRIS 4.0 as well - but this is out of scope for now :)

assume/scenario/loader_csv.py
kim-mskw: There are two ways of handling forecast updates during the simulation for now, with this architecture. The idea is that with the __get_item function, which allows to update the local forecasts of the unit right before getting the forecast. If global updates are required, we can also recall the calculator like in the beginning which does the global calculation and can then update these for all units. [TODO: Add this to documentation]

assume/common/forecaster.py
kim-mskw: The unit_forecaster can get the forecast from the first unit it has, since we do not see a reason right now to have unit_operator-specific forecasts, due to lack of forecast heterogeneity anyhow.

assume/units/building.py
kim-mskw: Okay great! Then add this just in the documentation.

assume/world.py
kim-mskw: delete freq all together not needed in learning_role

assume/scenario/loader_csv.py
kim-mskw: Yes or ForecastInitialisation or something similar

kim-mskw (approved):
As per discussion in the development sprint

PR #668, opened 2025-11-11
Renaming of bidding strategies to unit-market-method pattern

### LINKED ISSUE #549: Documentation of bidding strategies is missing

reinecfi:
## Related Issue
Closes #549 
Closes #559

## Description
Application of the naming convention for bidding strategies:
 - `[unit]_[market]_[method]_[comment]` for bidding strategy keys (in snake_case)
 - `[Unit][Market][Method][Comment]Strategy` for bidding strategy classes (in PascalCase for classes)
 
Units denote one of the following: powerplant (or renewable), storage, demand, household, industry, unit operator
Possible markets are capacity or energy.
Methods are one of the following: naive, heuristic, optimization, learning
Comments denote optional (descriptive) keywords like: flexable or balancing

## Checklist

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -19,6 +19,9 @@ Upcoming Release
 - **Update pytest dependency**: Tests now run with Pytest 9
 - **Add new docs feature**: dependencies to build docs can now be installed with `pip install -e .[docs]`
 
+**Improvements:**
+- **Application of new naming convention for bidding strategies: [unit]_[market]_[method]_[comment] for bidding strategy keys (in snake_case) and [Unit][Market][Method][Comment]Strategy for bidding strategy classes (in PascalCase for classes)
+
   **New Features:**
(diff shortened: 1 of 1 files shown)


maurerle:
@gugrimm I think it is B) - but as the strategy does not rely on optimization, it is still naive. Only the complex market clearing requires optimization.
Without the `bid_type "BB"`, this would be quite similar to the default naive strategy.
The `bid_type "BB"` is only used in the complex clearing as well.

maurerle:
@reinecfi could you please attach the mapping of old strategies to new strategies somehow here as a comment, please? :)

I would like to have this merged. Further restructuring should happen in smaller steps afterwards (for example moving the strategy implementations into matching files together)

PR #670, opened 2025-11-11
check simulation results in the integration test

jrasko:
## Description
Check results in the integration tests to detect calculation errors

## Checklist

### DIFF
--- a/tests/test_integration_cli.py
+++ b/tests/test_integration_cli.py
@@ -37,7 +37,8 @@ def test_cli():
 @pytest.mark.slow
 @pytest.mark.require_network
 def test_cli_network():
-    args = "-s example_01d -c base -db sqlite:///./examples/local_db/test_mini_net.db"
+    dburi = "sqlite:///./examples/local_db/test_mini_net.db"
+    args = f"-s example_01d -c base -db {dburi}"
     cli(args.split(" "))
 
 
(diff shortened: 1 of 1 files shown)


assume/scenario/loader_csv.py
```diff
@@ -1056,7 +1056,7 @@ def run_learning(
     # check if we already stored policies for this simulation
     save_path = world.learning_config["trained_policies_save_path"]
 
-    if Path(save_path).is_dir():
+    if Path(save_path).is_dir() and os.getenv("OVERWRITE_LEARNED_STRATEGIES") == "":
```
maurerle:
```suggestion
    if Path(save_path).is_dir() and not os.getenv("OVERWRITE_LEARNED_STRATEGIES"):
```

maurerle (approved):

tests/test_integration_cli.py
```diff
@@ -1,26 +1,49 @@
...
+    args = "-s example_01d -c base -db sqlite:///./examples/local_db/test_mini_net.db"
     cli(args.split(" "))
 
 
 @pytest.mark.slow
 @pytest.mark.require_learning
 def test_cli_learning():
+    os.putenv("OVERWRITE_LEARNED_STRATEGIES", "TRUE")
     args = "-s example_02a -c tiny -db sqlite:///./examples/local_db/test_mini_rl.db"
     cli(args.split(" "))
```
maurerle, line 51: would be good to also have a simple check for example_02a, to at least check one of the queries:

maybe only that there are 10 training episodes in the database:
```sql
SELECT 
    episode,
    unit, 
    SUM(reward) AS total_reward
FROM rl_params
WHERE simulation = '${simulation}'
    AND evaluation_mode = FALSE
GROUP BY episode, unit
```

PR #680, opened 2025-11-13
Fixup the forecast Refactoring

jrasko:
* Calculate fields *only if* they do not already exists
* Allow the saving of calculated fields
* Instantiate Buildings

The results of the new _forecast_init_ tests are equal to the results before the refactoring


maurerle (approved):
Looks Good - i would squash this PR though? :)

PR #681, opened 2025-11-14
improve forecast initialization methods

jrasko:
* Better handling of availabilities
* Reuse power-calculation logic
* (Hopefully) improved performance by having fewer loops

Further improvements and tests are on their way!

@maurerle do you have an opninion on further refactorin here?


assume/common/forecast_initialisation.py
```diff
@@ -265,22 +262,6 @@ def calculate_residual_load_forecast(self, market_id) -> pd.Series:
             5. If exchange units are available, imports and exports are subtracted and added to the demand forecast, respectively.
 
         """
-
-        vre_powerplants_units = self.powerplants_units[
-            self.powerplants_units["technology"].str.contains(
-                r"\b(?:wind|solar)\b", case=False, na=False
-            )
```
jrasko, line 272: it seems like this regex doesn't work correctly since it doesn't match powerplants named _Wind Onshore_, only _Solar_

maurerle: Thsi works fine for me though..?

```python
>>> df = pd.DataFrame([["Solar" ,2,3],["Wind Onshore",5,6]])
>>> df[0].str.contains(r"\b(?:wind|solar)\b", case=False, na=False)
0    True
1    True
Name: 0, dtype: bool
```

jrasko: Can't reproduce it anymore, maybe this was an issue from my testing and debugging hell

PR #682, opened 2025-11-18
add forecast initialization tests

jrasko:
## Description
Add tests to the forecast initialization to improve code-confidence and streamline later refactorings. Use realistic data to do so.

## Checklist

### DIFF
--- a/tests/test_forecast_initialization.py
+++ b/tests/test_forecast_initialization.py
@@ -3,13 +3,17 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
 
+from pathlib import Path
+
 import pandas as pd
 import pytest
 from pandas._testing import assert_frame_equal, assert_series_equal
 
 from assume.common.forecast_initialisation import ForecastInitialisation
 
-path = "./tests/fixtures/forecast_init"
+path = Path("./tests/fixtures/forecast_init")
+
+parse_date = {"index_col": "datetime", "parse_dates": ["datetime"]}
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/fixtures/forecast_init/results/load_forecast.csv
maurerle, line 1: This is missing a ffill to clear the NaN

jrasko: since the NaN values are part of the current behavior, we should keep them as part of the test

jrasko: I added checks to prevent different indices

tests/test_forecast_initialization.py
```diff
@@ -31,36 +29,81 @@ def forecast_init():
...
-        powerplants_units=powerplants_units,
-        demand_units=demand_units,
         market_configs=market_configs,
-        forecasts=pd.DataFrame(),
-        availability=pd.DataFrame(),
-        demand=demand,
-        exchanges=pd.DataFrame(),
-        fuel_prices=fuel_prices,
+        index=pd.date_range("2019-01-01", periods=24, freq="h"),
+        powerplants_units=pd.read_csv(f"{path}/powerplant_units.csv", index_col="name"),
```
maurerle: Use pathlib and `path / "powerplant_units.csv"` instead of an f string here

assume/common/forecast_initialisation.py
```diff
@@ -42,14 +52,14 @@ def __init__(
...
+        availability: pd.DataFrame = None,
+        exchanges: pd.DataFrame = None,
+        forecasts: pd.DataFrame = None,
         fuel_prices: pd.DataFrame = None,
-        exchange_units: pd.DataFrame | None = None,
-        buses: pd.DataFrame | None = None,
-        lines: pd.DataFrame | None = None,
+        exchange_units: pd.DataFrame = None,
+        buses: pd.DataFrame = None,
+        lines: pd.DataFrame = None,
```
maurerle, line 62: why did this change? I guess you still need `| None` in the type hints if this is initialized as None..?

Issue #664, opened 2025-11-10
Wishlist: Validation.


carl-wanninger:
| Task | Status | Location |
| :--- | :--- | :--- |
| Error: Datatypes for `MarketConfigurations` are not correct. | **specification needed: is this about clarity?** | **make_market_config(?)** |
| Warning: Input ranges do not make sense. | **specification needed: very general** | |
| Warning: PowerPlants: Availability > 1.| **drafted as Error** | **BaseUnit (since ValidationError requires id)** |
| Error: PowerPlants: Availability < 0. | **drafted** | **BaseUnit (since ValidationError requires id)** |
| Error: Market time exceeds simulation time. | implemented | add_market |
| Error: RD market opens before EOM market closes. | implemented |  _validate_setup |
| MarketClearing before Simulation runs (Chiara/Year definition) | maybe solved? | See Error: Market time exceeds simulation time.  |
| Warning: Refered market for UnitOperator does not exist. | implemented | add_unit_operator |
| Warning: No demand but generation or vice versa. | implemented | _validate_setup |
| PyTest for more sophisticated validations | implemented (can always be extended) | test_world_validation |
| Error: Market.additional_fields does not match bidding strategies. | **discussion needed: Strategies handle additional fields quite differently.** | |
| Warning: Performance can be improved by fewer database writes | **specification needed: Where to assess this?** | |
| Error: No UnitOperator defined. |   | |

maurerle:
We should also fail the simulation if less than two participants have a strategy for any of the existing markets @jrasko

mthede:
When using e.g. the loader_pypsa.py with market_mechanism=redispatch, two markets are created with the same name 

<img width="277" height="44" alt="Image" src="https://github.com/user-attachments/assets/7820aec8-8e4d-4624-9758-c72264254848" />

Would it make sense to validate the uniqueness of the market_id?

PR #684, opened 2025-11-18
Config validation

carl-wanninger:
## Related Issue
Relates to https://github.com/assume-framework/assume/issues/664 .

## Description
This branch featues a draft for config validation and respective tests.

## Checklist


assume/world.py
```diff
@@ -698,6 +699,79 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+        # Integrity of schedule.
+        for market_id, market_config in self.markets.items():
+            market_start = market_config.opening_hours[0] 
+            market_end = market_config.opening_hours[-1]
+            
+            if market_start < self.start or market_end > self.end:
+                msg = (f"Market {market_id} violates world schedule. \n"
+                       f"Market start: {market_start}, end: {market_end}. \n)"
+                       f"World start: {market_start}, end: {market_end}.)")
+                raise ValueError(msg)
```
jrasko: maybe its best to move this to the `add_market` method so the simulation fails even earlier, when adding an invalid market :thinking:

assume/world.py
```diff
@@ -698,6 +699,79 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+        # Integrity of reference.
+        # For each UnitOperator: strategies must reference existing markets.
+        unit_operators = list(self.unit_operators.values())
+        for operator in unit_operators:
+            for market_id in operator.portfolio_strategies.keys():
+                if market_id not in list(self.markets.keys()):
+                    msg = (f"Strategies of unit operator {operator} references"
+                           f"market {market_id} which is not known in world."
+                           f"Known markets are:\n{list(self.markets.keys())}.")
+                    raise ValueError(msg)
```
jrasko, line 690: see above, this may be moved to the `add_unit_operator` method

assume/world.py
```diff
@@ -698,6 +699,79 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+
+        # Each market should be referenced by a market strategy.
+        referenced_markets = {market 
+                              for market in operator.portfolio_strategies.keys()
+                              for operator in unit_operators}
+        for market_id in self.markets.keys():
+            if not market_id in referenced_markets:
+                msg = f"Added market {market_id}, has no participants."
+                warnings.warn(msg)
```
jrasko, line 701: I think this part is already ensured by the unit_operator constructor, where a new unit operator references all available markets automatically (i may be wrong though)

assume/world.py
carl-wanninger: I think you are right, that `add_market` would be the best place for this part if we organize validation locally. Regarding a possible `DataValidation` class, to me it sounded more like that the validation is intended to be done centrally for all data.

assume/world.py
```diff
@@ -784,6 +860,8 @@ def run(self):
         container.
         """
 
+        self._validate_setup()
```
kim-mskw: why do we need to do this in run and asynch_run?

assume/world.py
```diff
@@ -698,6 +699,79 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+        # Each market should be referenced by a market strategy.
+        referenced_markets = {market 
+                              for market in operator.portfolio_strategies.keys()
+                              for operator in unit_operators}
+        for market_id in self.markets.keys():
+            if not market_id in referenced_markets:
+                msg = f"Added market {market_id}, has no participants."
+                warnings.warn(msg)
+        
+        # Real-world integrity.
```
kim-mskw: This seems to be perfect for the post_init method of the config, right?

assume/world.py
carl-wanninger: We do not. I oversaw that `async_run` is called by `run`.

assume/world.py
carl-wanninger: Hmm. You are right, there is a similar logic in `UnitsOperator.__init__`. However (and this is why the corresponding test works): You theoretically can add a market to `World` after you added all `UnitsOperator`s.

This makes we wonder what the purpose of the respective logic in `UnitsOperator.__init__` is.

```python
for market in self.available_markets:
            if market.market_id not in self.portfolio_strategies.keys():
                self.portfolio_strategies[market.market_id] = (
                    DirectUnitOperatorStrategy()
                )
```

assume/world.py
```diff
@@ -698,6 +699,79 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+        if len(redispatch_markets) > 0:
+            if len(dispatch_markets) == 0:
+                msg = "Redispatch market but no dispatch market was defined."
+                raise ValueError(msg)
+            
+            earliest_dispatch_closing = min(x.opening_hours[-1]
+                                            for x in dispatch_markets)
+            earliest_redispatch_opening = min(x.opening_hours[0]
+                                              for x in redispatch_markets)
+
```
carl-wanninger, line 728: This logic will not work for markets that open multiple times.

x.opening_hours[0] + x.opening_duration

seems to be a applicable alterantive.

assume/world.py
maurerle: > You theoretically can add a market to World after you added all UnitsOperators.

Yes, theoretically - it is currently not supported though. So one first has to create the markets and then add the participants.

Fixing this would require to somehow inject the information of available markets prior to the start of the simulation into all UnitsOperators, which would add another layer of magic/complexity to the simulation.

assume/world.py
maurerle: After creation of the market, you do not know if there will be participants.
After creation of a single agent, you do not know if there will be further agents.
So the only place to check if there is any UnitsOperator, which has a strategy for this market configured is right prior to the `run` execution call of the simulation, by iterating over all units of all UnitsOperators and checking the BiddingStrategies..

maurerle:
@Graf-Wronski this issue is a little bit like a whish list and I guess that some of the comments here should rather be placed as additional entries in https://github.com/assume-framework/assume/issues/664

This might be overwhelming at first, but remember to always do small steps without trying to solve everything at once :)

kim-mskw:
> @Graf-Wronski this issue is a little bit like a whish list and I guess that some of the comments here should rather be placed as additional entries in #664
> 
> This might be overwhelming at first, but remember to always do small steps without trying to solve everything at once :)

Moved them to the issue :)

assume/world.py
carl-wanninger: I am a little confused. Does "Not supported" mean a code like this should raise an Exception?

```python
world = setup_simple_world()

@pytest.fixture
def demand():
    return Demand(
        id="test_unit",
        unit_operator="test_operator",
        min_power=0,
        max_power=-1000,
        technology="demand",
        bidding_strategies={},
        forecaster=DemandForecaster(index, demand=-100),
    )


def test_addition_order(demand):
    """ Add participants before market. """
    
    
    world.add_unit_operator(
        "unit_operator",
        {"dispatch": "naive_eom"})
    world.add_unit_instance(
        operator_id="unit_operator",
        unit=demand)
    
    opening_dispatch = rr.rrule(rr.HOURLY, dtstart=world.start, until=world.end)
    
    market_products= [MarketProduct(
        duration=datetime.timedelta(hours=1),
        count=1,
        first_delivery=datetime.timedelta(hours=1))]
    
    dispatch_config = MarketConfig(
        "dispatch",
        opening_hours=opening_dispatch,
        market_products=market_products)
    
    world.add_market_operator("market_operator")
    world.add_market(
        market_operator_id="market_operator",
        market_config=dispatch_config)
    
    world.run()
```

maurerle: @Graf-Wronski yes exactly - the `add_unit_operator` runs

```python
UnitsOperator(
    available_markets=list(self.markets.values()),
    portfolio_strategies=bidding_strategies,
)
```

if the bidding_strategies include markets which are not part of available_markets - this will fail.
We should handle this case and throw an error there.

assume/world.py
maurerle: You can also use `x.opening_hours.after(simulation.start)`

assume/world.py
carl-wanninger: To me the test as written above does not raise an Exception.

Maybe I overlook something. We can discuss this in the meeting tomorrow.

maurerle: Yes right, but if the example relies on this market, the unit operator will just not know about it.

As this is intended behavior for examples like `example_01d` (having a EOM and redispatch market in one case and a `DAM` in the other), we need to allow this somehow.

So we only show a warning there telling `you configured bidding strategies for market X,Y and Z which do not exist` or something, which can be ignored by the user if this is fine.

assume/world.py
carl-wanninger: Moved to add_unit_operator.

assume/world.py
```diff
@@ -647,10 +659,97 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+        # For each market: Should be referenced by a market strategy.
+        referenced_markets = {
+            market
+            for market in operator.portfolio_strategies.keys()
+            for operator in unit_operators
+        }
+        for market_id in self.markets.keys():
+            if market_id not in referenced_markets:
+                msg = f"Added market {market_id}, has no participants."
```
kim-mskw: Is this error message correct? Unit operators can register for a market and then they are participants but they just do not have a bidding strategy, or am I missing something?

assume/world.py
```diff
@@ -647,10 +659,97 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+                    "market has closed."
+                )
+                raise ValueError(msg)
+
+        # Existence of demand implies existence of generation and vice versa.
+        demand_exists, generation_exists = False, False
+        for operator in unit_operators:
+            for unit in operator.units.values():
+                # ToDo: Are the definitions of demand and generation exhaustive?
+                if type(unit) in [self.unit_types["demand"]]:
```
kim-mskw: This might not be exaustive, especially in the future. Hence I would suggest specifiying the warning.

kim-mskw: What do you think about my suggestion?

assume/world.py
```diff
@@ -647,10 +659,98 @@ def add_market(self, market_operator_id: str, market_config: MarketConfig) -> No
...
+                raise ValueError(msg)
+
+        # Existence of demand implies existence of generation and vice versa.
+        demand_exists, generation_exists = False, False
+        for operator in unit_operators:
+            for unit in operator.units.values():
+                if type(unit) in [self.unit_types["demand"]]:
+                    demand_exists = True
+                elif type(unit) in [
+                    self.unit_types["power_plant"],
```
kim-mskw: Is there a way to check this for particular markets? I think that was @maurele thought in his comment

assume/world.py
carl-wanninger: I think theoretically you are right. 

Ill make the warning more precise.

assume/world.py
carl-wanninger: I like it and adjusted the snippet to be more specific and extendable.

assume/world.py
carl-wanninger: In my opinion this is not possible.

If only markets were only adressed by units and their bidding strategies directly, it would be possible.

But as far as I understand e.g. `class EnergyNaiveStrategy(MinMaxStrategy)`, unit operators can place bids regardless of the bidding strategies of the single units and their known markets. So ultimately knowing which unit types will participate at which market is not possible a priori.

One could however extend the check as that it is required for each market that there exists an unit operator that has a demand unit - and one that has a generation unit.

Would you prefer this?

kim-mskw: No I think then it is fine as it is. devinetly an improvement :)

kim-mskw (approved):

PR #690, opened 2025-11-21
Change Notebooks to local installation of latest dev state

### LINKED ISSUE #678: Colab Notebooks: conflicting naming of forecasts and forecaster
Since we introduced per unit forecasters in https://github.com/assume-framework/assume/pull/660 the file assume.common.forecasts was renamed to "forecaster.py".
This leads to an import error when running Colab Notebooks, where we first do pip install (current version 0.5.5, having "forecasts.py" as name of the file) and then git clone our repository (having "forecaster.py" as filename).

One solution woul be to install from the cloned repo instead of pypi:
```
if IN_COLAB:
    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo
    !pip install -e assume-repo
```
But maybe it makes more sense for users to use the pypi install?
It seems quite unintuitive for me to clone a whole repository and then mix it with pip install. But I have no good idea on how to organize this in a way, that helps to avoid such errors. -> discussion needed

Error occured in Notebooks: 03, 04a. Potentially affected: all Notebooks

gugrimm:
## Related Issue
Closes #678 

## Description
This changes the notebooks when executed in colab.
The motivation is to avoid errors that occur when both the latest release from pypi and the assume git repo are used in one notebook and the release is behind the git repo in terms of namings, data structure etc.

The proposed solution is to replace `!pip install assume-framework` by `!pip install -e .` in all notebooks, where we clone the git repo in order to have access to tutorial data. 

## Checklist
- no ducumentation required
- [x] New unit/integration tests added (not applicable)
- [x] Changes noted in release notes (not applicable)


kim-mskw:
On the review @jrasko and @maurerle

examples/notebooks/03_custom_unit_example.ipynb
```diff
@@ -133,17 +133,10 @@
...
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "if IN_COLAB:\n",
-    "    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo"
+    "    # you can install the latest release of assume-framework with !pip install assume-framework\n",
```
kim-mskw: Should we not do this consistently, then, to avoid future errors? If we decide to install the dev state, do we always?

examples/notebooks/04a_reinforcement_learning_algorithm_example.ipynb
```diff
@@ -119,7 +118,8 @@
    "outputs": [],
    "source": [
     "if IN_COLAB:\n",
-    "    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo"
+    "    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo\n",
+    "    !pip install -e ./assume-repo[learning]"
```
kim-mskw: It is weird that we do not install assume in the section that is called installation process. I would suggest restructuring the chapter as well

examples/notebooks/04b_reinforcement_learning_example.ipynb
```diff
@@ -166,7 +166,8 @@
    "source": [
     "# Only run this cell if you are using Google Colab\n",
     "if IN_COLAB:\n",
-    "    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo"
+    "    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo\n",
```
kim-mskw: The installation command in the markdown cell still references the normal installation, there also with ". -e" then?

kim-mskw: This happens in multiple tutorials. Please make changes accordingly.

examples/notebooks/05_market_comparison.ipynb
```diff
@@ -98,7 +96,8 @@
    "outputs": [],
    "source": [
     "if IN_COLAB:\n",
-    "    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo"
+    "    !git clone --depth=1 https://github.com/assume-framework/assume.git assume-repo\n",
+    "    !pip install -e assume-repo"
```
kim-mskw: Adjust the comment above that the code does not need to be executed when one has the current dev state installed, not the release.

examples/notebooks/07_interoperability_example.ipynb
```diff
@@ -107,13 +106,105 @@
...
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 53,
    "metadata": {
     "vscode": {
      "languageId": "shellscript"
     }
    },
-   "outputs": [],
+   "outputs": [
```
kim-mskw: clear outputs please and push again :)

kim-mskw (dismissed):
Thank you for fixing this. I have a few remarks regarding consistency:

- I would suggest using the *dev* state in **all** tutorials, regardless of whether it is currently required. This helps avoid future issues and keeps everything consistent.
- Related to consistency: I know it’s a bit tedious, but please also update the respective markdowns in the tutorials.  
  Some of them still do not reference the *dev* state, and they mention that installation cells do not need to be executed if you already have the package installed.  
  However, with only the latest release installed, you *do* need to run the installation cells again to obtain the dev version.

gugrimm:
@assume-framework/core-devs This is bigger than initially thought and made me think of our approach to versioning:
With the last commit, I added cases to the installation part in notebook 03 that would
- in colab: clone and pip install -e .
- locally: clone and pip install -e . when repository "assume" and "assume-repo" dont exist
- locally: git pull and pip install -e . when the repo exists
This is a little bit long. But the only thing I could think of. However it would not work for users without admin rights, the tests fail, and most importantly we force users to install things on their local PC without asking or the option to refuse. (It is also unclear to me what happens if I am in a local branch with local changes and then inside a ipynb notebook cell do !git pull. What does this pull? Main? Interestingly this does not overwrite my local changes, but also does not abort and not tells me to stash my work first. Another thing in .ipynb that I don't fully understand...)

After discussing this with Ramiz, we came up with 3 potential solutions:
1) introduce a dev branch and keep main same as the pypi release until the next release. We would still do developments on small branches, but then merge them into dev. When we do a release, we merge dev into main. The version and therefore the versioning number of pypi and github main would then always be the same.
2) with each PR update the version number in github main. This is currently not done (we have pypi 0.5.5 and github 0.5.5, although the versions are clearly NOT the same.) This would then allow in the notebooks (and anywhere else) to check versions and print warnings, as it is for example done in PyPSA (e.g. if importing an old network file to a new PyPSA version).
3) give responsibility to the user (notebooks not runnable out of the box): We could git clone and pip install -e if working in colab. If users work locally, they must make sure for themselves to use tha latest dev state from main. We would indicate this more clearly in the notebooks.

I would be in favor of 1) but thinking about consistent versioning seems to be a good idea to me anyway. What are your thoughts?

maurerle (commented):
In todays meeting we discussed that people who have checked out the repository should also install the checked out source.
This would be result in every notebook mentioning `pip install -e .` 

To install the learning dependencies (torch) as well, one can run `pip install -e .[learning]`.

examples/notebooks/04c_reinforcement_learning_storage_example.ipynb
```diff
@@ -106,7 +122,8 @@
     "IN_COLAB = importlib.util.find_spec(\"google.colab\") is not None\n",
     "\n",
     "if IN_COLAB:\n",
-    "    !pip install assume-framework"
+    "    !pip install assume-framework\n",
+    "    !pip install ./assume-repo"
```
maurerle:
```suggestion
    "    !pip install ./assume-repo"
```

maurerle (approved):
Great!

PR #692, opened 2025-11-26
Restructure learning config

mthede:
## Description
With the new learning architecture on main, we can streamline how default parameters are handled. Previously, defaults were defined redundantly in multiple places, making it difficult to determine which values applied in practice.

The updated structure centralizes all defaults within the learning role or in the dedicated learning_config data class, improving clarity and maintainability.

It also improves the learning configuration logic, based on four different simulation options (with and without DRL)
1. No RL -> no learning_config -> no learning role
2. Single run with loaded RL strategies -> learning_config with learning_mode: false and trained_policies_load_path is provided -> learning role, but no rl algorithm etc. necessary
3. Training run -> learning_mode: true and full set of learning functions
3.1. Training episodes
3.2. Evaluation episodes (config item evaluation_mode exists, but learning loop overwrites it anyway)
4. Continue learning -> continue_learning: true and trained_policies_load_path is provided (-> internally sets learning_mode = True for entering the usual learning process)

## Checklist

### DIFF
--- a/assume/reinforcement_learning/learning_role.py
+++ b/assume/reinforcement_learning/learning_role.py
@@ -48,6 +48,8 @@ class Learning(Role):
         start: datetime,
         end: datetime,
     ):
+        super().__init__()
+
         # how many learning roles do exist and how are they named
         self.buffer: ReplayBuffer = None
         self.episodes_done = 0
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -191,6 +191,8 @@ class WriteOutput(Role):
         )
 
     def on_ready(self):
+        super().on_ready()
+
         if self.db_uri:
             self.db = create_engine(self.db_uri)
         if self.db is not None:


assume/scenario/loader_csv.py
```diff
@@ -1175,11 +1182,10 @@ def run_learning(
 
     world.reset()
 
-    # Set 'trained_policies_load_path' to None in order to load the most recent policies,
-    # especially if previous strategies were loaded from an external source.
-    # This is useful when continuing from a previous learning session.
+    # final simulation run with best policies
+    # TODO: discussion - in case of convergence towards lower rewards, should we use the last policies instead?
```
mthede: The previous comment did no longer fit to the code, right? Maybe a discussion is needed (now or later).

assume/reinforcement_learning/learning_role.py
```diff
@@ -42,51 +42,32 @@ class Learning(Role):
 
     def __init__(
         self,
-        learning_config: LearningConfig,
+        learning_config,
         start: datetime,
         end: datetime,
     ):
```
jrasko, line 50: Since Learning is derived from Role, we need to call the super() init  somewhere in the constructor

assume/reinforcement_learning/learning_role.py
```diff
@@ -42,51 +42,32 @@ class Learning(Role):
 
     def __init__(
         self,
-        learning_config: LearningConfig,
+        learning_config,
```
jrasko: keep the type hint here

assume/reinforcement_learning/learning_role.py
```diff
@@ -42,51 +42,32 @@ class Learning(Role):
...
-            if "cuda" in learning_config.get("device", "cpu")
+            self.learning_config.device
+            if self.learning_config and "cuda" in self.learning_config.device
+            else "cpu"
+        )
+        # always use CPU in evaluation mode for performance reasons
+        self.device = th.device(
+            cuda_device
+            if (th.cuda.is_available() and not self.learning_config.evaluation_mode)
             else "cpu"
```
jrasko: i would recommend to merge these multiple statements to determine the device into one

assume/world.py
```diff
@@ -221,10 +221,13 @@ def setup(
...
-        self.learning_config = learning_config
         # initiate learning if the learning mode is on and hence we want to learn new strategies
-        self.learning_mode = self.learning_config.get("learning_mode", False)
-        self.evaluation_mode = self.learning_config.get("evaluation_mode", False)
+        self.learning_mode = (
+            learning_config.get("learning_mode") if learning_config else False
+        )
+        self.evaluation_mode = (
+            learning_config.get("evaluation_mode") if learning_config else False
+        )
```
jrasko: just keep the shorter get() function with default value

assume/common/base.py
```diff
@@ -779,6 +918,10 @@ def __init__(
         """
         super().__init__(*args, **kwargs)
 
+        # access to the learning_role that orchestrates learning
+        self.learning_role = learning_role
+        self.learning_config = learning_role.learning_config
```
maurerle, line 924: learning_role can be None here, so this does not work.

assume/world.py
```diff
@@ -293,41 +300,43 @@ def setup(
...
-            )
-
-            rl_agent = agent_composed_of(
-                self.learning_role,
-                register_in=self.container,
-                suggested_aid="learning_agent",
-            )
-            rl_agent.suspendable_tasks = False
+            # create LearningConfig object
+            learning_config = LearningConfig(**learning_config)
```
maurerle: I would rename the dict to `learning_config_dict` or just `learning_dict` (in world.setup, setup_world, world.setup_learning)- overwriting the same variable with a different type is hard to detect otherwise.

Having the variable named learning_config but it is not of type LearningConfig is confusing

assume/scenario/loader_csv.py
```diff
@@ -1047,8 +1055,8 @@ def run_learning(
     world.learning_role.rl_algorithm.initialize_policy()
 
     # check if we already stored policies for this simulation
-    save_path = world.learning_config["trained_policies_save_path"]
-    continue_learning = world.learning_config.get("continue_learning", False)
+    save_path = world.learning_role.learning_config.trained_policies_save_path
+    continue_learning = world.learning_role.learning_config.continue_learning
```
maurerle, line 1058: Are you sure `world.learning_role` is not None here?
This was `world.learning_config.trained_policies_save_path` in the learning_config_restructure branch

assume/scenario/loader_csv.py
```diff
@@ -756,26 +755,35 @@ def setup_world(
...
-        learning_config["learning_mode"] = False
-        learning_config["evaluation_mode"] = False
-
-    if not learning_config.get("trained_policies_save_path"):
-        learning_config["trained_policies_save_path"] = (
-            f"learned_strategies/{simulation_id}"
-        )
+    learning_config = config.get("learning_config")
+    # those settings need to be overridden before passing to the LearningConfig
+    if learning_config:
```
maurerle: If we use LearningStrategies, but do not have a learning_config set - an error should be raised quite early.
Then we can remove this check here?

tests/test_learning_role.py
```diff
@@ -25,28 +25,31 @@
...
     }
+
     # test init
-    learn = Learning(learning_config, start, end)
+    learn = Learning(config["learning_config"], start=start, end=end)
     assert len(learn.rl_strats) == 0
 
     # we need to add learning strategies first
-    learn.rl_strats["test_id"] = LearningStrategy(**learning_config)
+    learn.rl_strats["test_id"] = LearningStrategy(**config, learning_role=learn)
```
maurerle, line 52:
```suggestion
    learn.rl_strats["test_id"] = LearningStrategy(**config["learning_config"], learning_role=learn)
```
This should not have the whole config itself passed right?
I would like to ditch this, by removing the unpacked dict:

```suggestion
    learn.rl_strats["test_id"] = LearningStrategy(learning_role=learn)
```

As we have access through `learning_role.learning_config` anyway

assume/common/base.py
mthede: Actually, with this PR's restructuring, a learning role should always exist if LearningStrategies are used. So it can be moved as a required argument. I originally wanted to be typed, but that caused circular import issues ...

assume/scenario/loader_csv.py
mthede: I agree with raising an error in that case - need to check the current behavior in that case. 

But this check is to avoid creating a learning_config at all if no RL is used (no training and no LearningStrategies as bidding strategies). Before, there was always a learning_config with defaults created.

assume/world.py
mthede: Sure! Don't know why I overcomplicated this :D

assume/reinforcement_learning/learning_role.py
mthede: I'm not sure if this is mango agents specific (@maurerle: please help out here) but we haven't had that before. I think the "init" equivalent is called below with on_ready()?

tests/test_learning_role.py
mthede: It needs the rest of the config (in particular obs_dim, act_dim, unique_obs_dim) which is currently not specified within the learning_config. It's learning strategy specific and currently set as default values in the classes (e.g. EnergyLearningStrategy with kwargs.pop()). So we need to pass it in case of this test to the superclass. But I'm open to restructuring it and not leaving it as before in the dict!

assume/scenario/loader_csv.py
mthede: Yes. Otherwise, the line above with intialize policies shouldn't have worked before either?

assume/world.py
```diff
@@ -537,11 +542,13 @@ def _prepare_bidding_strategies(self, unit_params, unit_id):
...
                 # check if created cache has learning_strategy
-                if (
-                    issubclass(self.bidding_strategies[strategy], LearningStrategy)
-                    and self.learning_mode
-                ):
+                if issubclass(self.bidding_strategies[strategy], LearningStrategy):
                     # add learning role to the strategy to have access to store training data etc
+                    if not hasattr(self, "learning_role") or self.learning_role is None:
+                        raise ValueError(
+                            f"Learning strategy '{strategy}' requires a configured 'learning_role', but none was found. "
```
mthede: @maurerle is failing here too late?

assume/scenario/loader_csv.py
maurerle: True - the learning_role is never None anymore.

assume/common/base.py
maurerle: Without the default to None, this looks good - thanks.

assume/world.py
maurerle: I would switch to always having a self.learning_config which can be none, and creating the config early.
So no need for `hasattr` (which in my opinion almost always tells you that something is wrong about your input assumptions).

I just pushed a commit which tries to implement this.

jrasko:
> @jrasko Will look into the cuda problem tomorrow! Could you please try removing [the "and not self.learning_config.evaluation_mode" in the learning role](https://github.com/assume-framework/assume/blob/485b1ae60c5a1f84d962aca3fb9dafb69e36ea24/assume/reinforcement_learning/learning_role.py#L65)?

@mthede yep, thats why. Removing this condition fixes the error.

assume/reinforcement_learning/learning_role.py
```diff
@@ -50,43 +52,22 @@ def __init__(
...
-        cuda_device = (
-            learning_config["device"]
-            if "cuda" in learning_config.get("device", "cpu")
+        self.device = th.device(
+            self.learning_config.device
+            if (
+                self.learning_config
+                and "cuda" in self.learning_config.device
+                and th.cuda.is_available()
+                and not self.learning_config.evaluation_mode  # always use CPU in evaluation mode for performance reasons
```
mthede: This leads to tensors on different devices (cf. https://github.com/assume-framework/assume/pull/692#issuecomment-3588364747). Are you ok that we'll remove this @maurerle?

maurerle: Yes, sounds good!

assume/reinforcement_learning/neural_network_architecture.py
```diff
@@ -197,7 +197,7 @@ def __init__(
         obs_dim: int,
         act_dim: int,
         float_type,
-        unique_obs_dim: int = 0,
+        unique_obs_dim: int,
```
kim-mskw, line 200: num_timeseries_obs_dim appears to be set as a default value twice. Is that intentional?

assume/strategies/learning_strategies.py
```diff
@@ -59,42 +59,39 @@ def __init__(self, *args, **kwargs):
...
-                dt=kwargs.get("noise_dt", 1.0),
+                scale=self.learning_config.noise_scale,
+                dt=self.learning_config.noise_dt,
             )
 
             self.learning_role.register_strategy(self)
 
-        elif Path(kwargs["trained_policies_load_path"]).is_dir():
-            self.load_actor_params(load_path=kwargs["trained_policies_load_path"])
+        # actor policies are only loaded from file if learning mode is off
```
kim-mskw: Why? What about continue learning?

mthede: Policies from continue learning are initialized here: https://github.com/assume-framework/assume/blob/e3adddb66d878758eefb2b9b9b22609334975016/assume/reinforcement_learning/learning_role.py#L469
After that inital load, the latest policies are kept in actors_and_critics between the learning episodes. 

And I've decided to set learning_mode = True when continue_learning = True, because it should trigger the same learning process. So even with continue learning, the first if clause evaluates to true.

kim-mskw: Fair did not pay enough attention, just quickly changed the comment

kim-mskw (approved):
all the changes I wanted where discussed bilateral

assume/reinforcement_learning/neural_network_architecture.py
mthede: Good catch! Will remove the default here, because it's set in the learning strategy and then passed here

tests/test_learning_role.py
kim-mskw: Fair point @maurerle, but it is only the test so I would not spend a lot of energy on it.

tests/test_drl_storage_strategy.py
```diff
@@ -107,7 +106,7 @@ def test_storage_rl_strategy_sell_bid(mock_market_config, storage_unit):
         # Mock the calculate_marginal_cost method to return a fixed marginal cost
         with patch.object(Storage, "calculate_marginal_cost", return_value=10.0):
             # Calculate bids using the strategy
-            bids = strategy.calculate_bids(
+            bids = strategy.calculate_bids(  # TODO
```
maurerle, line 109: What did this TODO say?
At least add a comment here?

assume/reinforcement_learning/learning_role.py
maurerle: It is not super important (pun intended), as the mango Role init does not do a lot, but technically we should call the base class constructor, to be prepared if something gets added in the mango role constructor in the future.

PR #703, opened 2025-12-10
Add Nodal Pricing with OPF

### LINKED ISSUE #688: Unintuative Redispatch Behavior
The combination of redispatch and a nodal market design using the complex_clearing method produced highly unintuitive results.
Even though a nodal market design inherently accounts for all technical constraints of the underlying grid, we still observe redispatch in the subsequent simulation step.
Why is redispatch happening at all?

## Problem Description

From an energy economics perspective, redispatch after a nodal market clearing should not occur, since nodal pricing internalizes grid constraints directly in the optimization.
One possible explanation could be that the grid representation in complex_clearing is too simplified, resulting in cases that are considered feasible in the clearing but become infeasible once evaluated with the more detailed grid model in PyPSA, thereby triggering redispatch.

However, this discrepancy should not occur if both models describe the same physical system consistently. 
## Open Questions

Is our market → redispatch workflow reliable, or is there a conceptual issue? Am I missing something obvious?

Did someone see such behavior with zonal markets and redispatch so that redispatch is happening between zones still?

Or is this caused by another bug (clearing logic, grid representation, data transfer, etc.)?

## Reproducabilty

You can reproduce that by running example_01d on this branch: [redispacth_and_nodal_pricing_bug](https://github.com/assume-framework/assume/tree/redispacth_and_nodal_pricing_bug)

kim-mskw:
## Related Issue
Closes #688 

## Description
This PR fixes unintuitive redispatch results when using **nodal market design with `complex_clearing`**. In the current implementation, `complex_clearing` behaves as a **transport model** based on NTCs and **ignores line reactances `x`**, even if they are present in `lines.csv`. As a result, the market clearing can produce dispatch patterns that are **infeasible** for the underlying PyPSA grid model, which then shows up as large redispatch needs and even unmet demand in the subsequent simulation.

### Changes in this PR

- **Warning for `complex_clearing`**  
  - When `complex_clearing` is used and a column `x` is present in `lines.csv`, a warning is issued that the underlying transport model does **not** take grid physics (reactances) into account.

- **New nodal clearing with physics-based grid representation**  
  - Introduces a **nodal clearing** that:
    - Uses **line reactances** and a **PTDF-based formulation** (via PyPSA’s OPF).
    - Formulates nodal energy balances for non-slack buses plus a system-wide balance, ensuring physical feasibility of dispatch and flows.
  - For `example_01d` this new nodal clearing produces **feasible dispatch and flows without additional redispatch** (storage tests still pending).

- **Direct extraction of dispatched bids**  
  - Adds a code path to extract **accepted bids directly from the optimization model**, avoiding the LMP-based reconstruction of dispatch.  

## Checklist


assume/markets/clearing_algorithms/nodal_clearing.py
```diff
@@ -0,0 +1,306 @@
...
+        self.network.add("Generator",
+                         self.grid_data['loads'].index,
+                         bus=self.grid_data['loads']['node'],
+                         p_nom=self.grid_data['loads']['max_power'],
+                         p_min_pu=-1,
+                         p_max_pu=0,
+                         )
+        # storage units
+        # TODO: how are storages included in grid data dict?
+        # also add them as generators, as we only regard bids here
```
kim-mskw: @mthede, have you had a look into this during your thesis?

assume/markets/clearing_algorithms/nodal_clearing.py
```diff
@@ -0,0 +1,306 @@
...
+        # set snapshots as the products start time, end time and freq = duration of the products
+        # snapshots = pd.date_range(self.marketconfig.opening_hours._dtstart,
+        #                          self.marketconfig.opening_hours._until,
+        #                          freq=self.marketconfig.market_products[0].duration)
+        n.set_snapshots(snapshots)
+
+        # Update p_max_pu for all units based on their bids in the actual snapshots
+        # generators
+        gen_idx = self.grid_data['generators'].index
+        n.generators_t.p_max_pu.loc[snapshots, gen_idx] = volume_pivot[gen_idx] / n.generators.loc[gen_idx, 'p_nom'].values
```
kim-mskw: What happens if a generator does not submit to this market? Just being in the grid_data dos not mean I necessarly submit to the market, correct?

Could it be that if a powerplant does not place an order it is still dispatched because max_power was added beforehand? If so I would initialize max_power with zero when adding generators to avoid this.

assume/markets/clearing_algorithms/nodal_clearing.py
```diff
@@ -0,0 +1,306 @@
...
+        n.generators_t.marginal_cost.loc[snapshots, gen_idx] = price_pivot[gen_idx]
+        # demand
+        demand_idx = self.grid_data['loads'].index
+        n.generators_t.p_min_pu.loc[snapshots, demand_idx] = volume_pivot[demand_idx] / n.generators.loc[demand_idx, 'p_nom'].values
```
kim-mskw: Why are you setting p_min for the demand?

assume/markets/clearing_algorithms/nodal_clearing.py
mthede: As I've just considered the cleared volumes for redispatch calculations in redispatch.py, I added them similarly to the generators and loads (like methods `add_redispatch_generators ` and `add_redispatch_loads`) as loads with fixed infeed.

mthede: And I just added them in read_grid() as another output

assume/markets/clearing_algorithms/nodal_clearing.py
gugrimm: as we have only bids from generators bidding in this market in the volume_pivot, I don't see a problem. Or am I wrong?

assume/markets/clearing_algorithms/nodal_clearing.py
gugrimm: demand is negative, so p_min_pu represents the maximum consumption at the bid price

assume/markets/clearing_algorithms/nodal_clearing.py
kim-mskw: Ah fair, so then a different error might occur: If gen_idx contains IDs not present in volume_pivot.columns, volume_pivot[gen_idx] raises a KeyError.

assume/markets/clearing_algorithms/nodal_clearing.py
kim-mskw: Ah thanks!

assume/markets/clearing_algorithms/nodal_clearing.py
```diff
@@ -0,0 +1,403 @@
...
+# Set the log level to WARNING
+logging.getLogger("linopy").setLevel(logging.WARNING)
+logging.getLogger("pypsa").setLevel(logging.WARNING)
+
+
+def calculate_meta(accepted_demand_orders, accepted_supply_orders, product):
+    supply_volume = sum(map(itemgetter("accepted_volume"), accepted_supply_orders))
+    demand_volume = -sum(map(itemgetter("accepted_volume"), accepted_demand_orders))
+    prices = list(map(itemgetter("accepted_price"), accepted_supply_orders)) or [0]
+    # can also be self.marketconfig.maximum_bid..?
```
kim-mskw: Good question, I guess the questions is wheter the meta should describe the empirical market results or wheter it should describe the theoretical market config. As we also report the supplied demand etc and avg price I would suggest that these metrics hould be descriptive of the market result.

assume/markets/clearing_algorithms/nodal_clearing.py
gugrimm: Thanks, changed gen_idx, demand_idx and storage_idx to be the intersection of grid_data and orderbook units.

PR #705, opened 2025-12-11
export a world simulation as json to allow ASSUME-GUI import

jrasko:
Signed-off-by: Jannik <raskob@fh-aachen.de>

## Description
add a json encoder to allow a visualisation of the simulation using assume-gui

## Checklist

### DIFF
--- a/assume/__init__.py
+++ b/assume/__init__.py
@@ -5,11 +5,6 @@
 from importlib.metadata import version
 
 from assume.common import MarketConfig, MarketProduct
-from assume.scenario.loader_csv import (
-    load_custom_units,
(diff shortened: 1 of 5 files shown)


maurerle (commented):
I do like the idea but wait to merge this until we have the current release out.
We probably also need to have a json importer?

jrasko:
> I do like the idea but wait to merge this until we have the current release out. We probably also need to have a json importer?

We currently have the json importer in the assume-gui part of the framework. I'd like to also move the import logic in this repository, so that it is closer to the code and can adjusted easier if the code changes.

PR #709, opened 2025-12-16
Soc renaming

### LINKED ISSUE #525: SoC name is misleading as it is used with absolute values

gugrimm:
## Related Issue
Closes #525 

## Description
This pull request standardizes the handling of energy storage units across the codebase by consistently using a normalized state of charge (`soc` between 0 and 1) and a new `capacity` field (in MWh) instead of the previously used `max_soc`. It updates storage models, strategies, and scenario loaders to use these new conventions, ensuring calculations and constraints are based on normalized values and absolute capacity. This improves clarity, correctness, and interoperability between different modules.

**Core data model updates:**
* Replaced `max_soc` with `capacity` in the `SupportsMinMaxCharge` class and related storage classes, and updated all references accordingly. The `soc` is now always a float between 0 and 1, and all calculations are adjusted to use `capacity` for absolute values. 
* Updated docstrings and function arguments to clarify that `soc` is normalized and to document the new `capacity` field. 

**Calculation and logic changes:**
* Adjusted all calculations involving state of charge, charging/discharging, and energy deltas to use normalized `soc` and scale by `capacity`. This affects dispatch planning, bid calculations, and reward functions.
* Updated optimization models and constraints to use `capacity` and normalized `soc`, including Pyomo model bounds and initializations. 

**Scenario and loader updates:**
* Modified all scenario loaders (`loader_amiris.py`, `loader_csv.py`, `loader_oeds.py`, `loader_pypsa.py`, `oeds/infrastructure.py`) to output `capacity` and normalized `initial_soc`, and to require the `capacity` field in input data. 

These changes make the codebase more robust and less error-prone by enforcing a clear separation between normalized and absolute values for storage units.

## Checklist
- New unit/integration tests added (not applicable)


assume/scenario/loader_oeds.py
```diff
@@ -302,8 +302,9 @@ def load_oeds(
                     {
                         "max_power_charge": storage["max_power_charge"] / 1e3,
                         "max_power_discharge": storage["max_power_discharge"] / 1e3,
-                        "max_soc": storage["max_soc"] / 1e3,
-                        "min_soc": storage["min_soc"] / 1e3,
+                        "capacity": storage["capacity"] / 1e3,
+                        "max_soc": storage["max_soc"],
```
maurerle: max_soc is not used anymore right?

tests/test_building.py
```diff
@@ -38,13 +39,14 @@ def thermal_storage_config(generic_storage_config):
...
-        "min_capacity": 0,
+        "capacity": 10.0, # EV battery capacity in MWh
+        "min_soc": 0,
+        "max_soc": 1,
         "max_power_charge": 3,  # Charge values will reflect a fraction of the capacity
         "max_power_discharge": 2,  # Discharge values will also be a fraction of the capacity
         "efficiency_charge": 0.95,
         "efficiency_discharge": 0.9,
-        "initial_soc": 0,  # SOC initialized to 50% of capacity
+        "initial_soc": 0,  # initial SOC
```
maurerle, line 49: Oh well..

assume/units/dst_components.py
```diff
@@ -448,23 +450,24 @@ def __init__(
 
         # check if initial_soc is within the bounds [0, 1] and fix it if not
         if initial_soc > 1:
-            initial_soc /= max_capacity
+            initial_soc = 1.0
             logger.warning(
```
kim-mskw, line 453: I think I would throw an error here. Beforehand, an SoC higher than 1 indicated that it was interpreted as an absolute value bounded by the capacity and was set to a relative value. 
If we still get an SoC hgher than 1 now there must be an nconsitency somewhere. Setting it to 1 simply disregrads that the initial SoC is maybe 2 (capacity is 200), so the correctly scaled SoC would be far from 1, with this overwirte and warning this might be overlooked.

PR #714, opened 2026-01-07
consistently use s_nom * s_max_pu in line capacities

gugrimm:
## Description
Using s_nom * s_max_pu for line capacities was introduced in the latest release. This is a helpful feature, as it allows to model restricted line capacity scenarios, redispatch, or n-1 without touching the s_nom parameter of lines. Approach is consistent with PyPSA.
Unfortunately it was not done in the congestion_forecast, the nodal_dashboard and some notebooks. This is hereby fixed.

## Checklist
- [x] New unit/integration tests added (not applicable)

### DIFF
--- a/examples/notebooks/10_DSU_and_flexibility.ipynb
+++ b/examples/notebooks/10_DSU_and_flexibility.ipynb
@@ -1875,7 +1875,6 @@
     "    s_max_pu=list(lines.s_max_pu),\n",
     "    s_nom_extendable=False,\n",
     "    x=list(lines.x),\n",
-    "    s_max_pu=1,\n",
     "    r=list(lines.r),\n",
     ")"
    ]


docs/source/redispatch_modeling.rst
```diff
@@ -57,6 +57,7 @@ Attributes
 4. **Line**: Transmission grids that transmit electricity
     - **name**: Unique identifier of the line
     - **s_nom**: Nominal transmission capacity
+    - **s_max_pu**: Maximum loading as a fraction of s_nom
```
jrasko: Thanks for documenting the meaning of this - I already wondered 😄. As far as I can tell from the usage here *s_max_pu* must be between 0 and 1, right? maybe we should validate or at least document this.

gugrimm: Yes, s_max_pu should be between 0 and 1 normally. I updated the docu accordingly.
I am however hesitant with validating. From PyPSA I am used to "play around" with this factor in simulations. It is useful when simulating a zonal dispatch without clustering nodes or reconfiguring all buses, lines, assets and availability timeseries just by setting s_max_pu very high. So I think there is a use case for setting it high. If users don't set it, it defaults to 1. And users familiar with pypsa might find it useful to set it higher...

maurerle (approved):
Looks good! :)

PR #715, opened 2026-01-07
Global seeding for reproducibility

### LINKED ISSUE #713: Improve reproducibility: global seeding for DRL runs
I repeatedly ran into issues with reproducibility when running DRL examples. We currently set a random seed in the DRL notebook, but this does not cover all relevant RNG sources, so results are still not fully deterministic across runs.

@chiara-fb suggested the following more comprehensive seeding setup:

```python
SEED = 42
random.seed(SEED)
np.random.seed(SEED)
th.manual_seed(SEED)
th.cuda.manual_seed(SEED)
th.cuda.manual_seed_all(SEED)

th.backends.cudnn.deterministic = True
th.backends.cudnn.benchmark = False
```

Based on my understanding, this should improve determinism substantially (at least for the major RNG sources in Python / NumPy / PyTorch, and cuDNN settings).

Before implementing this only in the notebook, **I’d like to discuss whether we should support global seeding at the framework level**—i.e., providing a single place to configure a seed that applies across the entire simulation/training run.

Questions / discussion points
- Do we want an official seed parameter in the config (or scenario setup) that initializes all RNG sources consistently?
- Where would be the best place to set this (e.g., during environment/world setup vs. trainer initialization)?
- Are there known sources of non-determinism we should explicitly document (e.g., GPU ops, parallelism, environment randomness)?

mthede:
## Related Issue
Closes #713

## Description
Adds functionality to set global seeds. This enables reproducible simulations, i.e. for DRL simulations. Seeds in notebooks are now set more comprehensively, ensuring deterministic simulation output of showcased examples. 

## Checklist

## Additional Notes
- Default deterministic behavior applies for loaded scenarios with loader_csv.py, not other loaders like loader_amiris, loader_oeds, loader_pypsa or custom loaders.
- Simulations with PyTorch-based learning strategies may experience performance decreases due to determinism. It's recommended to set `seed: 'null'` for DRL simulations and only, if reproducibility is necessary, make it deterministic. DRL convergence stability should be investigated between multiple runs with different random initializations.
- Full reproducibility cannot be guarenteed across hardware and software versions, see https://docs.pytorch.org/docs/stable/notes/randomness.html.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -50,17 +50,20 @@ def test_convert_rrule():
 
 def test_reproducability_with_seed():
     set_random_seed(42)
-    rand_nums_1 = np.random.rand(5)
+    rand_nums_1_1 = np.random.rand(5)
+    rand_nums_1_2 = np.random.rand(5)
 
-    rand_nums_2 = np.random.rand(5)
+    set_random_seed(42)
+    rand_nums_2_1 = np.random.rand(5)
+    rand_nums_2_2 = np.random.rand(5)
 
-    assert np.array_equal(rand_nums_1, rand_nums_2), (
-        "Random numbers should be the same for the same seed"
-    )
+    assert np.array_equal(
+        [rand_nums_1_1, rand_nums_1_2], [rand_nums_2_1, rand_nums_2_2]
+    ), "Random numbers should be the same for the same seed"
 
-    set_random_seed("null")
+    set_random_seed(None)
     rand_nums_3 = np.random.rand(5)
-    assert not np.array_equal(rand_nums_1, rand_nums_3), (
+    assert not np.array_equal(rand_nums_1_1, rand_nums_3), (
         "Random numbers should differ for different seeds"
     )
 


assume/scenario/loader_csv.py
```diff
@@ -492,6 +493,14 @@ def load_config_and_create_forecaster(
         study_case = list(config.keys())[0]
     config = config[study_case]
 
+    # Default deterministic behavior
```
mthede: Can someone think of a better place to set the seeds, which may also apply for other loaders? world.setup() may be called multiple times for RL simulations, so I thought that it's not a suitable place.

assume/common/utils.py
```diff
@@ -843,3 +844,39 @@ def confirm_learning_save_path(save_path: str, continue_learning: bool) -> None:
...
+    try:
+        import torch as th
+
+        if "CUBLAS_WORKSPACE_CONFIG" not in os.environ:
+            os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
+
+        th.manual_seed(seed)
+
+        if torch_deterministic:
+            th.backends.cudnn.deterministic = True
```
mthede, line 875: This may introduce significant performance decreases (read about ranges between 15-200%, so actually quite high!). Should we still stick with deterministic (DRL) simulations as a default?

assume/common/utils.py
```diff
@@ -843,3 +844,39 @@ def confirm_learning_save_path(save_path: str, continue_learning: bool) -> None:
...
+     torch_deterministic (bool): If True, enforces PyTorch deterministic algorithms.
+                           May reduce performance. Default is True.
+
+     Notes:
+         - Completely reproducible results are not guaranteed across different PyTorch versions, hardware, or CUDA configurations.
+         - See https://docs.pytorch.org/docs/stable/notes/randomness.html
+    """
+    random.seed(seed)
+    np.random.seed(seed)
+    os.environ["PYTHONHASHSEED"] = str(seed)
```
maurerle: PYTHONHASHSEED does not work when set within the running process:
https://stackoverflow.com/questions/30585108/disable-hash-randomization-from-within-python-program

Nice idea though.

assume/scenario/loader_csv.py
```diff
@@ -492,6 +493,14 @@ def load_config_and_create_forecaster(
...
+    if "seed" not in config:
+        set_random_seed(42)
+    # If seed is provided, set the random seed accordingly
+    elif config.get("seed") is not None:
+        set_random_seed(config["seed"])
+    # 'seed: null' in config -> no seeding
```
maurerle: This can be much simpler:
```suggestion
    set_random_seed(config.get("seed", 42))
    # 'seed: null' in config -> no seeding
```

 This requires adjusting set_random_seed to return directly if seed is None

assume/common/utils.py
mthede: Ah damn, okay :( thanks for the hint. I removed it.

assume/common/utils.py
kim-mskw: I thought that we let DRL be stochastic unless a seed is set. So the default behavior should be non-deterministic, and then it is faster, and if one wants completely reproducible results, they must set a seed, which makes this deterministic.

kim-mskw: Ah well then we would have different default behavior between learning and the rest of the simulation.

assume/scenario/loader_csv.py
kim-mskw: It makes total sense to have it in load config, since when people want to simulate multiple runs, they may use a loop and overwrite the config, including, if wanted, a new seed.

assume/common/utils.py
maurerle: Sounds like we should always stick to random unless a seed is set - in which case we will also set cuda to be deterministic..?

kim-mskw: Disagree as written below added documentation and seed: null in learning configs

assume/common/utils.py
```diff
@@ -843,3 +844,41 @@ def confirm_learning_save_path(save_path: str, continue_learning: bool) -> None:
...
+         - See https://docs.pytorch.org/docs/stable/notes/randomness.html
+    """
+    if seed is None:
+        return
+
+    random.seed(seed)
+    np.random.seed(seed)
+
+    try:
+        import torch as th
```
maurerle, line 867: Importing torch without using it costs us a few startup seconds for non-learning scenarios.
We are doing this elsewhere anyway as well - but just wanted to note this.

It is a different issue though.

kim-mskw:
@mthede I added the test but it does not seem to work properly if i do not have a "Denkfehler", can you double check your function

assume/common/utils.py
mthede: Mhmm, yes, that's a good point. I think I kind of wanted to have a check if it's a learning simulation with torch or not. And then skip the seed setting for torch things. Do you have a good idea how to do this?

mthede: We could pass the learning_mode attribute into the method and skip the torch stuff based on this.

maurerle: Hm, well yes this works here - but the issue also exists in fast_pandas:
https://github.com/assume-framework/assume/blob/8df8d44e1b734d4edf5b875239590722a723a754/assume/common/fast_pandas.py#L13

there is not a nice way to handle this - I think we have to just live with it..

PR #720, opened 2026-01-13
Fix forecast elastic demand

### LINKED ISSUE #719: When using elastic_demand and no demand_df, market price forecast is wrong
We introduced an option for elastic demand in #575 which does not require a demand_df timeseries (because it does not make sense for an elastic demand unit, that formulates bids according to its own willingness to pay). In this case, demand_df is filled with zeros.
However we have overlooked that calculate_market_price_forecast() requires the timeseries for estimation of a market price. With the current implementation (demand is always 0), the price forecast is always close to zero.

Options to calculate a price forecast in presence of elastic demand:
- a. take the max_power of the demand unit as constant demand (overestimating consumption and prices, will for sure be totally off)
- b. take something between min and max as constant demand (might be totally off)
- c. calculate the bids of the demand unit for each timestep and include it in the "clearing" loop in forecast_initialisation lines 305 ff (will be most in line with the price forecast approach implemented for powerplants)
- d. Require the forecast in demand_df (users need to calculate it themselves depending on their use case and provide it to the simulation as an input)

Discussion points:
- Do you have additional ideas?
- Which of the proposed solutions fits best?

gugrimm:
## Related Issue
Closes #719 

## Description
changed the forecast to use PayAsClear when elastic demand is present

## Checklist

not yet working!


assume/common/forecast_initialisation.py
```diff
@@ -299,31 +330,77 @@ def calculate_market_price_forecast(self, market_id) -> pd.Series:
...
+                demand_bids = pd.DataFrame({
+                    'price': elastic_demand_prices,
+                    'volume': elastic_demand_volumes
+                }).reset_index().rename(columns={'index': 'bid_id'})
+                # create an orderbook containing all supply offers and demand bids
+                orderbook = []
+                orderbook.extend(supply_offers.to_dict('records'))
+                orderbook.extend(demand_bids.to_dict('records'))
+                if demand_t > 0:
+                    orderbook.append({'price': 3000.0, 'volume': demand_t})
```
kim-mskw: Why is this necessary? Can you explain and add a comment?

assume/common/forecast_initialisation.py
```diff
@@ -282,6 +287,32 @@ def calculate_market_price_forecast(self, market_id) -> pd.Series:
         ]
         sum_demand = self.demand[demand_units.index].sum(axis=1)
```
kim-mskw, line 297: sun_demand is here zero is that correct as it is?

assume/common/forecast_initialisation.py
kim-mskw: it adds a zero as price of 3000, why? Is that the price for imports and exports? I am confused by sum_demand and demand_t

assume/common/forecast_initialisation.py
```diff
@@ -299,31 +330,77 @@ def calculate_market_price_forecast(self, market_id) -> pd.Series:
...
+
+                # Compute the cumulative sum of available power in the sorted order.
+                # cumsum_power = sorted_power.cumsum()
+                supply_offers = pd.DataFrame({
+                    'price': mc_t,
+                    'volume': sorted_power
+                }).reset_index().rename(columns={'index': 'unit_id'})
+                # get the demand bids
+                demand_t = sum_demand.loc[t]
+                demand_bids = pd.DataFrame({
```
kim-mskw: Why dont you use all_bids here?

assume/common/forecast_initialisation.py
```diff
@@ -299,31 +330,77 @@ def calculate_market_price_forecast(self, market_id) -> pd.Series:
...
+                marketconfig = self.market_configs[market_id]
+                marketconfig_dict = marketconfig.copy() if isinstance(marketconfig, dict) else marketconfig.__dict__
+                marketconfig_dict['opening_hours'] = rr.rrule(
+                    rr.HOURLY,
+                    dtstart=start,
+                    until=end,
+                    cache=True,
+                )
+                # remove 'operator' from marketconfig and marketconfig_dict
+                if 'operator' in marketconfig_dict:
```
kim-mskw: Why? What did the operator do? do you need it initialized then which did not happen yet? If so comment.

gugrimm: yes - I am confused: the MarketConfig throws an error when initializing it with the marketconfig or marketconfig_dict and I did not understand how to initialize it correctly

assume/common/forecast_initialisation.py
gugrimm: if demand_df.csv was not in the inputs, sum_demand and deman_t will always be zero, yes

assume/common/forecast_initialisation.py
gugrimm: the demand_t is the inelastic demand potentially coming from demand_df.csv and it should be considered if present with a very high price. It is appended to the orderbook containing the elastic bids in line 388, if it is > 0

assume/common/forecast_initialisation.py
gugrimm: if there is inelastic demand coming from demand_df, it shoud be added to the orderbook with a high price

assume/strategies/naive_strategies.py
```diff
@@ -387,6 +387,10 @@ def calculate_bids(
         **kwargs,
     ) -> Orderbook:
         bids = []
+        num_bids = int(unit.num_bids)
+        # set node to 'node0' if not defined
+        if not hasattr(unit, "node"):
```
maurerle, line 392: why is this needed? Powerplant and Demand have exactly this default already configured..?

assume/strategies/naive_strategies.py
```diff
@@ -387,6 +387,10 @@ def calculate_bids(
         **kwargs,
     ) -> Orderbook:
         bids = []
+        num_bids = int(unit.num_bids)
```
maurerle, line 390: unit.num_bids already gets converted to int when set, the conversion should not be needed here

assume/common/forecast_initialisation.py
maurerle: The MarketConfig does not have an operator keyword - it only exists when converting the `markets_config` yaml to the actual MarketConfig object in the CSV loader.

We should not add this notation outside of the yaml..?

I somewhat like the idea of using the actual configured clear function, I wonder how that performs though.

assume/strategies/naive_strategies.py
gugrimm: This is needed if I call the strategy from the forecaster, as at this time of the run, no units and or strategies are initialized.

assume/strategies/naive_strategies.py
gugrimm: same here

kim-mskw:
Waiting for the new forecaster architecture.

reinecfi:
@gugrimm this should be fixed with the new forecaster interface right? Then we could close this PR

PR #462, opened 2024-10-28
Algorithmic Expansion to PPO

kim-mskw:
# Pull Request

## Related Issue
Closes #239 

## Description
This pull request introduces key improvements to the PPO implementation, specifically in buffering, action handling, and role-specific configurations. 
Original paper: https://arxiv.org/abs/1707.06347
Based on implementation in https://github.com/marlbenchmark/on-policy/blob/main/onpolicy/algorithms/r_mappo/algorithm/r_actor_critic.py
PPO-Clip: https://stable-baselines.readthedocs.io/en/master/modules/ppo1.html
Paper that shows handling of non-stationartiy: 	https://arxiv.org/abs/2207.05742

Below is an overview of the changes by file:

## Changes Proposed
- **Buffer.py**  
  - Introduced a dynamic buffer for PPO to minimize dependency on varied configurations.
    - `add()`: Automatically expands buffer size when called, triggering `expand_buffer()`.
    - Integrated a warning when memory demand is expected to exceed available space.
    - `reset()`: Fully resets the buffer.
    - `get()`: Returns all necessary data for policy updates.
    - **Note:** For reverting to a static buffer, pre-configured methods in `learning_utils.py` calculate required buffer sizes and are currently commented out for reference.

- **PPO Mini-Batches:** 
  - Implementation of mini-batches for optimization.
    - **Advantages of Mini-Batches:** Reduced memory and computational requirements during optimization.
    - **Resources for Reference:**
      - [PPO with Mini-Batches](https://stable-baselines.readthedocs.io/en/master/_modules/stable_baselines/ppo1/pposgd_simple.html#PPO1)
      - [PPO without Mini-Batches](https://github.com/nikhilbarhate99/PPO-PyTorch/blob/master/PPO.py)

- **Algorithm-specific Files: `matd3.py` and `ppo.py`**  
  - Implemented `get_actions()` method:
    - Initialized in `LearningStrategy` within `base.py` and assigned accordingly.
    - Methods in each file return distinct values:
      - **MATD3:** returns `action & noise`.
      - **PPO:** returns `action & log probability`.

- **learning_strategies.py**  
  - `calculate_bids()`:
    - Processes the second return value of `get_actions` as `extra_info` based on its shape.
    - Adds this value to `unit.outputs`.

- **World.py**  
  - `add_rl_units_operator()`:
    - Issues a warning if `train_freq` exceeds the `end - start` time set in config.

- **Loader_csv.py**  
  - `run_learning()`:
[...]

PR #721, opened 2026-01-14
MAPPO Implementation

Harshul-18:
## Description
Implemented PPO algorithm in assume framework.

## Checklist
- [ ] Implement PPO specific Actor and Critic Classes in `neural_network_architecture.py`.
- [ ] Add PPO specific Buffer class in `buffer.py`.
- [ ] Update the existing `learning_role.py` to work with PPO.
- [ ] Configured the input data pipeline to be compatible with PPO.
- [ ] Implement the MAPPO algorithm in `mappo.py`

## Additional Comments
- The results of the already implemented algorithm might be different from previously benchmarked results due to change in common hyper-parameters like learning_rate, episodes, noise, constants, etc.

## Results

| Scenario | PPO | TD3 | DDPG |
|--------|-----|-----|------|
| **example_02a** |  |  |  |
| **example_02b** |  |  |  |
| **example_02c** |  |  |  |
| **example_02d** |  |  |  |
| **example_02e** |  |  |  |
| **custom_scenario** |  |  |  |


assume/reinforcement_learning/algorithms/maddpg.py
```diff
@@ -0,0 +1,614 @@
...
+    # =========================================================================
+    # CHECKPOINT SAVING METHODS
+    # =========================================================================
+
+    def save_params(self, directory: str) -> None:
+        """Save all actor and critic network parameters to disk."""
+        self.save_critic_params(directory=f"{directory}/critics")
+        self.save_actor_params(directory=f"{directory}/actors")
+
+    def save_critic_params(self, directory: str) -> None:
```
kim-mskw: This code is completly redundant, I see why it is not in the BAseAlgoithm class because not all hav actors and critics, does it makes sense to have another parent class though? Please check

assume/reinforcement_learning/algorithms/maddpg.py
```diff
@@ -0,0 +1,614 @@
...
+            "unique_obs_dim": self.unique_obs_dim,
+        }
+
+    # =========================================================================
+    # CORE TRAINING: POLICY UPDATE
+    # =========================================================================
+
+    def update_policy(self) -> None:
+        """
+        Update actor and critic networks using the DDPG algorithm.
```
kim-mskw: make self explainatory dokstring that does not depend on dokstring of TD3

assume/reinforcement_learning/buffer.py
```diff
@@ -172,3 +172,255 @@ def sample(self, batch_size: int) -> ReplayBufferSamples:
...
+    def add(
+        self,
+        obs: np.ndarray,
+        action: np.ndarray,
+        reward: np.ndarray,
+        done: np.ndarray,
+        value: np.ndarray,
+        log_prob: np.ndarray
+    ) -> None:
+        """Add a transition to the buffer."""
```
kim-mskw: not a proper dokstring

assume/reinforcement_learning/buffer.py
```diff
@@ -172,3 +172,255 @@ def sample(self, batch_size: int) -> ReplayBufferSamples:
...
+        """Generate batches of samples for training."""
+        if not self.generator_ready:
+            raise ValueError(
+                "Must call compute_returns_and_advantages before sampling."
+            )
+        
+        buffer_size = self.pos if not self.full else self.buffer_size
+        indices = np.random.permutation(buffer_size)
+
+        if batch_size is None:
```
kim-mskw, line 425: is that a sensible default behavior? hwere does it come from? When I have no batch size it uses the entire buffer?

assume/reinforcement_learning/buffer.py
```diff
@@ -172,3 +172,255 @@ def sample(self, batch_size: int) -> ReplayBufferSamples:
...
+            batch_size = buffer_size
+
+        start_idx = 0
+        while start_idx < buffer_size:
+            batch_indices = indices[start_idx : start_idx + batch_size]
+            yield self._get_samples(batch_indices)
+            start_idx += batch_size
+
+    def _get_samples(self, indices: np.ndarray) -> RolloutBufferSamples:
+        """Convert numpy arrays to torch tensors for given indices."""
```
kim-mskw: Also not a porper dokstring. In the original implementation of the PPO the dokstring are good, use these! https://github.com/assume-framework/assume/pull/462/files#diff-e7cf9fcee75b300570d21e1894f6aa672a59d0faa37d8afdcf9611d0ff28fab7R498

assume/reinforcement_learning/learning_role.py
```diff
@@ -121,6 +128,10 @@ def __init__(
             self.all_rewards = defaultdict(lambda: defaultdict(list))
             self.all_regrets = defaultdict(lambda: defaultdict(list))
             self.all_profits = defaultdict(lambda: defaultdict(list))
+            # PPO algorithm specific caches for on-policy learning
```
kim-mskw: The learning_role knw the algorithm, can we do these things algorithms specific?

assume/reinforcement_learning/learning_role.py
```diff
@@ -290,12 +309,84 @@ async def _store_to_buffer_and_update_sync(self, cache, device) -> None:
...
+                    device
+                )
+                rewards_data = transform_buffer_data(
+                    {
+                        timestamp: cache["rewards"][timestamp]
+                    },
+                    device
+                )
+                
+                if cache["values"].get(timestamp):
```
kim-mskw: why the if here? Is there an option that I do not have values in the cache and if so why?

assume/reinforcement_learning/learning_role.py
```diff
@@ -290,12 +309,84 @@ async def _store_to_buffer_and_update_sync(self, cache, device) -> None:
...
+                # Helper to convert to numpy
+                def to_numpy(data):
+                    if isinstance(data, th.Tensor):
+                        return data.cpu().numpy()
+                    return np.array(data)
+
+                # Add to rollout buffer
+                if self.rollout_buffer is not None:
+                    self.rollout_buffer.add(
+                        obs = to_numpy(obs_data),
```
kim-mskw: why is this necessary here and not for the replay bufer we already convert the stuff into numpy at the end of transform_buffer_data

assume/reinforcement_learning/learning_role.py
```diff
@@ -362,6 +474,7 @@ def load_inter_episodic_data(self, inter_episodic_data):
         self.rl_eval = inter_episodic_data["all_eval"]
         self.avg_rewards = inter_episodic_data["avg_all_eval"]
         self.buffer = inter_episodic_data["buffer"]
+        self.rollout_buffer = inter_episodic_data["rollout_buffer"]
```
kim-mskw: only load one buffer, get rid of redunandcy and make this algorithm specific

assume/reinforcement_learning/neural_network_architecture.py
```diff
@@ -26,39 +30,41 @@ def __init__(
...
         """
         layers = nn.ModuleList()
         input_dim = (
             self.obs_dim + self.act_dim
-        )  # Input includes all observations and actions
+        ) # Input includes all observations and actions
 
-        for h in hidden_sizes:
-            layers.append(nn.Linear(input_dim, h, dtype=float_type))
+        for h in self.hidden_sizes:
```
kim-mskw, line 65: This is not correct MATD3 and MAPPO use different information in the critic network. Check BenchMARL implementation. The MAPPO does not have action values in ciritc input space. So you in fact cannot use this in all contexts.

assume/reinforcement_learning/neural_network_architecture.py
```diff
@@ -247,18 +381,193 @@ def forward(self, obs):
...
         x = self.activation_function(self.FC2(x))
 
         if not is_batched:
             x = x.squeeze(0)
 
         return x
+
+
+class ActorPPO(nn.Module):
+    activation_function_limit = {
```
kim-mskw: This is used across many classes and redundant. I would suggest moving it to the learning utils.

assume/reinforcement_learning/neural_network_architecture.py
```diff
@@ -247,18 +381,193 @@ def forward(self, obs):
...
+        mean: th.Tensor,
+        std: th.Tensor,
+    ) -> th.Tensor:
+        """Compute log probability of actions under Gaussian distribution."""
+        var = std.pow(2)
+        log_prob = -0.5 * (
+            ((actions - mean).pow(2) / var)
+            + 2 * th.log(std)
+            + th.log(th.tensor(2 * th.pi))
+        )
```
kim-mskw, line 574: suggested a change limit and map need to be adjusted now though

assume/strategies/learning_strategies.py
```diff
@@ -283,23 +284,49 @@ def get_actions(self, next_observation):
...
+                    
+                    # Get value estimate from critic (if available)
+                    if hasattr(self.learning_role, 'critics') and self.unit_id in self.learning_role.critics:
+                        critic = self.learning_role.critics[self.unit_id]
+                        self._last_value = critic(next_observation.unsqueeze(0)).squeeze().detach()
+                    else:
+                        self._last_value = th.tensor(0.0, device=self.device)
+                    
+                    # PPO uses stochastic policy, no external noise needed
+                    noise = th.zeros_like(curr_action, dtype=self.float_type)
```
kim-mskw: PPO should not have an initial experience mode or should it? since it is a rollout buffer it will be gone after one update anyhow. Do you know papers with PPO and initial experience? Can we integrate this better? If you look into the original ilementation the get_actions function was moved to the algirthm file and hence can have all the algorithmtic features. which is cleaner!

kim-mskw (changes requested):
Thanks for the initial implementation!

I added multiple comments about code redundancy and handling in the framework. I pushed non-working versions of what I mean so that you get a clear picture of how it could be improved. Many of these thoughts have already been made in https://github.com/assume-framework/assume/pull/462/files#diff-d9acf7632f3702baead73bae50128bce95b87863f5575f2731c87efbeacbc508, which is a working version of the PPO on an old branch. Please revisit and further revise your implementation.  

Also, PPO is known to be less sample efficient, so before comparing them, please let it train longer.

assume/reinforcement_learning/buffer.py
Harshul-18: If you meant, the reason for if..else.. condition `self.pos if not self.full else self.buffer_size` here, then after a check yes it is redundant and can simply be equated with `self.pos` without the if..else.. condition. I assumed `self.pos` it as a circular buffer pointer as in ReplayBuffer.

If no batch_size is specified, it uses the entire buffer till the point where it is full (determined by the `self.pos`)

[StableBaselines3's implementation](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/buffers.py#L501)

assume/reinforcement_learning/learning_role.py
Harshul-18: I wrote it earlier to check the execution of the other files and the algorithm. But yes, it is not required here and is incorrect here to add zeros or any data explicitly if for some reason the values aren't added. the if..else.. can be removed and can be used in the same manner as rewards_data, actions_data, etc.

assume/strategies/learning_strategies.py
Harshul-18: Yeah, PPO should not have an initial experience mode theoretically based on my knowledge. 

But when I was running the training, the rewards graph was almost all of the time showing the decreasing trend. So I thought to have the initial experience where random actions are performed (which were rewarding in positively mostly), so thought first update might cause it to improve from start, but it did improve slightly though on the custom scenario example. The original idea I had was to always save the K number of samples with top K maximum rewards where K would be less than the buffer size and will be a hyper-parameter, just for the newer updates to be improved keeping the best performances in mind. (Yet to do a research on it, but I think that deviates from the PPO). 

For now, at the use case point, the initial experience mode is could be set to 0.

kim-mskw:
Current performance: 
example_02a results
 
image

example_02b results
 
image

example_02c results
image

assume/reinforcement_learning/learning_role.py
```diff
@@ -336,25 +468,45 @@ def add_actions_to_cache(self, unit_id, start, action, noise) -> None:
...
+            profit: The profit to be added.
         """
         self.all_rewards[start][unit_id].append(reward)
         self.all_regrets[start][unit_id].append(regret)
         self.all_profits[start][unit_id].append(profit)
 
-    def load_inter_episodic_data(self, inter_episodic_data):
-        """
-        Load the inter-episodic data from the dict stored across simulation runs.
+    def add_ppo_data_to_cache(
```
kim-mskw, line 484: Function never used?

assume/common/base.py
```diff
@@ -843,52 +922,48 @@ class LearningConfig:
...
+            
+        self.off_policy.actor_architecture = self.actor_architecture
+        self.on_policy.actor_architecture = self.actor_architecture
+
         if self.early_stopping_steps is None:
             self.early_stopping_steps = int(
                 self.training_episodes / self.validation_episodes_interval + 1
             )
 
-        # if we do not have initial experience collected we will get an error as no samples are available on the
```
kim-mskw, line 878: Where did these checks go? They ar needed.

kim-mskw: I added them above or am I missing something?

assume/reinforcement_learning/algorithms/base_algorithm.py
```diff
@@ -65,27 +97,632 @@ def update_learning_rate(
...
+                action + noise,
+                strategy.actor.min_output,
+                strategy.actor.max_output,
+            )
+            return action, noise
+
+        # Evaluation
+        action = strategy.actor(obs).detach()
+        noise = th.zeros(strategy.act_dim, dtype=strategy.float_type, device=strategy.device)
+        return action, noise
```
kim-mskw: I am not conviced that the general A2C algorithm has off-policy default behavior that seems prone to mistakes. I moved get_actiosn to the respective sub classes.

assume/reinforcement_learning/learning_role.py
```diff
@@ -303,40 +349,139 @@ async def _store_to_buffer_and_update_sync(self, cache, device) -> None:
                 )
                 return
 
-        # rewrite dict so that obs.shape == (n_rl_units, obs_dim) and sorted by keys and store in buffer
-        self.buffer.add(
-            obs=transform_buffer_data(cache["obs"], device, self.rl_strats.keys()),
-            actions=transform_buffer_data(
-                cache["actions"], device, self.rl_strats.keys()
```
kim-mskw, line 310: I do not agree with this implemntation. We tlaked about this and I told you to use the implemenation on the main branch, which never sorts the rl_starts.keys(). they are used as a single source of toruth for unit order.

assume/reinforcement_learning/algorithms/mappo.py
```diff
@@ -0,0 +1,491 @@
...
+        3. Updating the Actor and Critic networks over multiple epochs using mini-batches.
+        4. Calculating the surrogate objective with clipping.
+        5. Calculating value function loss (MSE) and entropy bonus.
+        6. Logging metrics and gradients.
+        7. Clearing the on-policy buffer after the update.
+        """
+        logger.debug("Updating Policy")
+
+        # Keeping strategy order aligned with rollout-buffer column order.
+        sorted_unit_ids = sorted(self.learning_role.rl_strats.keys())
```
kim-mskw: same with sorting as befroe

assume/reinforcement_learning/neural_network_architecture.py
```diff
@@ -26,133 +36,224 @@ def __init__(
...
+        actions: th.Tensor
+    ) -> th.Tensor:
+        """Compute only Q1 (used during actor updates)."""
         x = th.cat([obs, actions], dim=1)
 
-        for layer in self.q1_layers[:-1]:  # All hidden layers
-            x = F.relu(layer(x))
-
-        x = self.q1_layers[-1](x)  # Output layer (no activation)
+        x = nn.Sequential(*self.q1_layers)(x)
```
kim-mskw: I am a bit hestitant about this change, because this PR is huge as it is and this seems to be a style question. I will recert it we can do these chanegs in a seperate PR.

assume/reinforcement_learning/learning_utils.py
```diff
@@ -182,28 +194,42 @@ def get_hidden_sizes(state_dict: dict, prefix: str) -> list[int]:
...
 
     Returns:
-        th.Tensor: Shape (n_timesteps, n_powerplants, feature_dim)
+        Shape (n_timesteps, n_powerplants, feature_dim).
     """
+    if not nested_dict:
+        return np.zeros((0, 0, 1), dtype=np.float32)
+
     # Get sorted lists of units and timestamps (for consistent ordering)
     all_times = sorted(nested_dict.keys())
```
kim-mskw, line 225: This is a super important function. Why did you want to change that? 
We must be using the key order with which they come in here and not sort them at any given place. We talked about taking the approach form the main branch. I commented it multiple times, please explain why you did it that way.

assume/reinforcement_learning/learning_utils.py
```diff
@@ -182,28 +194,42 @@ def get_hidden_sizes(state_dict: dict, prefix: str) -> list[int]:
...
 
     Args:
-        nested_dict: Dict with structure {datetime -> {unit_id -> list[tensor]}}
+        nested_dict: Dict with structure {datetime -> {unit_id -> list[tensor]}}.
+        device: PyTorch device config.
 
     Returns:
-        th.Tensor: Shape (n_timesteps, n_powerplants, feature_dim)
+        Shape (n_timesteps, n_powerplants, feature_dim).
     """
```
kim-mskw, line 224: Why did you suggest this default behavior?

assume/reinforcement_learning/learning_utils.py
```diff
@@ -218,10 +244,10 @@ def transform_buffer_data(
         if feature_dim is not None:
             break
 
+    # Some on-policy fields (e.g. log_probs/values) can be empty for some timesteps.
+    # Keep zeros in that case instead of failing the entire training loop.
     if feature_dim is None:
-        raise ValueError(
-            "Error, while transforming RL data for buffer: No data found to determine feature dimension"
-        )
+        feature_dim = 1
```
kim-mskw: This is sensible for the PPO then but not for all other algorithms, not a robust error handling here.

assume/common/base.py
Harshul-18: Sorry, during the refactoring I missed adding those back in the off policy. Thanks for adding them above there.

assume/reinforcement_learning/algorithms/base_algorithm.py
Harshul-18: Ok, I agree and get it making it more generalised for A2C and implementing only in the algorithm classes.

assume/reinforcement_learning/neural_network_architecture.py
Harshul-18: The CriticPPO initializes it with act_dim=0.

assume/reinforcement_learning/learning_role.py
Harshul-18: Thank you for checking, it is called/used to store values and log probs for PPO within add_action_to_cache.

assume/reinforcement_learning/algorithms/maddpg.py
kim-mskw: this is resolved with a normal base algorithm and a A2C base algorithms

assume/reinforcement_learning/learning_role.py
```diff
@@ -122,16 +135,24 @@ def __init__(
             self.all_rewards = defaultdict(lambda: defaultdict(list))
             self.all_regrets = defaultdict(lambda: defaultdict(list))
             self.all_profits = defaultdict(lambda: defaultdict(list))
+            # On-policy (PPO/MAPPO) only: value estimates, log-probs, and done
+            # flags collected per time-step for GAE computation.
+            if is_on_policy(self.learning_config.algorithm):
```
kim-mskw: I would like to have this in algorithm-specific sub-docs where the algorithm specifies the keys it needs, I think

assume/reinforcement_learning/learning_role.py
kim-mskw: was changed now always rl_starts used apparently

assume/reinforcement_learning/learning_utils.py
kim-mskw: now outsourced as new funktion convert_to_tensor (former to_torch) in utils instead of weirdly being placed in buffer

assume/reinforcement_learning/algorithms/mappo.py
```diff
@@ -0,0 +1,521 @@
...
+                float_type=self.float_type,
+            ).to(self.device)
+
+            # Create optimizer
+            strategy.critics.optimizer = AdamW(
+                strategy.critics.parameters(),
+                lr=self.learning_role.calc_lr_from_progress(1),
+            )
+
+    def extract_policy(self) -> dict:
```
kim-mskw, line 378: this seem very general why not in base class with A2C?

assume/reinforcement_learning/algorithms/base_algorithm.py
```diff
@@ -65,27 +99,606 @@ def update_learning_rate(
...
+
+        Args:
+            directory: Path to the directory containing saved parameters.
+
+        Note:
+            This is an abstract method that must be overridden by subclasses.
+        """
+
+
+class A2CAlgorithm(RLAlgorithm):
```
kim-mskw: Rename to ActorCritic Algorithm, because a2C is specific to advantage calculation is tehchnically not completly correct, as MATD3 does not calculate advantage

assume/reinforcement_learning/learning_role.py
kim-mskw: decision: store keys to handel here in dem respecitve algorithm

assume/reinforcement_learning/learning_role.py
```diff
@@ -352,26 +554,69 @@ def add_actions_to_cache(self, unit_id, start, action, noise) -> None:
...
+            profit: The profit to be added.
         """
         self.all_rewards[start][unit_id].append(reward)
         self.all_regrets[start][unit_id].append(regret)
         self.all_profits[start][unit_id].append(profit)
 
-    def load_inter_episodic_data(self, inter_episodic_data):
-        """
-        Load the inter-episodic data from the dict stored across simulation runs.
+    def add_ppo_data_to_cache(
```
kim-mskw: kann add data to chache algorithm specific sien?

assume/reinforcement_learning/learning_role.py
```diff
@@ -383,22 +628,24 @@ def load_inter_episodic_data(self, inter_episodic_data):
         self.initialize_policy(inter_episodic_data["actors_and_critics"])
 
         # Disable initial exploration if initial experience collection is complete
-        if (
-            self.episodes_done
-            >= self.learning_config.episodes_collecting_initial_experience
-        ):
-            self.turn_off_initial_exploration()
+        # Only for off-policy algorithms
+        if is_off_policy(self.learning_config.algorithm):
```
kim-mskw: im base algorithm mit non überschreiben und dann nur im matd3 oder allem was off policy ist die klasse überschreiben

assume/reinforcement_learning/learning_role.py
```diff
@@ -429,60 +675,75 @@ def turn_off_initial_exploration(self, loaded_only=False) -> None:
...
+                self.learning_config.training_episodes - initial_experience_episodes
+            )
 
-        if (
-            self.episodes_done
-            < self.learning_config.episodes_collecting_initial_experience
-        ):
-            progress_remaining = 1
+            if self.episodes_done < initial_experience_episodes:
+                progress_remaining = 1
```
kim-mskw: initila expereince sollte handelbar sien mit dem 0 aus der base klasse wenn es keine gibt

assume/reinforcement_learning/learning_role.py
```diff
@@ -681,41 +946,65 @@ def write_rl_params_to_output(self, cache):
...
             )
-            * int(
-                (timestamp2datetime(self.end) - timestamp2datetime(self.start))
-                / pd.Timedelta(self.learning_config.train_freq)
+        else:
+            # For on-policy, no gradient steps concept - use 0 for calculation purposes
+            gradient_steps_done = 0
+
+        # Handle different parameter structures for on-policy vs off-policy
+        if self.learning_config.algorithm == "mappo":
```
kim-mskw: hier müsste man wriklcih viele daten übergeben... ugly, anders? Berehcnung für dne alg spezifisch auslagern

assume/reinforcement_learning/learning_role.py
```diff
@@ -151,9 +152,46 @@ def on_ready(self):
             src="no_wait",
         )
 
+    def initialize_buffer(self, time_step, validation_interval):
+        """Initialize the replay buffer for reinforcement learning training.
+
+        Args:
+            buffer: The replay buffer to be initialized.
+        """
+        if is_off_policy(self.learning_config.algorithm):
```
mthede: Move to RLAlgorithm

assume/reinforcement_learning/learning_role.py
```diff
@@ -284,10 +331,10 @@ async def store_to_buffer_and_update(self) -> None:
             logger.warning("No experience retrieved to store in buffer at update step!")
 
     async def _store_to_buffer_and_update_sync(self, cache, device) -> None:
```
kim-mskw, line 312: hier noch mappo und is on policy abfrage

maurerle:
I did create a code review of the current state, which still has some smaller and larger findings, which should be addressed:

| # | Issue | Where |
|---|-------|-------|
| 1 | `import assume` now hard-requires optional `torch` | `assume/__init__.py:8` |
| 2 | 15 branch-introduced test failures | see [Test results](#test-results) |
| 3 | Pre-existing user configs raise `TypeError` on load | `assume/common/base.py:949-969` |
| 4 | MAPPO produces NaN weights on a size-1 mini-batch | `algorithms/mappo.py:426-429` |

<details>
<summary>
Full Code review
</summary>

# Code review - `with_PPO_and_DDPG` vs `main`

**Repo:** assume · **Branch:** `with_PPO_and_DDPG` (the `update_agent` label in the session snapshot was stale)
**Base:** `main` @ `9343b0fb` · **Diff:** 60 commits, +5492 / −1057 across 31 files
**Reviewed:** 2026-08-24

Adds MADDPG and MAPPO, refactors the learning role around a `buffer_fields` cache, introduces
nested `on_policy` / `off_policy` config, and adds top-level public-API re-exports.

`ruff check assume/ tests/` passes. All modules import (with `torch` present).

> **Rebase note:** local `main` is at `9343b0fb`, several commits behind real upstream
> (`77f76b38`). This diff is against a stale base.

Each finding is marked **CONFIRMED** (traced through the code, or reproduced) or
**PLAUSIBLE** (strongly indicated, not executed end-to-end).

---

## Contents

- [Blockers](#blockers)
- [Test results](#test-results)
- [A. Packaging & configuration compatibility](#a-packaging--configuration-compatibility)
- [B. Algorithms (MAPPO / MADDPG / MATD3 / base)](#b-algorithms-mappo--maddpg--matd3--base)
- [C. Buffer & learning role](#c-buffer--learning-role)
- [D. Networks & strategies](#d-networks--strategies)
- [E. Tests](#e-tests)
- [F. Configs & docs](#f-configs--docs)
- [Checked and clean](#checked-and-clean)
- [Suggested order of work](#suggested-order-of-work)
- [Environment note](#environment-note)

---

## Blockers

Four issues are currently blocking.

| # | Issue | Where |
|---|-------|-------|
| 1 | `import assume` now hard-requires optional `torch` | `assume/__init__.py:8` |
| 2 | 15 branch-introduced test failures | see [Test results](#test-results) |
| 3 | Pre-existing user configs raise `TypeError` on load | `assume/common/base.py:949-969` |
| 4 | MAPPO produces NaN weights on a size-1 mini-batch | `algorithms/mappo.py:426-429` |

---

## Test results

Full suite on this branch: **18 failed, 534 passed, 3 skipped**.

Re-running the same tests in a clean worktree at `main` shows **3 of those 18 are pre-existing**
and unrelated to this branch:

```
tests/test_world_validation.py::test_redispatch_no_dispatch
tests/test_world_validation.py::test_redispatch_too_early
tests/test_integration_cli.py::test_cli_network
```

The remaining **15 are introduced by this branch**, from three root causes:

### T1 — `Learning.all_obs` / `all_rewards` removed, callers not updated · CONFIRMED

8 failures. The `buffer_fields` refactor replaced the per-field cache attributes with a single
`self.cache` dict (`learning_role.py:123`, `_new_cache`). The commit updated only the config
kwargs in `tests/test_learning_role.py`, not the attribute reads.

```
AttributeError: 'Learning' object has no attribute 'all_rewards'. Did you mean: 'avg_rewards'?
AttributeError: 'Learning' object has no attribute 'all_obs'
```

- `tests/test_learning_role.py:150,164,182,244` — all three atomic-swap tests
- `tests/test_rl_strategies.py` — 3 parametrized cases
- `tests/test_portfolio_learning_strategy.py` — 2 tests
- `tests/test_drl_storage_strategy.py:142,246` — reads `learning_role.all_rewards`

The dead atomic-swap tests matter disproportionately: they were the regression guard for the
concurrent cache-swap logic that this very refactor rewrote.

### T2 — `A2CAlgorithm` renamed but the branch's own new test still imports it · CONFIRMED

4 failures. The class is `ActorCriticAlgorithm` (`algorithms/base_algorithm.py`, exported at
`algorithms/__init__.py:6`); `tests/test_public_api.py:29,65,113,208` imports `A2CAlgorithm`.

```
ImportError: cannot import name 'A2CAlgorithm' from 'assume.reinforcement_learning.algorithms'
```

The new public-API test was written against an intermediate naming and never re-synced — so the
actual base class is asserted nowhere. "A2C" is also a different algorithm, not a generic
actor-critic base, so the old name was semantically wrong anyway.

### T3 — `get_actions` arity change · CONFIRMED

3 failures. `get_actions` now returns `(actions, noise, extra_data)`; the mocks in
`tests/test_drl_storage_strategy.py:323` still return 2-tuples.

```
assume/strategies/learning_strategies.py:958: ValueError: not enough values to unpack (expected 3, got 2)
```

Beyond the tests, this is a **silent breaking change for any user-defined strategy** overriding
`get_actions`, and it surfaces as an unpacking error rather than anything diagnostic. It needs a
changelog entry, and ideally a clearer failure.

---

## A. Packaging & configuration compatibility

### A1 — `import assume` makes `torch` and `tensorboard` hard dependencies · CONFIRMED (reproduced)

`assume/__init__.py:8-22`

`from assume.reinforcement_learning import (...)` pulls in `algorithms/__init__.py` →
`matd3`/`mappo` → `torch`, and `learning_role.py` → `tensorboard_logger.py` →
`from torch.utils.tensorboard import SummaryWriter`. Both packages are listed **only** under the
optional `learning` extra (`pyproject.toml:49-54`), and `assume/strategies/__init__.py:98-128`
deliberately wraps learning-strategy imports in `try/except ImportError` to support installs
without them.

Reproduced by blocking `torch` in `sys.meta_path`:

```
branch: FAILS -> ImportError No module named 'torch'
main:   IMPORT OK
```

Consequences:

- `pip install assume-framework` without the extra → the entire non-learning framework is unusable.
- `tests/conftest.py:11` imports `assume.common.base`, so **every test in the repo** fails at
  collection without torch.
- The `require_learning` pytest marker (`pyproject.toml:139`) — whose whole purpose is letting a
  non-learning install deselect these tests — is now useless, because collection dies first.

**Fix:** lazy imports, a `try/except ImportError` guard around the RL re-exports, or a module
`__getattr__`.

### A2 — Existing `learning_config` YAML now raises `TypeError` · CONFIRMED

`assume/common/base.py:949-969`

`episodes_collecting_initial_experience`, `gradient_steps`, `noise_sigma`, `noise_scale`,
`noise_dt`, `action_noise_schedule`, `tau`, `policy_delay`, `target_policy_noise`,
`target_noise_clip`, `replay_buffer_size` moved from `LearningConfig` into `OffPolicyConfig`.
`World.setup` does `LearningConfig(**learning_dict)` (`assume/world.py:226`) — a dataclass with
no `**kwargs`, no key filtering, no migration shim.

```
TypeError: __init__() got an unexpected keyword argument 'episodes_collecting_initial_experience'
```

The migration is only half-done — verified still on the old flat schema:

- `examples/inputs/example_03a/config.yaml:21,23,28`
- `examples/inputs/example_03b/config.yaml:20,22,26`
- `examples/inputs/example_03c/config.yaml:21,23,28`
- `examples/notebooks/04a`, `04b`, `04c`, `09_example_Sim_and_xRL.ipynb:401,407,419`

The `example_02*` configs *were* migrated, so CI does not catch it.

**Fix:** migrate the remaining examples/notebooks, and add a shim in `__post_init__` that maps
legacy top-level keys into the nested dataclasses with a `DeprecationWarning`.

### A3 — Nested `off_policy` / `on_policy` values are unconditionally clobbered · CONFIRMED

`assume/common/base.py:978-985`

```python
for config in [self.off_policy, self.on_policy]:
    if config:
        config.batch_size = self.batch_size
        config.gamma      = self.gamma
        config.train_freq = self.train_freq

self.off_policy.actor_architecture = self.actor_architecture
self.on_policy.actor_architecture  = self.actor_architecture
```

`__post_init__` parses the nested dicts and then immediately overwrites four fields from the
top-level values. A user writing `on_policy: {actor_architecture: lstm}` silently gets MLP
actors, because `mappo.py:166` reads `learning_config.on_policy.actor_architecture` — which
line 985 has just overwritten with the top-level default `"mlp"`. Same for nested
`batch_size` / `gamma` / `train_freq`, all documented as configurable in the
`OffPolicyConfig` / `AlgorithmConfig` docstrings.

**Fix:** copy down only when the nested value is unset (sentinel default), or drop these fields
from the nested dataclasses entirely.

### A4 — `learning_mode: false, evaluation_mode: true` now crashes · CONFIRMED

`learning_role.py:278`, `strategies/learning_strategies.py:93`

`register_strategy` now calls `self.rl_algorithm.setup_strategy_noise(strategy)`, and
`TorchLearningStrategy.get_actions` calls `self.learning_role.rl_algorithm.get_action(...)`.
But `rl_algorithm` is only created under `if self.learning_config.learning_mode:`
(`learning_role.py:104`), while `world.py:318` attaches the learning role on
`learning_mode or evaluation_mode`.

```
AttributeError: 'Learning' object has no attribute 'rl_algorithm'
```

`self.cache` and `self.buffer_fields` are likewise never created. On `main`, `register_strategy`
touched no algorithm, so this combination worked. `run_learning`'s own eval runs keep
`learning_mode: true`, so in-tree paths are safe — this only bites external configs.

### A5 — Unknown `algorithm` name gives `AttributeError` instead of a config error · CONFIRMED

`learning_role.py:106,122,661-668`

`create_learning_algorithm` only `logger.error`s on an unrecognised name and returns without
setting `self.rl_algorithm`; the next statement `self.buffer_fields = self.rl_algorithm.buffer_fields`
raises. New in this diff — the attribute access moved into `__init__`.

Worse, the same swallowed error makes `is_on_policy` and `is_off_policy` **both** `False` for a
typo, which routes an unknown algorithm to a `RolloutBuffer` in `initialize_buffer`
(`learning_role.py:161-188`, `else` branch) but down the `ReplayBuffer` path in
`_store_to_buffer_and_update_sync` (`learning_role.py:359,465`) — two mutually inconsistent
defaults for the same name.

**Fix:** raise `ValueError` on an unknown algorithm.

---

## B. Algorithms (MAPPO / MADDPG / MATD3 / base)

### B1 — Advantage normalisation yields NaN on a mini-batch of size 1 · CONFIRMED (reproduced)

`algorithms/mappo.py:426-429`

```python
advantages_flat = advantages_i.flatten()
advantages_i = (advantages_i - advantages_flat.mean()) / (advantages_flat.std() + 1e-8)
```

`Tensor.std()` is the **unbiased** std, which is `nan` for a single element. Verified in this
repo's torch:

```
>>> th.tensor([0.5]).std()
tensor(nan)
```

Reachable paths:

- `RolloutBuffer.get()` (`buffer.py:398-401`) yields a trailing remainder batch, so any config
  where `pos % effective_batch_size == 1` hits it — e.g. `train_freq: 24h`, hourly steps,
  `batch_size: 11` → batches of 11, 11, **1**.
- Most directly: `rollout_buffer.pos == 2` passes the guard at `mappo.py:283`, is decremented to
  1 at line 324, so `effective_batch_size = 1` and *every* mini-batch is size 1.

`policy_loss` becomes NaN → `loss.backward()` → actor and critic weights are permanently NaN,
with no error raised.

**Fix:** `std(unbiased=False)`, or skip normalisation when `numel() < 2`.

The adjacent comment — "Normalize advantages across the entire batch, not per-mini-batch" — also
contradicts the code: `advantages_i` is already the mini-batch slice, and `flatten()` is a no-op
on a 1-D tensor. Nothing tests it.

### B2 — `dones` is hard-coded to `0.0`, so GAE never terminates an episode · CONFIRMED

`algorithms/mappo.py:137`

```python
extra_data = {"values": 0.0, "log_probs": log_prob, "dones": 0.0}
```

This is the **only** writer of the `dones` buffer column (`learning_role.py:448`). Nothing ever
sets `1.0`. So `self.dones` is a permanently-zero array, `next_non_terminal` is always 1, and GAE
bootstraps straight across every episode boundary. The terminal handling in
`compute_returns_and_advantages` and `last_dones` in mappo are inert.

`values` has the same problem in the other direction: written as `0.0`, then never read — for
`algorithm == "mappo"`, `_store_to_buffer_and_update_sync` (`learning_role.py:404-434`)
recomputes value estimates from the centralized critic and ignores `cache["values"]` entirely.
The `else` branch at `learning_role.py:436-440` that *would* read it is unreachable, since mappo
is the only on-policy entry in `ALGORITHM_CATEGORIES`. The field is carried through cache,
carry-over, and output for nothing.

**Decide:** either produce real terminal flags, or delete the `dones`/`values` columns and the
bootstrap-flag logic. Keeping dead fields that look load-bearing is how B3 stayed hidden.

### B3 — Off-by-one on the bootstrap `done` flag · CONFIRMED (latent, masked by B2)

`algorithms/mappo.py:318-321`, `buffer.py:349-356`

GAE in `compute_returns_and_advantages` uses `1 - self.dones[step+1]` paired with
`self.values[step+1]` — i.e. `dones[t]` is the terminal flag *of state t*. The bootstrap uses
`observations[last_idx]` (correct) but pairs it with `dones[last_idx - 1]` instead of
`dones[last_idx]`.

The two branches in the buffer disagree for the same quantity: for `step < buffer_size-1` the
code uses `self.dones[step+1]`; for the last step it uses the caller-supplied `dones`, and the
caller passes `dones[step]`.

Currently masked only by B2. The moment real terminal flags exist, the final transition of every
rollout gets the wrong bootstrap mask, and the error propagates backwards through the whole GAE
recursion.

### B4 — `continue` in the key-validation loop continues the wrong loop · CONFIRMED

`algorithms/base_algorithm.py:451-456`

```python
for key in required_keys:
    if key not in critic_params:
        logger.warning(f"Missing {key} in critic params for {u_id}; skipping.")
        continue          # <-- continues the *key* loop, not the agent loop

if direct_load:
    strategy.critics.load_state_dict(critic_params["critic"])
    strategy.critics.optimizer.load_state_dict(critic_params["critic_optimizer"])  # KeyError
```

The "skipping" warning is a lie. An older checkpoint lacking `critic_optimizer` logs the warning,
falls through, loads the critic weights anyway, then raises `KeyError` at line 460 — which the
broad `except` at line 510 swallows as "Failed to load critic for X". Net result: a half-loaded
critic reported as a total failure.

**Fix:** collect missing keys, then `continue` the outer agent loop (or `break` out of the inner).

### B5 — MAPPO critic weight transfer can never succeed · CONFIRMED

`algorithms/base_algorithm.py:470-484`, `learning_utils._get_q_prefixes`

`transfer_weights` only recognises the prefixes `q_layers` / `q1_layers` / `q2_layers`, but
`CriticPPO` builds `self.v_layers` (`neural_network_architecture.py:182`). `act_dim=0` for
`CriticPPO` is also not modelled by the slicing logic.

`continue_learning` for `algorithm: mappo` with any change in the agent set or order →
`direct_load` is False → `transfer_weights` returns `None` → "Critic weights transfer failed;
skipping" for every agent → training silently restarts from randomly initialised centralized
critics, while the actors *are* loaded (`load_actor_params` has no such gate). The mismatch
between a pretrained actor and a random critic is worse than a clean restart.

### B6 — Critic / target-critic transfer is no longer atomic · CONFIRMED (regression vs main)

`algorithms/base_algorithm.py:486-506`

`main`'s matd3 computed both `critic_weights` and `target_critic_weights` and skipped the agent if
*either* was `None`. The refactor loads `strategy.critics` first (line 486) and only then attempts
the target transfer; if that returns `None` it `continue`s, leaving the online critic updated and
the target critic random.

TD3/DDPG `continue_learning` with a changed agent set → `polyak_update` starts from a random
target critic while the online critic is pretrained → the first hundreds of gradient steps
regress toward garbage targets.

### B7 — `th.load` without `map_location` on the critic path · CONFIRMED (inconsistency)

`algorithms/base_algorithm.py:444`

`load_obj` (line 246) passes `map_location=self.device`; the critic path does not. A checkpoint
trained with `device: cuda` loaded on a CPU-only machine raises inside the `try`, is swallowed at
line 510, and training continues with randomly-initialised critics plus successfully-loaded
actors — a silent, hard-to-diagnose restart.

### B8 — `clip_range_vf` is never read from config, so value clipping is dead · CONFIRMED

`algorithms/mappo.py:87`

`self.clip_range_vf = clip_range_vf` (constructor default `None`), unlike every other
hyperparameter, which falls back to `on_policy_config`. `learning_role.create_learning_algorithm`
constructs `PPO(learning_role=self)` with no kwargs, so `mappo.py:447-456` — and the `old_values`
plumbing through `RolloutBufferSamples` — is unreachable in any real run.

### B9 — Early returns leave the rollout buffer un-reset · PLAUSIBLE

`algorithms/mappo.py:277-287`

Both guards (`pos == 0`, `pos < 2`) return before `rollout_buffer.reset()` at line 544. A
train_freq window in which most timestamps are skipped for missing units
(`learning_role.py:363-384`) leaves a stale transition behind; the next window appends
`buffer_size` more and `RolloutBuffer.add` raises `OverflowError` (`buffer.py:301-312`), killing
the simulation. Stale data from the previous window would also be trained on. See also C2.

### B10 — Target actions hard-clamped to `(-1, 1)` · PLAUSIBLE

`algorithms/matd3.py:222`, `algorithms/maddpg.py:205`

`get_action` clamps with `strategy.actor.min_output` / `max_output` (`matd3.py:124-125`), but the
target-policy path hard-codes `(-1, 1)`. `Actor.activation` is `"softsign"` today so they
coincide; switching to `"sigmoid"` or `"relu"` — explicitly offered in the comment at
`neural_network_architecture.py:222` — makes target actions silently inconsistent with behaviour
actions, biasing every TD3/DDPG Q-target.

### B11 — `max(p.grad.norm() ...)` assumes every parameter got a gradient · PLAUSIBLE

`algorithms/matd3.py:320`, `algorithms/maddpg.py:289`

Any parameter not reached by the accumulated loss has `p.grad is None` → `AttributeError`
mid-update. PPO guards this with `if p.grad is not None` and `default=0.0`
(`mappo.py:471-486`); the off-policy paths do not.

Also `matd3.py:251` (`actor = strategy.actor`) is dead — unused inside the critic loop.

### B12 — Target actor silently left random when `actor_target` is absent · PLAUSIBLE

`algorithms/base_algorithm.py:543`

`if self.uses_target_networks and "actor_target" in actor_params` swallows the missing key, where
previously a `KeyError` aborted the whole agent load. Loading a PPO-saved actor checkpoint into a
TD3/DDPG run gives `actor.loaded = True`, `turn_off_initial_exploration(loaded_only=True)`
disables exploration, and the TD3 target actor used at `matd3.py:221` is uninitialised.

---

## C. Buffer & learning role

### C1 — On-policy `get_progress_remaining` is dimensionally wrong · CONFIRMED

`learning_role.py:643-647`

The off-policy branch (`:632-639`) scales the intra-episode term by `1/learning_episodes`; the
on-policy branch adds the raw `elapsed_duration / total_duration`, which sweeps 0→1 *within a
single episode*. So `progress_remaining` runs 1→0 during episode 1 and goes negative in every
later episode.

`linear_schedule_func` (`learning_utils.py:157`) returns `end` (= 0) as soon as
`1 - progress_remaining > end_fraction`. With `learning_rate_schedule: linear` + `mappo`, the LR
set in `mappo.update_policy` becomes exactly **0** in the last window of every episode, and the
intended global decay is replaced by a per-episode sawtooth.

### C2 — RolloutBuffer sized to exactly one window; overflow is a hard crash · PLAUSIBLE (high)

`buffer.py:301-312`, `learning_role.py:177`

`rollout_buffer_size = max(2, int(train_freq / time_step))` equals the number of market steps per
nominal window. But `store_to_buffer_and_update` (`learning_role.py:310-320`) carries incomplete
timesteps into the *next* window, so a window's complete-timestep count is
`nominal + carried_prev − carried_now`.

Whenever the reward lag shortens — the last window of an episode, when trailing steps finally get
rewards, or any unit reporting a reward late once — that exceeds `buffer_size` and `add()` raises
`OverflowError` from inside a mango-scheduled async task, aborting the simulation rather than
degrading. `int()` truncation makes it worse when `time_step` does not divide `train_freq`
(see F1). Compounded by C3 and B9.

### C3 — The RolloutBuffer is never reset at an episode boundary · CONFIRMED

`learning_role.py:560,591` (`get_inter_episodic_data` / `load_inter_episodic_data`)

The buffer object is created once and passed across episodes via `inter_episodic_data["buffer"]`.
The only `reset()` is at the end of `PPO.update_policy` (`mappo.py:544`), skipped by both early
returns (B9). So leftover transitions from episode N are still present when episode N+1 starts
writing, and GAE bootstraps straight across the boundary — nothing cuts the chain, since `dones`
are all zero (B2). It also eats capacity, feeding C2.

### C4 — `continue_learning` no longer disables initial exploration for off-policy · CONFIRMED (regression)

`learning_role.py:566-576`

```python
if is_off_policy(self.learning_config.algorithm):
    if self.episodes_done >= self.learning_config.off_policy.episodes_collecting_initial_experience:
        self.turn_off_initial_exploration()
# For on-policy algorithms, no initial exploration to disable
elif self.learning_config.continue_learning:
    self.turn_off_initial_exploration(loaded_only=True)
```

`main` had a flat `if episodes_done >= … / elif continue_learning:`. The new nesting makes the
`elif` reachable **only for on-policy algorithms** — which have no initial-exploration phase at
all (`PPO` never sets `collect_initial_experience_mode`).

Failure: resume a MATD3 run with `continue_learning: true` and
`episodes_done < episodes_collecting_initial_experience` → the loaded actors are ignored and the
agents bid pure noise for the first N episodes. Exactly the case the flag exists for.

### C5 — `RolloutBuffer` hardcodes `np.float32` and ignores its own `float_type` · PLAUSIBLE

`buffer.py:252,257-268`

`ReplayBuffer` derives `np_float_type` from `float_type` (`:75-77`); `RolloutBuffer.reset` uses
`np.float32` unconditionally while `sample()` converts with `dtype=self.float_type`. With the
documented-but-unimplemented `float16` option this stores at float32 and down-casts per batch —
and the two buffers now disagree about what `float_type` means.

### C6 — `transform_buffer_data` drops values and ignores `float_type` · PLAUSIBLE

`learning_utils.py:247-257`

`result = th.zeros(...)` uses the global default dtype regardless of `learning_role.float_type`,
so a float64 observation tensor is truncated with no warning. And `result[t, u] = values[0]`
silently ignores `values[1:]`.

That interacts badly with `add_actions_to_cache` (`learning_role.py:522-533`), which appends
`action`/`noise` unconditionally but skips `extra_data` fields when already present — so a double
call for one unit/timestep leaves `actions` with two entries and `log_probs` with one. The `[0]`
indexing hides that desync rather than detecting it.

### C7 — `added_timestamps` computed and never used · CONFIRMED (minor)

`learning_role.py:363,463`

Dead counter in the on-policy branch, presumably intended to guard the "everything got skipped"
case: the `missing_units` `continue` at `:384` can silently drop *every* timestep, after which
`update_policy` runs on an empty or partial buffer and only logs at debug level.

### C8 — `ReplayBuffer.add` wrap-around off-by-one · pre-existing, not a regression

`buffer.py:142-145`

`self.pos += len_obs` then `if self.pos + len_obs >= self.buffer_size: self.pos = 0` discards up
to `len_obs - 1` slots on each wrap, and `sample()` still allows `batch_inds + 1` to cross the
wrap point, pairing a transition with a next-observation from a different point in time. Noted
only because the file is in scope — it is unchanged by this diff.

---

## D. Networks & strategies

### D1 — `CriticTD3` lost its weight initialisation · CONFIRMED

`neural_network_architecture.py:89-97`

`CriticTD3.__init__` builds `q1_layers`/`q2_layers` and returns without calling
`self._init_weights()`. On `main`, the old `__init__` ended with that call. `CriticDDPG` (`:157`)
and `CriticPPO` (`:195`) both still call it.

TD3 — the **default** algorithm — is now the only one using PyTorch's default Kaiming-uniform
init with non-zero biases instead of Xavier-uniform with zero bias. A silent training-behaviour
regression for every existing MATD3 run, with no error and no test to catch it.

### D2 — `nn.ReLU()` in the Q-network `ModuleList` silently breaks `transfer_weights` · CONFIRMED

`neural_network_architecture.py:67`

`_build_q_network` now interleaves `nn.ReLU()` modules, so `state_dict` keys become
`q1_layers.0.*`, `q1_layers.2.*`, `q1_layers.4.*` — indices 1 and 3 hold no parameters.

`learning_utils.get_hidden_sizes` (`:186-194`) walks `i = 0,1,2,…` and stops at the first missing
`.{i}.weight`, so it returns `[256][:-1] == []`. In `transfer_weights` (`learning_utils.py:298-365`,
called from `base_algorithm.py:470,490`):

- the architecture-mismatch guard `loaded_hidden != new_hidden` becomes `[] != []` and never
  fires, so a 2-layer checkpoint is happily "transferred" into a 4-layer critic;
- the deeper-layer copy loop `for i in range(1, len(new_hidden) + 1)` becomes `range(1, 1)` — empty.

Net effect: `continue_learning` and pretrained-critic loading copy only the **first** layer's
weights and leave every remaining layer randomly initialised, while reporting success. Forward
passes are unaffected (`F.relu(layer(x))` over a `ReLU` module is idempotent), so nothing else
signals the problem.

### D3 — PPO value-head gain contradicts its own docstring · PLAUSIBLE

`neural_network_architecture.py:197-202`

The docstring says "sqrt(2) gain for hidden layers, **1.0 for the value head**"; the code uses
`gain = 0.01 if layer.out_features == 1 else np.sqrt(2)`. 0.01 is the standard *policy*-head gain;
the conventional PPO value head uses 1.0. With gain 0.01 the critic starts near-zero and its early
advantage estimates are essentially uninformative, slowing and destabilising MAPPO value learning.

`out_features == 1` is also a fragile way to identify the head — a width-1 hidden layer would match.

### D4 — Log-prob is computed on the *clamped* action · PLAUSIBLE

`neural_network_architecture.py:411-441`, `:619-637`

`get_action_and_log_prob` clamps to `[-1, 1]` and then evaluates
`Normal(mean, std).log_prob(clamped_action)` — the density of the unclamped Gaussian evaluated at
a boundary point, not the density of the clamped (mixed continuous/atomic) distribution.

It stays self-consistent with `evaluate_actions` (which also uses the stored clamped action), so
ratios are not obviously broken. But with `log_std_init = 0.0` (std = 1.0 against a `tanh`-bounded
mean in `[-1,1]`), clamping is the *common* case: a large mass of actions piles up at ±1 with
mis-stated log-probs, and both the PPO ratio and the constant `entropy` term then misrepresent
the actual policy.

**Consider:** a proper squashing correction, or a much smaller `log_std_init`.

### D5 — `ActorPPO` never sets `self.activation_function` · PLAUSIBLE (latent)

`neural_network_architecture.py:340-382`

`ActorPPO.__init__` sets `min_output`/`max_output` but — unlike `LSTMActorPPO` (`:506`) and the
`Actor` base class (`:227`) — never assigns `self.activation_function`. Any user subclass
mirroring `MLPActor.forward` and reaching for `strategy.actor.activation_function` gets an
`AttributeError` under MAPPO. No in-tree caller today.

### D6 — Divergent behaviour between `ActorPPO` and `LSTMActorPPO` · minor

`neural_network_architecture.py:443-473`

`LSTMActorPPO` adds `+1e-6` to `std` in `forward` / `get_action_and_log_prob` / `evaluate_actions`;
`ActorPPO` does not — so the two architectures evaluate slightly different distributions for
identical parameters. Separately, `ActorPPO.evaluate_actions`' docstring promises a 3-tuple
`(log_prob, entropy, values)` but returns 2 (`mappo.py:431` correctly unpacks two).

### D7 — `LSTMActor.FC1` hardcodes `+ 2` · pre-existing, not from this diff

`LSTMActor.FC1` hardcodes `+ 2` where `LSTMActorPPO` correctly uses `+ unique_obs_dim` —
`LSTMActor` will shape-error for any strategy with `unique_obs_dim != 2`.

---

## E. Tests

### E1 — No test executes `update_policy` for any algorithm · CONFIRMED

This is the single biggest coverage gap, and it is why B1 and B2 survived to review.

`grep -rn update_policy tests/` returns only comments and a stub. In `tests/test_mappo.py`, the
helpers `_make_rollout_buffer` (`:113`) and `_setup_for_update` (`:144`) are written "for
update_policy tests" and are **never called** — dead code. The one integration-ish test
monkey-patches the method away:

```python
tests/test_mappo.py:341:  learn.rl_algorithm.update_policy = lambda: None
```

So the ~270-line PPO update — clipped surrogate, GAE bootstrap with the `rollout_buffer.pos -= 1`
sacrifice trick at `mappo.py:322`, centralized-critic assembly, buffer reset — and the entire
MADDPG update path have **zero** coverage.

### E2 — Tautological GAE assertions · PLAUSIBLE

`tests/test_rl_rolloutbuffer.py:376` (`test_gae_returns_equal_advantages_plus_values`) and the
same check inside `test_full_episode_rollout` (~`:490`) assert `returns == advantages + values`,
which is literally the assignment at `buffer.py:372`. They can never fail, and give false
confidence about GAE coverage.

The hand-computed GAE tests at `:156-305` are genuinely good — the strongest part of the new suite.

### E3 — `test_maddpg_load_corrupted_critic` mocks away what it claims to test · PLAUSIBLE

`tests/test_maddpg.py:441`

It corrupts only the *target* critic tensors, then asserts the *main* critic equals the original —
never asserting anything about `target_critics` after load. It passes whether the loader repairs,
ignores, or silently leaves the target inconsistent. Given B6, that is precisely the behaviour
that needs asserting.

### E4 — Mini-batch tests only count rows · PLAUSIBLE

`tests/test_rl_rolloutbuffer.py:425`

Sums `observations.shape[0] == T` but never checks that the shuffled indices form a partition (no
duplicates or omissions), nor that obs/action/advantage rows stay aligned to the same index after
permutation — the exact failure mode that would silently poison PPO.

### E5 — Rollout-buffer overflow is reachable and untested · PLAUSIBLE

No test covers the exact-fit case behind C2. `tests/test_rl_rolloutbuffer.py` only exercises a
standalone buffer, and `tests/test_mappo.py:307` deliberately over-sizes (`buffer_size=4` for one
step; `_make_rollout_buffer` used `n_steps + 10`).

---

## F. Configs & docs

### F1 — `example_02a` breaks train_freq alignment · CONFIRMED

`examples/inputs/example_02a/config.yaml:7` — `end_date: 2019-03-30`

`sync_train_freq_with_simulation_horizon` (`learning_role.py:220`) truncates with `int(...)`:
696h / 7 intervals → **99h**, leaving a 3h remainder, so the last training window is short. On
`main` (30 days = 720h → 8 × 90h) it divided exactly. `02b` (28 d → 7 × 96h) is still exact.
Shortening `02a` to 29 days is what triggers it. For MAPPO this also desyncs
`rollout_buffer_size = train_freq / time_step` (`learning_role.py:177`).

### F2 — Unexplained behaviour changes in the `02*` configs · PLAUSIBLE

All five are structurally consistent, and every key they set is actually read
(`off_policy.*` → `learning_role.py:163`, `matd3.py:68-89,170-444`, `maddpg.py:169-381`;
`on_policy.*` → `mappo.py:84-93`, `learning_role.py:186`). But these changes are not explained by
the nested-config refactor:

- `02a` base: `episodes_collecting_initial_experience` 5 → 3
- `02a` base_lstm + tiny gain `action_noise_schedule: linear` (default is `None`, so this newly
  enables noise decay)
- all five drop `replay_buffer_size` from the 50000 default to 10000
- `02b` base: `training_episodes` 100 → 150 (longer example runs)
- `02b` base_lstm: 100 → 30 episodes **and** horizon 31 days → 6 days (`2019-03-07`) — a
  substantive change to a published example's results

### F3 — Docs contradict the implementation · CONFIRMED

- `docs/source/learning_algorithm.rst:264` — "implemented as a fixed-size **circular** buffer".
  It is not: `buffer.py:301-312` raises `OverflowError` when full and explicitly refuses to wrap
  or drop.
- `docs/source/learning_algorithm.rst:~288` — "Once the buffer accumulates enough data (specified
  by `batch_size`), the PPO algorithm's update function is triggered". Wrong: the update is
  triggered by the `train_freq` recurrent task (`learning_role.py:137-153`, `:483`); `batch_size`
  only sizes mini-batches (`mappo.py:378`).
- `docs/source/learning.rst:~107` — "In MAPPO, a **single** centralized value network is used
  across all agents". Wrong: each strategy gets its own critic (`strategy.critics`, iterated
  per-agent at `mappo.py:335` and `:445`).

### F4 — Cosmetic · minor

- Every `02*` config now carries an `on_policy:` block with `n_epochs: 25` while
  `algorithm: matd3`. Inert, and the comment "required for PPO/MAPPO algorithms" is inaccurate —
  these are optional.
- `docs/source/assume.reinforcement_learning.rst` still omits `neural_network_architecture` and
  `tensorboard_logger` (pre-existing).
- `learning_algorithm.rst:206`'s new `.. _buffer:` label is never referenced.

---

## Checked and clean

Verified as *not* defects, recorded so they are not re-reviewed:

- `assume/common/utils.py` `convert_to_tensors` — correct; the `ImportError` fallback is fine.
- `loader_csv.py:1316-1330` reordering — `sync_train_freq_with_simulation_horizon()` now correctly
  runs *before* `initialize_buffer`, which needs the synced `train_freq` for `rollout_buffer_size`;
  `min_episode_for_eval` reproduces the old off-policy condition exactly.
- `__all__` in `assume/__init__.py` — every listed name is actually imported.
- Call-site arity for the new 3-tuple `get_actions` — all six call sites in
  `learning_strategies.py` / `portfolio_learning_strategies.py` are updated consistently, and
  `add_actions_to_cache` handles `extra_data=None`. (Only the *test mocks* are stale — T3.)
- LSTM `_extract_features` split — `LSTMActorPPO._compute_mean` faithfully reproduces
  `LSTMActor.forward`'s reshape/batching/hidden-state handling (fresh zero states per call,
  correct `is_batched` squeeze). `mappo.get_action`'s `obs.unsqueeze(0)` → `squeeze(0)` is
  consistent for both actor types.
- `RolloutBufferSamples` field order matches `sample()`'s tuple and mappo's `batch.*` usage.
- `Path` is imported and the `continue_learning` load path validates and raises correctly.
- `compute_gradient_step_range` uses the synced `train_freq`.
- mappo's external `rollout_buffer.pos -= 1` / `full = False` is consistent with
  `compute_returns_and_advantages` and `get()`.
- Algorithm hierarchy is sound: TD3 / DDPG / PPO → `ActorCriticAlgorithm` → `RLAlgorithm`.

---

## Suggested order of work

1. **A1** — restore the optional-`torch` contract. Nothing else can be validated on a
   non-learning install until this is fixed, and it gates CI's `require_learning` marker.
2. **T1, T2, T3** — get the suite green. These are mechanical, and T1 restores the atomic-swap
   regression tests that currently guard nothing.
3. **A2** — migrate `example_03*` and the notebooks, and add a legacy-key shim with a
   `DeprecationWarning`.
4. **B1, D1, D2, C4, B4** — silent-wrongness bugs. Each produces plausible-looking output while
   doing the wrong thing, so none of them will announce itself in a training run.
5. **B2 / B3** — decide whether terminal flags are real. If yes, fix the off-by-one; if no, delete
   the `dones` and `values` columns rather than leaving load-bearing-looking dead fields.
6. **E1** — add a real `update_policy` test for MAPPO and MADDPG before the next refactor. B1 is
   a one-line bug that any executed update over a size-1 batch would have caught.
7. **C1, C2, C3, A3, A4, A5, B5-B9** — remaining correctness and robustness work.
8. **F1, F2, F3** — config alignment, unexplained example changes, doc corrections.

---

## Environment note

- `ruff check assume/ tests/` - clean.
- Comparison runs against `main` used a temporary git worktree, since removed. **No repo files
  were modified during this review**; `git status` is clean.
- `torch` and `tensorboard` were installed into `.venv` in order to run the test suite (the venv
  lacked them). No project files or lockfiles were touched.
- Unrelated, but noticed: `uv run` fails to resolve in this repo — the `docs` extra requires
  `ipython>=9`, which needs Python ≥ 3.11, while `requires-python` is `>=3.10`. Tests were run via
  `.venv/bin/python -m pytest` instead.

</details>

kim-mskw:
Comment for my review status:

Tests are not fixed yet.

A1 — Done. I was already skeptical of the public-API tests anyway, so I deleted them entirely along with the unusual __init__.py re-exports. Everything runs cleanly now, and this resolves the torch hard-dependency issue.
A3 — actor_architecture doesn't need to be defined separately for off-policy and on-policy, and the same applies to the other duplicated fields. These were leftovers from the old AlgorithmConfig, which is no longer used — any config item shared across all algorithms now lives in LearningConfig directly.
A4 — Done.
A5 — Done.

@Harshul-18 please take care of the remaining point:

A2

examples/inputs/example_02a/config.yaml
```diff
@@ -18,18 +18,31 @@ base:
     algorithm: matd3
     learning_rate: 0.001
     training_episodes: 100
-    episodes_collecting_initial_experience: 5
+    actor_architecture: mlp
     train_freq: 100h
-    gradient_steps: 10
     batch_size: 128
```
mthede, line 23: for MAPPO, isn't the train_freq actively restricting the batch size? so e.g. in 03a with 10h, the rollout buffer will only contain 10 samples before updating and resetting the on-policy buffer? can we somehow align the naming to conventional on-policy parameter names to make the impact easier understandable?

Harshul-18:
Thanks everyone for the detailed reviews, help, and contributions. I’ve addressed several [findings](https://github.com/assume-framework/assume/pull/721#issuecomment-5407195867). Could you review the updated changes and approve the pending workflows?

PR #729, opened 2026-01-30
Smooth reward calculation for nice gradients

kim-mskw:
## Description

The current reward calculation in `EnergyLearningStrategy` uses a hard threshold (step function) to determine whether a bid is accepted or not. This creates:
- Discontinuous reward signals that make learning difficult
- Sharp gradients ( or honestly no gardients) at the market clearing price boundary

In another study, I replaced a nasty binary with a sigmoid, and it did miracles. 
Replace the discrete acceptance logic with a **smooth sigmoid-based acceptance probability** that provides:
- Continuous, differentiable reward signals
- Gradual transitions around the market clearing price
- More stable learning dynamics?

image
Instead of the orange dashed line (old reward) I implemented the sigmoid rewrad to have proper gradients. This will lead to inaccuracies for the reward close to the marginal costs. They get smaller the steeper the curve (chose k higher) 

### In `EnergyLearningStrategy.calculate_reward()`:

1. **Sigmoid acceptance probability:**
   ```python
   k = 2  # controls steepness
   accept_prob = 1 / (1 + exp(k * (bid_price - market_clearing_price)))
   accepted_volume_smooth = accepted_volume * accept_prob

## Checklist

## Testing
Compared it for example_02a (green and grey with sigmoid) and example_02b (oragne and blue with sigmoid)

<img width="1371" height="297" alt="image" src="https://github.com/user-attachments/assets/c257cbf6-f13a-4cfe-98d5-b236e96bbbba" />


For an agent that sets the market price as in example_02a, it does not have any effect besides a differently scaled reward. As you can see the profit behavior is identical. However, for many agents who are supposed to bid their marginal costs most of the time, they seem to find it quicker and stay close to their marginal costs.

### DIFF
--- a/assume/strategies/learning_strategies.py
+++ b/assume/strategies/learning_strategies.py
@@ -694,7 +694,7 @@ class EnergyLearningStrategy(TorchLearningStrategy, MinMaxStrategy):
 
         opportunity_cost_smooth = (
             (market_clearing_price - marginal_cost)
-            * (unit.max_power - accepted_volume_smooth)
+            * (unit.max_power - accepted_volume_total_smooth)
             * duration
         )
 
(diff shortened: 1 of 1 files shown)


reinecfi:
Smoothing the reward sounds like a great idea in general and this might come handy for a lot of cases.

Still I think there might be an issue regarding price setters:
When a supplier would be able to set the price between the prices M0 and M1 the reward can give false incentives: 
As at the market price the smoothing function is exactly 1/2, the reward suddenly spikes below the lower boundary (M0 = 10 in the plot below) because the smoothing value rapidly rises towards 1. 
The global optimum will be in the price taking region if M1 <= 2*M0 because the smoothing function is halfing the reward while oneself is price setting. 

As a result price setters might not choose to do so because the global optimum (max reward) is at a way lower price.

grad_smoothing

PR #731, opened 2026-02-02
Upgrade pandas and pypsa

jrasko:
Clones #727 
## Description
Upgrade to pandas 3 and to pypsa 1.0.8

## Checklist


jrasko:
It seems like pandas 3 breaks PyPSA. The upgrade results in some shape casting errors inside of pypsa, they don't seem related to ASSUME. I will add a local example and open an issue in their project.

examples/notebooks/10_DSU_and_flexibility.ipynb
```diff
@@ -2390,6 +2395,7 @@
     "    title=\"Line Loading\",\n",
     "    bus_sizes=5e-2,  # Size of bus markers\n",
     "    bus_alpha=1,  # Transparency of bus markers\n",
+    "    bus_colors=[0, 0],  # TODO remove when fixed in pypsa\n",
```
maurerle, line 2398: Link to issue?

assume/markets/clearing_algorithms/redispatch.py
```diff
@@ -193,6 +193,7 @@ def clear(
             status, termination_condition = redispatch_network.optimize(
                 solver_name=self.solver,
                 solver_options=self.solver_options,
+                log_to_console=False,
```
maurerle, line 189: How did you know about that? :) this should set the OutputFlag correctly as well - so we might remove this as well?

One can run the redispatch example and see that the simulation runs through without being interrupted by pypsa/linopy/solver output to verify.

maurerle (approved):
tested example_01d with gurobi and highs - no output anymore. Tested with existing env and fresh created env.

Looks good! Thanks

examples/notebooks/10_DSU_and_flexibility.ipynb
jrasko: https://github.com/assume-framework/assume/issues/743

PR #732, opened 2026-02-02
added portfolio-rl-strategy

chiara-fb:
For @kim-mskw to review

Draft of portfolio strategy with reward function which includes
- profits relative to competitive solution
- additional reward for anticompetitive behavior
- additional reward for price-setting unit


assume/strategies/portfolio_learning_strategies.py
```diff
@@ -0,0 +1,690 @@
...
+    #                 units_on[unit_id] = inflex_gen, flex_gen, marginal_cost
+
+    #             # if expected profits lower than start cost --> OFFLINE
+    #             elif mustrun_profit < - unit.hot_start_cost: 
+    #                 units_off[unit_id] = inflex_gen, flex_gen, marginal_cost
+            
+    #             # else --> stay ONLINE
+    #             else:
+    #                 units_on[unit_id] = inflex_gen, flex_gen, marginal_cost    
+
```
kim-mskw: I think you should double check if this call is correct. In flexable we call 
``            marginal_cost_flex = unit.calculate_marginal_cost(
                start, current_power + max_power
            )
``
might be only due to ramping, but I am not completly sure right now and it should not work properly in your 2019 example when you have ramping there.

kim-mskw: Is this CPU call necessary? we do not do it in `learning_strategies.py`and should slow down performance.

kim-mskw: is that comment outdated because your are bidding the mc of all units correct?

kim-mskw: Is that the spoon feeding? :D

kim-mskw: Why is your random exploration an int? As far as I understand you are drawing from {-1.0, -0.9, -0.8, ..., 0.0, ..., 0.9, 1.0} correct? This is then scaled to 1 to 3 * mc

kim-mskw: Am I overlooking something or does your reward have nothing to do with your profit right now?

assume/strategies/portfolio_learning_strategies.py
```diff
@@ -0,0 +1,690 @@
...
+    Actions are: total flexible capacity to be bid and prices for flexible generation, rescaled 
+    from a range of [-1, 1] in the `calculate_bids` method, then translated into unit-specific bids.
+
+    Rewards are based on profit from transactions, minus operational and opportunity costs. Key components include:
+
+    - **Profit**: Determined from the income generated by accepted bids, calculated as the product of
+      accepted price, volume, and duration.
+    - **Operational Costs**: Includes marginal costs and start-up costs when a unit transitions between
+      on and off states.
```
kim-mskw, line 40: I could not find the benchmark mc bidding? Where is it?

chiara-fb: In `calculate_reward` under `comp_profit`

assume/strategies/portfolio_learning_strategies.py
chiara-fb: That's correct. I wanted to have random, uniformly distributed noise in the collect initial experience instead than normally distributed noise. A cleaner way to do this could be: 
`curr_action = -2 * th.rand_like(curr_action) + 1`

PR #734, opened 2026-02-04
Notebook fixes

### LINKED ISSUE #726: Notebook examples fail on main
It seems like some example notebooks have errors on the current main HEAD.

kim-mskw:
## Related Issue
Closes #726 

## Description
Notebook 9a depends on the scenario results of notebook 8. Also implemented a logic that runs notebook 8 locally if the files are not present yet. Former this was only implemented for the colab version. 
Adjusted documentation. 

## Checklist

Colab notebook will not work as they are in line with the current main which as set_random_seed but the last release does not have that...


jrasko (commented):
The notebook does now execute on a clean work environment ✔️ 

When running again, the user will be asked if he want to overwrite, selecting 'no' results in an Assume error. We may prevent this behavior by always overwriting.

Additionally, the notebook now creates a "08_market_zone_coupling.nbconvert.ipynb" file. Is this intended? If yes, please add the file to the .gitignore so it can't be commited by accident.

jrasko (approved):

PR #737, opened 2026-02-06
Calculation of available minimum/maximum power output

### LINKED ISSUE #722: Minimum power output > maximum power output, if unit availability == 0.0
When incorporating unit (un)availabilities, like planned and unplanned outages, the min_power calculation is unintuative. During an outage, the availability in availibility_df.csv would be 0.0. This leads to a max_power value of 0.0, but the min_power value is still at the unit's min_power > 0. 

With the typical two-bid RL strategy, the power plant was dispatched despite availability == 0.0.

Image

Can we just clip the min_power value [here ](https://github.com/assume-framework/assume/blob/8df8d44e1b734d4edf5b875239590722a723a754/assume/units/powerplant.py#L287) to max_power? Or do we need to take it differently into account due to heat, capacity, base load, etc?

mthede:
## Related Issue
Closes #722

## Description
- Fix infeasible power outputs: if maximum additional power output is below minium power, then do not provide additional power at all
- Adjust test accordingly

## Checklist

### DIFF
--- a/assume/units/powerplant.py
+++ b/assume/units/powerplant.py
@@ -289,47 +289,59 @@ class PowerPlant(SupportsMinMax):
         self, start: datetime, end: datetime, product_type="energy"
     ) -> tuple[np.ndarray, np.ndarray]:
         """
-        Calculates the minimum and maximum power output of the unit and returns it,
-        while considering heat demand and positive capacity reserve power.
+        Calculates the additional minimum and maximum power output.
+
+        "Additional" means relative to the already-committed power, e.g. at other markets.
+        Considers heat demand, positive capacity reserve (capacity_pos reduces headroom),
+        and negative capacity reserve (capacity_neg raises the minimum obligation).
+
         Args:
-            start (pandas.Timestamp): The start time of the dispatch.
-            end (pandas.Timestamp): The end time of the dispatch.
-            product_type (str, optional): The product type of the unit. Defaults to "energy".
+            start (datetime): The start time of the dispatch.
+            end (datetime): The end time of the dispatch (exclusive).
+            product_type (str, optional): The product type. Defaults to "energy".
 
         Returns:
-            tuple[pandas.Series, pandas.Series]: The minimum and maximum power output of the unit.
+            tuple[np.ndarray, np.ndarray]: The additional minimum and maximum power output of the unit.
 
         Note:
             The calculation does not include ramping constraints and can be used for arbitrary start times in the future.
+            Ramping constraints can be applied afterwards via calculate_ramp().
         """
         # end includes the end of the last product, to get the last products' start time we deduct the frequency once
         end_excl = end - self.index.freq
 
         base_load = self.outputs["energy"].loc[start:end_excl]
         heat_demand = self.outputs["heat"].loc[start:end_excl]
+        capacity_pos = self.outputs["capacity_pos"].loc[start:end_excl]
         capacity_neg = self.outputs["capacity_neg"].loc[start:end_excl]
 
-        # needed minimum + capacity_neg - what is already sold is actual minimum
-        min_power = self.min_power + capacity_neg - base_load
-        # min_power should be at least the heat demand at that time
-        min_power = min_power.clip(min=heat_demand)
+        availability = self.forecaster.availability.loc[start:end_excl]
+        available_power = availability * self.max_power
 
-        available_power = self.forecaster.availability.loc[start:end_excl]
-        max_power = available_power * self.max_power
         # check if available power is larger than max_power and raise an error if so
-        if (max_power > self.max_power).any():
+        if (available_power > self.max_power).any():
             raise ValidationError(
                 message=f"Available power is larger than max_power for unit {self.id} at time {start}.",
                 id=self.id,
                 field="availability",
             )
 
-        # provide reserve for capacity_pos
-        max_power = max_power - self.outputs["capacity_pos"].loc[start:end_excl]
-        # remove what has already been bid
-        max_power = max_power - base_load
-        # make sure that max_power is > 0 for all timesteps
+        # subtract positive reserve commitment and already-dispatched base_load from available power
+        max_power = available_power - capacity_pos - base_load
+        # additional power can never be negative
         max_power = max_power.clip(min=0)
+
+        # actual minimum is technical minimum + negative reserve obligation - already dispatched base_load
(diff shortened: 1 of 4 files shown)


assume/common/utils.py
```diff
@@ -544,29 +544,45 @@ def create_incidence_matrix(lines, buses, zones_id=None):
...
                     f"Unit '{column}' not found in powerplants dataframe. Skipping normalization for this unit."
                 )
+        # Get the max_power for the current unit from the powerplants dataframe
+        min_power = powerplants_df.loc[column, "min_power"]
+        if (validated_df[column] * max_power < min_power).any():
+            logger.warning(
+                f"Availability for unit '{column}' contains values that would result in available power below the minimum power. Setting it to 0.0 for those hours. "
+                f"Please check input availabilities and minimum power for this unit."
+            )
+            validated_df.loc[validated_df[column] * max_power < min_power, column] = 0.0
```
kim-mskw: From my understanding, this is a technical issue with the power plant. I am not sure we should correct this here in general for all power plants. Most likely it will apply for all of them but we do not know that error tracing will be harder then.

kim-mskw (changes requested):

assume/units/powerplant.py
```diff
@@ -279,28 +279,24 @@ def calculate_min_max_power(
...
+        max_power = available_power - capacity_pos
         # remove what has already been bid
         max_power = max_power - base_load
         # make sure that max_power is > 0 for all timesteps
         max_power = max_power.clip(min=0)
 
+        # 'needed minimum' + capacity_neg - 'what is already sold' is actual minimum
+        min_power = self.min_power + capacity_neg - base_load
+        # min_power should be at least the heat demand at that time
+        min_power = min_power.clip(min=heat_demand, max=max_power)
```
maurerle:
```suggestion
        heat_demand_still_needed = heat_demand - base_load
        min_power = min_power.clip(min=heat_demand_still_needed)
        # if the max_power is smaller than min_power due to reduced availability, we can not turn the powerplant on
        min_power[max_power < min_power] = 0
        max_power[max_power < min_power] = 0
```

What do you think of this?
This should fix the heat demand (which we are currently barely using) and correctly set min and max power if the availability is constrained

assume/common/utils.py
mthede: Okay, I understand it would be quite a strong general adjustment. I'll remove it.

assume/units/powerplant.py
mthede: Thanks for your suggestion! I've slightly adjusted it so that it works as expected (some tests were failing).

mthede:
I have finally addressed your comments and removed the validation of the availability inputs. The technical feasibility check now happens inside calculate_min_max_power(). This changes our expected behavior (see test_powerplant_availability, where we wanted to set the power output to min_power instead of 0).

mthede:
I just don't understand why the nodal clearing tests are failing for 3.12 and 3.14 suddenly due to PyPSA related things I haven't touched ... it doesn't seem to be related to the PyPSA version, as another PRs pipeline ran yesterday without issues with the same version

**EDIT:** Well, it's because of a new xarray version from two days ago. Do we need to pin it down?

assume/units/powerplant.py
```diff
@@ -292,48 +292,60 @@ def calculate_min_max_power(
...
                 field="availability",
             )
 
-        # provide reserve for capacity_pos
-        max_power = max_power - self.outputs["capacity_pos"].loc[start:end_excl]
-        # remove what has already been bid
-        max_power = max_power - base_load
-        # make sure that max_power is > 0 for all timesteps
+        # subtract positive reserve commitment and already-dispatched base_load from available power
+        max_power = available_power - capacity_pos - base_load
```
kim-mskw, line 333: I think instead of just clipping here we should raise a warning, that former commitments violate max_power.

mthede: Instead of clipping or in addition? So do you want to pass negative max_power values? I guess the bidding couldn't handle this

assume/units/powerplant.py
```diff
@@ -292,48 +292,60 @@ def calculate_min_max_power(
...
         max_power = max_power.clip(min=0)
 
+        # actual minimum is technical minimum + negative reserve obligation - already dispatched base_load
+        min_power = self.min_power + capacity_neg - base_load
+        # must also cover any heat demand not yet met by base_load
+        # clip to 0: negative values mean base_load already covers heat demand fully
+        heat_demand_still_needed = (heat_demand - base_load).clip(min=0)
+        min_power = min_power.clip(min=heat_demand_still_needed)
+
+        # if additional max_power is below the technical minimum, the unit cannot run at all
```
mthede, line 347: @maurerle I slightly changed your proposition. what do you think now?

assume/units/powerplant.py
```diff
@@ -292,48 +292,60 @@ def calculate_min_max_power(
...
-            start (pandas.Timestamp): The start time of the dispatch.
-            end (pandas.Timestamp): The end time of the dispatch.
-            product_type (str, optional): The product type of the unit. Defaults to "energy".
+            start (datetime): The start time of the dispatch.
+            end (datetime): The end time of the dispatch (exclusive).
+            product_type (str, optional): The product type. Defaults to "energy".
 
         Returns:
-            tuple[pandas.Series, pandas.Series]: The minimum and maximum power output of the unit.
+            tuple[np.ndarray, np.ndarray]: The additional minimum and maximum power output of the unit.
```
maurerle, line 304: I think that this needs to be `numpy.ndarray` for intersphinx to catch this and link to the numpy docs, but this is not crucial though.
I think we do have some inconsistencies anyway.

maurerle (approved):

assume/units/powerplant.py
mthede: Ahh okay, I think I just aligned it with other docs in our code. But we'll clean the inconsistencies up seperately.

assume/units/powerplant.py
kim-mskw: In addition, or maybe even raise an error, or are there situations in which such behavior could happen and is fine?

mthede: I added a warning, but wouldn't throw an error. Maybe some multi-market scenarios exist where at some intermediate point infeasible dispatch is ok?

kim-mskw (approved):

PR #745, opened 2026-02-24
Add structured validation error

jrasko:
## Description
Use structured errors to identify the faulty inputs and allow the frontend to make them visible to the user

## Checklist

### DIFF
--- a/assume/common/exceptions.py
+++ b/assume/common/exceptions.py
@@ -9,5 +9,12 @@ class AssumeException(Exception):
     pass
 
 
+class ValidationError(ValueError):
+    def __init__(self, message: str, id: str, field: str):
+        super().__init__(message)
+        self.field = field
+        self.id = id
+
+
 class InvalidTypeException(AssumeException):
     pass
--- a/assume/units/powerplant.py
+++ b/assume/units/powerplant.py
@@ -323,6 +323,8 @@ class PowerPlant(SupportsMinMax):
         if (max_power > self.max_power).any():
             raise ValidationError(
                 message=f"Available power is larger than max_power for unit {self.id} at time {start}.",
+                id=self.id,
+                field="availability",
             )
 
         # provide reserve for capacity_pos
--- a/assume/units/powerplant.py
+++ b/assume/units/powerplant.py
@@ -9,6 +9,7 @@ from functools import lru_cache
 import numpy as np
 
 from assume.common.base import MinMaxStrategy, SupportsMinMax
+from assume.common.exceptions import ValidationError
 from assume.common.fast_pandas import FastSeries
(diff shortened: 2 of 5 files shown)


docs/source/release_notes.rst
```diff
@@ -20,16 +20,8 @@ Upcoming Release
   - **Delete environment.yaml**: The environment.yaml file has been removed from the repository to simplify maintenance and was completely redundant with the `pyproject.toml`. Users can as before create their own environment using the provided pip installation instructions, which allows for more flexibility and easier updates.
   - **Add validation for simulation setup**: Added checks to validate the simulation setup for common issues, such as missing bidding strategies or inconsistent market configurations. Warnings are issued to inform users of potential problems that could affect simulation results.
   - **Added reward calculation for unit operators**: Unit operators have now the opportunity to calculate rewards based on the returned orderbooks for their own purposes. This enables learning strategies on unit operator level / portfolio learning strategies.
-
-0.5.7 - (18th February 2026)
-============================
-
-**Bug Fixes:**
-
-- **Notebook 10a**: Bidding strategy- strings were replaced by the method. Tutorial works error-free.
```
jrasko, line 29: removed this since it is no new feature and should work in all versions anyhow.
Also this isn't published as a version yet, so this paragraph makes no sense either way

assume/common/exceptions.py
```diff
@@ -9,5 +9,12 @@ class AssumeException(Exception):
     pass
 
 
+class ValidationError(ValueError):
```
maurerle, line 12: I would like to have every custom Exception in this File to inherit from AssumeException as well..?
You should inherit from both to not break existing catches of ValueError, but also align with the AssumeException idea.

assume/units/powerplant.py
```diff
@@ -290,8 +321,8 @@ def calculate_min_max_power(
         max_power = available_power * self.max_power
         # check if available power is larger than max_power and raise an error if so
         if (max_power > self.max_power).any():
-            raise ValueError(
-                f"Available power is larger than max_power for unit {self.id} at time {start}."
+            raise ValidationError(
+                message=f"Available power is larger than max_power for unit {self.id} at time {start}.",
```
maurerle, line 325: also needs self.id?

maurerle (approved):
The rational behind this change is to support #705 and the assume-gui part.
This helps to have more information about which component causes a validation error in tools using assume (like the assume-gui).

assume/common/exceptions.py
jrasko: I still not get the idea of the Assume exception. What is its purpose (when should it be thrown)? To which use-case does this exception add value? We might use this opportunity to clarify and document this.

PR #747, opened 2026-02-25
Rework forecast interface

### LINKED ISSUE #724: Generic Forecast Interface
## Motivation
We need a **generic forecast interface** that allows users to **select between different forecast implementations**. This should provide a stable abstraction for current and future forecast options and avoid hard-coding forecasting logic into specific components.

## Scope / Requirements
- Introduce a **common forecaster interface** (e.g., `BaseForecaster`)
- Enable users to **choose a forecaster implementation via config** or in the **csv file**?
- Support plugging in additional forecasters without changing core logic (extensible design).
- Ensure compatibility with future forecast types, including:
  - residual load forecasts,
  - forecasts derived from multiple signals (e.g., demand, RES generation),
  - potentially portfolio- or market-specific forecast variants.

Some inspiration might be found here: https://www.mlsustainableenergy.com/feat/ and applied in AMIRIS https://www.mlsustainableenergy.com/research/application-in-amiris/

## Relation to existing work
- **Issue #719** will become **one forecaster option** a`la @gugrimm that users can select through this interface.

## Open Questions
- What forecast objects should be supported initially (price, load, RES, residual load)?
- Should the interface return point forecasts only, or also probabilistic forecasts / uncertainty? Then we would need more output variabels that all units can handel, but that is one potential extension

## Deliverables
- Interface + one default implementation with the current naive forecaster
- Integration into configuration / initialization flow
- Documentation + small example showing how to switch forecasters (incl. #719 as option)

reinecfi:
## Related Issue
Closes #724 

## Description

We need a generic forecast interface that allows users to select between different forecast implementations. This should provide a stable abstraction for current and future forecast options and avoid hard-coding forecasting logic into specific components.

The inferface will look like the following:
There is a BaseForecaster that has access to the following method :

1. `preprocess`: this method prepares information / forecasts / etc. for steps 2. and 3.
2. `initialize`: this method initializes all forecast timeseries, after all units are created
3. `update`: this method is used to change forecasts during runtime (shall be called during calculate bids of bidding strategies)

All three methods (preprocess, initialize, update) can be set either via config or csv (otherwise default values are set).

Each Unit type has one corresponding Forecaster class that is a subclass of BaseForecaster and thus inherits the interface.

## Checklist

- DSM units use the `setup_model` method in their `__init__`, which expects correct forecasts. This is not possible anymore as the units are created before the forecasts are initialized. I am not sure how to handle this properly in a clean way right now. Still, in order to have a working version, I have put 'setup_model' into the respective Forecaster.initialize().
- It was planned to cache the forecast function outputs: this is currently not, as the inputs are not hashable (till now, at least).
- The `update` calls are currently not integrated in any `calculate_bids` methods.


assume/common/base.py
```diff
@@ -751,6 +751,14 @@ def remove_empty_bids(self, bids: list) -> list:
 
         return cleaned_bids
 
+    def update_forecasts_if_needed(unit: BaseUnit, *args, **kwargs):
+        # if last_forecast_update <= 30 day ago..?:
+        #    return False
+
+        # hier bekommt der Forecaster die Infos um sich zu aktualisieren
+        # unit.forecaster.update(unit.outputs["eom_price"])
```
maurerle: please use english language to be consistent across the code base :)

assume/common/forecast_algorithms.py
```diff
@@ -0,0 +1,575 @@
...
+    from assume.units.storage import Storage
+
+ForecastIndex: TypeAlias = FastIndex | pd.DatetimeIndex | pd.Series
+ForecastSeries: TypeAlias = FastSeries | list | float | pd.Series
+
+log = logging.getLogger(__name__)
+
+
+# advanced on: https://discuss.python.org/t/memoizing-based-on-id-to-avoid-implementing-and-computing-hash-method/87701/7
+def custom_lru_cache(func_or_None=None, maxsize=128, typed=False, hasher=id):
```
maurerle: I would rather just implement a `__eq__` (or `__hash__`) function for BaseUnit, ForecastIndex, MarketConfig - then the list[BaseUnit] ist hashable as well and we do not need this huge wrapper which is hard to maintain.

assume/common/forecast_algorithms.py
```diff
@@ -0,0 +1,575 @@
...
+        wrapper.cache_info = decorated.cache_info
+        wrapper.cache_clear = decorated.cache_clear
+        if hasattr(decorated, "cache_parameters"):
+            wrapper.cache_parameters = decorated.cache_parameters
+        return wrapper
+
+    return decorator if func_or_None is None else decorator(func_or_None)
+
+
+def is_renewable(name: str) -> bool:
```
maurerle: There is `only_renewables` in market_objects.py

Maybe put both into utils..? Not sure, right now

assume/common/forecast_algorithms.py
```diff
@@ -0,0 +1,575 @@
...
+    Returns: max available power: shape (num_units, forecast_len)
+    """
+    return pd.DataFrame(
+        [unit.max_power * unit.forecaster.availability for unit in units], index=index
+    )
+
+
+@custom_lru_cache
+def sort_units(units: list[BaseUnit], market_id: str | None = None):
+    # FIXME: This is not nice but circumvents circular import for now
```
maurerle: We need a clear hierarchy: (-> means imports)

World -> forecast_algorithm -> forecaster
Units -> forecaster

The forecaster should not import forecast_algorithm

assume/units/building.py
```diff
@@ -142,7 +142,7 @@ def __init__(
             self.components["pv_plant"][key] = self.forecaster.pv_profile
 
         # Initialize the model
-        self.setup_model(presolve=True)
+        # self.setup_model(presolve=True)  # NOTE: called in forecaster initialization again!!!
```
maurerle, line 145: I think it looks fine!
just as question: why did you move this to the forecaster?
Just to know the background of this change :)

assume/world.py
```diff
@@ -917,3 +811,30 @@ def add_unit_instance(self, operator_id: str, unit: BaseUnit):
...
+                markets,
+                forecast_df,
+                unit,
+            )
+
+    # The update routine is not needed. We call it in calculate_bids or through a scheduled task during runtime
+    # see update_forecasts_if_needed
+    def update_forecasts(self):
+        for unit in self.units.values():
+            # for forecast in unit.forecaster:
```
maurerle: remove comment

examples/examples.py
```diff
@@ -121,7 +121,7 @@
     data_format = "local_db"  # "local_db" or "timescale"
 
     # select the example to run from the available examples above
-    example = "small_with_vre_and_storage"
+    example = "small_household"
```
maurerle: revert changes to this file

maurerle (commented):
I like this change in general. Conceptionally, this looks fine.
I just have some comments to address, would like to test this myself when you are ready, so that we can merge this soon :)

assume/common/forecast_algorithms.py
reinecfi: Maintainability of the wrapper is a good point. Still the problem is only the pandas.DataFrame `forecast_df ` (currently the Units and Marketconfigs are given as `dict.values()` and thus hashable.)

To provide a bit more context: `forecast_df `is only used like a dictionary (similar to this: `forecast_df.get("key")`). 
I thought about a few other hacky solutions to this, but they weren't nice either (e.g. create a frozenset of the dataframe and in the functions create a dict from it).

Maybe you know a better solution to this @maurerle?

examples/examples.py
```diff
@@ -121,7 +121,7 @@
     data_format = "local_db"  # "local_db" or "timescale"
 
     # select the example to run from the available examples above
-    example = "small_household"
+    example = "small_with_vre_and_storage"
```
maurerle, line 124: please revert this change

assume/common/forecaster.py
```diff
@@ -59,13 +54,17 @@ def __init__(
         residual_load: dict[str, ForecastSeries] = None,
         availability: ForecastSeries = 1,
         forecast_algorithms: dict[str, str] = {},
+        forecast_registries: dict[str, dict] = None,
```
maurerle, line 180: That's nice :)
But well - we need docs for it.

examples/examples.py
reinecfi: I did change it back from "small_household" to "small_with_vre_and_storage" 

> revert changes to this file

"small_with_vre_and_storage" is consistent to main so for me this sound correct like it is currently

assume/common/forecast_algorithms.py
maurerle: This is resolved as discussed on Tuesday

assume/units/building.py
maurerle: We still need to look into better ways for this..?

reinecfi: Yes probably. Afaik, the dsm model depends on the given forecast. If one wants to use any `update` functionality of forecasts, this would impose the need of an `update` functionality in the dsm units model.
As a result: it would be easy if we had such an update functionality in the dsm units model. Then we could just call `update` it  after / in world.initialize_forecasts.

Till then i would stick to doing `setup_model` at roughly this place (in the forecaster.initialize or world.initialize_forecasts).

Note: Whenever the forecaster.initialize functionality is not used (tests) this would impose that one calls dsm_unit.setup_model by hand. 

For convenience one could also add a flag or sanity check that does dsm_unit.setup_model automatically if not done already at the start of simulation

maurerle (approved):

Issue #746, opened 2026-02-24
Convergence 03a


kim-mskw:
## Convergence irregularities in selected runs (e.g., 03a)

During the preparation of several studies, we observed convergence issues in some runs (most notably **03a**). While changes to the reward specification and related components that have been made in the last months yielded **similar aggregate reward levels** as before, the learning signal appears to be **driven disproportionately by a small number of time steps**—in particular, isolated hours in which **maximum prices** were realized. As a consequence, the overall reward can look stable or even improved even though the underlying policy may not have converged in a robust or economically meaningful way.

I am currently **debugging** this behavior and using this issue to **systematically document** what was changed, which diagnostics were applied, and how these changes affected convergence.

kim-mskw:
## Convergence main
`Note: In my tensorboard = example_03a_base_case_2019_main`

<img width="1797" height="438" alt="Image" src="https://github.com/user-attachments/assets/e9178b17-6ce2-4128-81b9-dcbb6276b376" />

The first part of the convergence graph, in general, looks really good again. Nice increase to levels that are around the initial_exploration values again. However, everything raises suspicion similarly to before and the distribution of the actions shows that again. 

<img width="1190" height="407" alt="Image" src="https://github.com/user-attachments/assets/e4606f49-251d-47ad-a2df-d025b876ef3a" />

<img width="1201" height="395" alt="Image" src="https://github.com/user-attachments/assets/e5a0b663-7cb4-49dc-b3e2-08ad5300fd0d" />

<img width="1189" height="405" alt="Image" src="https://github.com/user-attachments/assets/51c55191-210c-4df4-8ae5-b4ad20746af2" />

Very quickly, they converge to the extreme behavior of bidding max and min prices, and besides the profit climbing again, they stick to it and further intensify it.

PR #750, opened 2026-03-01
Tests & comments for convergence 03a

kim-mskw:
## Related Issue
Closes stuff related to #746 

## Description
See issue, vurrent changes: 
- add convert tensor tests
- add buffer tests
- some comments and make sure marginal costs is last individual obs

## Checklist

### DIFF
--- a/assume/strategies/learning_strategies.py
+++ b/assume/strategies/learning_strategies.py
@@ -1212,7 +1212,7 @@ class RenewableEnergyLearningSingleBidStrategy(EnergyLearningSingleBidStrategy):
         scaled_available_power = available_power[0] / unit.max_power
 
         individual_observations = np.array(
-            [scaled_total_dispatch, scaled_marginal_cost, scaled_available_power]
+            [scaled_total_dispatch, scaled_available_power, scaled_marginal_cost]
         )
 
         return individual_observations
--- a/tests/test_utils_rl.py
+++ b/tests/test_utils_rl.py
@@ -1,0 +1,474 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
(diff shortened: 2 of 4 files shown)


tests/test_utils_rl.py
```diff
@@ -0,0 +1,474 @@
...
+from assume.common.utils import convert_tensors
+
+
+def test_convert_orderbook_list_with_numpy_floats():
+    """Test conversion of orderbook list with numpy floats (real data structure)."""
+
+    try:
+        import torch as th
+    except ImportError:
+        pytest.skip("PyTorch not installed")
```
maurerle: just add `torch = pytest.importorskip('torch')` to the top level import:
https://docs.pytest.org/en/4.6.x/reference.html#pytest-importorskip

And please also add `@pytest.mark.require_learning` to the learning tests to allow to select or deselect these :)

assume/strategies/learning_strategies.py
```diff
@@ -1212,7 +1212,7 @@ def get_individual_observations(
         scaled_available_power = available_power[0] / unit.max_power
 
         individual_observations = np.array(
-            [scaled_total_dispatch, scaled_marginal_cost, scaled_available_power]
+            [scaled_total_dispatch, scaled_available_power, scaled_marginal_cost]
```
maurerle, line 1215: So this is the bug which introduced the different behavior? Great catch!

learning_strategies.py Line 527 says: `# Assumes last dimension of the observation corresponds to marginal cost`

We should find a way to do this more robust.. :|

maurerle (approved):

assume/strategies/learning_strategies.py
kim-mskw: No it is not unfortunately, because this is only in the renewable_learning staretgy which is not even present in 03a...

PR #751, opened 2026-03-02
Learning with new min power default

kim-mskw:
## Related Issue
related to #746 

## Description

We had a change of the default behavior of min_power that "breaks" some learning function. 
**Change:** `min_operating_time` changed from `0` → `1` when not relevant as new default

**Consequence:** Agents now submit two bids per unit:
- Bid 1: Full capacity at calculated price
- Bid 2: Only 1 MW at (possibly different) price

**Example:**
```python
{'price': tensor(58.8953), 'volume': np.float64(1.0), 'bid_id': 'pp_6_1'}  # 1 MW
{'price': tensor(86.1664), 'volume': np.float64(499.0), 'bid_id': 'pp_6_2'} # 499 MW
```

Added a distinction that only formulates two bids when min_power>1.

## Checklist

### DIFF
--- a/assume/strategies/learning_strategies.py
+++ b/assume/strategies/learning_strategies.py
@@ -436,9 +436,8 @@ class EnergyLearningStrategy(TorchLearningStrategy, MinMaxStrategy):
         end = product_tuples[0][1]
         # get technical bounds for the unit output from the unit
         min_power, max_power = unit.calculate_min_max_power(start, end)
-        # default behavior of min_power is that a running power plant has the minimum power of 1 timestep always, because it needs to be running to be running
-        # a value of 1 imposes no real constraint for the bid formulation so we can set it to 0 for the bid formulation to not have this constraint in place
-        min_power = 0 if min_power[0] == 1 else min_power[0]
+
+        min_power = min_power[0]
         max_power = max_power[0]
 
         # =============================================================================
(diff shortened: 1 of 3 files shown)


assume/strategies/learning_strategies.py
```diff
@@ -436,7 +436,9 @@ def calculate_bids(
         end = product_tuples[0][1]
         # get technical bounds for the unit output from the unit
         min_power, max_power = unit.calculate_min_max_power(start, end)
-        min_power = min_power[0]
+        # default behavior of min_power is that a running power plant has the minimum power of 1 timestep always, because it needs to be running to be running
```
mthede: How is this related to the PR problem description with the new default for `min_operating_time`?

assume/strategies/learning_strategies.py
```diff
@@ -436,7 +436,9 @@ def calculate_bids(
         end = product_tuples[0][1]
         # get technical bounds for the unit output from the unit
         min_power, max_power = unit.calculate_min_max_power(start, end)
-        min_power = min_power[0]
```
mthede, line 439: Default for `unit.min_power` is 0.0, so in case of no provided `min_power` in the input files, the bidding should result in a single bid at maximum power per default. The `min_power` in the example you posted was set in the input files. I feel like the submission of two bids is therefore expected behavior and shouldn't be overwritten?

kim-mskw: Yes I agree, it was way to late in the day to open that issue

PR #755, opened 2026-03-05
changed simulation column name to fix db query error

### LINKED ISSUE #754: DB queries fail on fresh install due to ambiguos column name simulation
### Failure of db queries on fresh install

This error happens for fresh installs in the learning examples and was originally mentioned in issue #752 

> @gugrimm I got the same issue as and there seems to be an issue with the the word "simulation" here:
> 
> https://github.com/assume-framework/assume/blob/fc3334d02048507af5bb7bb132cbbe243c68ff7b/assume/common/outputs.py#L666
> 
> As far as I could figure out this seems to be fixed if "simulation" is replaced by "ud.simulation".
> 

 _Originally posted by @reinecfi in [#752](https://github.com/assume-framework/assume/issues/752#issuecomment-4003939994)_

reinecfi:
## Related Issue
Closes #754 

## Description

A kpi query has the ambiguos column query "simulation" for kpi `capacity_factor`. This results in query errors like the following during runtime.

This is fixed by using "ud.simulation" as column name instead.

```
ERROR:assume.common.outputs:could not read query: Execution failed on sql 'select 'capacity_factor' as variable, market_id as ident, avg(power/max_power) as value from market_dispatch ud join power_plant_meta um on ud.unit_id = um."index" and ud.simulation=um.simulation where simulation = 'example_02a_base' group by market_id,variable': (sqlite3.OperationalError) ambiguous column name: simulation
[SQL: select 'capacity_factor' as variable, market_id as ident, avg(power/max_power) as value from market_dispatch ud join power_plant_meta um on ud.unit_id = um."index" and ud.simulation=um.simulation where simulation = 'example_02a_base' group by market_id,variable]
```

## Checklist

### DIFF
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -668,7 +668,7 @@ class WriteOutput(Role):
                 f"select '{variable}' as variable, market_id as ident, {kpi_def['value']} as value from {kpi_def['from_table']} where {simulation_col} = '{self.simulation_id}' group by {group_bys}"
             )
 
-        if self.episode:
+        if self.learning_mode or self.evaluation_mode:
             queries.extend(
                 [
                     f"SELECT 'sum_reward' as variable, simulation as ident, sum(reward) as value FROM rl_params WHERE episode='{self.episode}' AND simulation='{self.simulation_id}' GROUP BY simulation",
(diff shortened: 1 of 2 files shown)


reinecfi:
I think whats happening is, that market_dispatch (named ud) and power_plant_meta (named um) are joined in line 121: 
```
"from_table": 'market_dispatch ud join power_plant_meta um on ud.unit_id = um."index" and ud.simulation=um.simulation'
```

and thus afterwards there exist two simulation columns (ud.simulation and um.simulation, which are by definition of the query the same). 

So to circumvent the problem we just be explicit about which of them we choose.

maurerle (approved):

PR #756, opened 2026-03-06
Fix learning convergence with heterogeneous units

### LINKED ISSUE #746: Convergence 03a

kim-mskw:
## Related Issue
Closes #746 

## Description
**Two** things change here and tests were added: 

**1**
Pretty sure we (more @mthede) found the root cause of the convergence issue described and analysed in #746 

In `transform_buffer_data` we ensure that unit ordering is consistent when writing to the replay buffer, which is necessary for correct sampling. For that, unit IDs are explicitly sorted:

```python
unit_ids = sorted(
    set(dt for unit_data in nested_dict.values() for dt in unit_data.keys())
)
```

However, when we update the rl_algorithm gets the order here: 
` strategies = list(self.learning_role.rl_strats.values())`

The order of dict.values() is insertion order and therefore not guaranteed to match the alphabetical order used in the buffer. As a result, when iterating over strategies, heterogeneous units can end up sampling and updating on the wrong experience, i.e., they receive transitions of a different unit (the one that happens to occupy the same position in the sorted unit list). 

This did not lead to error in 02a with all learning units beacuase the order in the csv happens to be alphabetical... I can observe a major learning performance decrease if I mix up the order in the csv file to not match the alphabetical one. We can see the impact immediately, and it was confirmed to fix 03a, which is also not alphabetical. 

**2**
This PR further includes one less severe fix: 

When `store_to_buffer_and_update()` processes cached RL experience data, the last timestamp's observation and action data may not yet have an associated reward (rewards arrive one timestep later). Without proper handling, this incomplete data was lost during the cache swap.

Extended the atomic swap pattern with carry-over logic:

1. **Atomic swap**: Immediately swap cache dictionaries to new empty ones, so any new data arriving during processing goes to the new cache
2. **Carry-over**: The last timestamp's incomplete data (obs, actions, noises) is explicitly copied to the new cache

```python
# Carry over the last timestamp's data (obs, actions, noises)
last_ts = all_timestamps[-1]
self.all_obs[last_ts] = current_obs[last_ts]
self.all_actions[last_ts] = current_actions[last_ts]
self.all_noises[last_ts] = current_noises[last_ts]
```

## Checklist

### DIFF
--- a/assume/reinforcement_learning/learning_role.py
+++ b/assume/reinforcement_learning/learning_role.py
@@ -293,9 +293,13 @@ class Learning(Role):
 
         # rewrite dict so that obs.shape == (n_rl_units, obs_dim) and sorted by keys and store in buffer
         self.buffer.add(
-            obs=transform_buffer_data(cache["obs"], device),
-            actions=transform_buffer_data(cache["actions"], device),
-            reward=transform_buffer_data(cache["rewards"], device),
+            obs=transform_buffer_data(cache["obs"], device, self.rl_strats.keys()),
+            actions=transform_buffer_data(
+                cache["actions"], device, self.rl_strats.keys()
+            ),
+            reward=transform_buffer_data(
+                cache["rewards"], device, self.rl_strats.keys()
+            ),
         )
 
         if (
--- a/assume/reinforcement_learning/learning_role.py
+++ b/assume/reinforcement_learning/learning_role.py
@@ -248,14 +248,20 @@ class Learning(Role):
         # Get timestamps from cache we took
         all_timestamps = sorted(current_obs.keys())
         if len(all_timestamps) > 1:
-            # Remove last timestamp that has no reward yet
-            timestamps_to_process = all_timestamps[:-1]
+            # Identify all incomplete timesteps (no reward yet)
+            incomplete_timestamps = [
+                ts for ts in all_timestamps if ts not in current_rewards
+            ]
 
-            # Carry over the last timestamp's data (obs, actions, noises)
-            last_ts = all_timestamps[-1]
-            self.all_obs[last_ts] = current_obs[last_ts]
-            self.all_actions[last_ts] = current_actions[last_ts]
-            self.all_noises[last_ts] = current_noises[last_ts]
+            # Process only complete timesteps
+            timestamps_to_process = [
+                ts for ts in all_timestamps if ts not in incomplete_timestamps
+            ]
+            # Carry over incomplete timesteps to new cache dicts
+            for ts in incomplete_timestamps:
+                self.all_obs[ts] = current_obs[ts]
+                self.all_actions[ts] = current_actions[ts]
+                self.all_noises[ts] = current_noises[ts]
 
             # Create filtered cache (only complete timesteps)
             cache = {
--- a/assume/reinforcement_learning/learning_role.py
+++ b/assume/reinforcement_learning/learning_role.py
@@ -251,6 +251,12 @@ class Learning(Role):
             # Remove last timestamp that has no reward yet
             timestamps_to_process = all_timestamps[:-1]
 
+            # Carry over the last timestamp's data (obs, actions, noises)
+            last_ts = all_timestamps[-1]
+            self.all_obs[last_ts] = current_obs[last_ts]
+            self.all_actions[last_ts] = current_actions[last_ts]
+            self.all_noises[last_ts] = current_noises[last_ts]
+
             # Create filtered cache (only complete timesteps)
             cache = {
                 "obs": {t: current_obs[t] for t in timestamps_to_process},
--- a/assume/reinforcement_learning/learning_utils.py
+++ b/assume/reinforcement_learning/learning_utils.py
@@ -188,7 +188,9 @@ def copy_layer_data(dst, src):
             dst[k].data.copy_(src[k].data)
 
 
-def transform_buffer_data(nested_dict: dict, device: th.device) -> np.ndarray:
+def transform_buffer_data(
+    nested_dict: dict, device: th.device, keys_unit_order: list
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -120,7 +120,7 @@ ignore = ["E501", "G004", "E731"]
     "F811", # allow import redeclaration
 ]
 
-[tool.pytest]
+[tool.pytest.ini_options]
```
maurerle: The `ini_options` is old and has been removed a while ago on purpose.
Probably needs to update pytest?
Which version were you using?

kim-mskw: Really? I just installed everything fresh. I am using
```
(assume-framework) PS C:\Users\tg3533\Documents\Code\assume> pip show pytest
Name: pytest
Version: 8.3.3
Summary: pytest: simple powerful testing with Python
Home-page: 
Author: Holger Krekel, Bruno Oliveira, Ronny Pfannschmidt, Floris Bruynooghe, Brianna Laugher, Florian Bruhin, Others (See AUTHORS)
Author-email:
License: MIT
Location: C:\Users\tg3533\AppData\Local\miniconda3\envs\assume-framework\Lib\site-packages
Requires: colorama, iniconfig, packaging, pluggy
Required-by: pytest-asyncio, pytest-cov
```

assume/reinforcement_learning/learning_utils.py
```diff
@@ -188,7 +188,9 @@ def copy_layer_data(dst, src):
             dst[k].data.copy_(src[k].data)
 
 
-def transform_buffer_data(nested_dict: dict, device: th.device) -> np.ndarray:
+def transform_buffer_data(
+    nested_dict: dict, device: th.device, keys_unit_order: list
```
maurerle, line 192: So we sorted alphabetically before.

Now, we have an additional order which is provided.
I'd like to have this deterministic without providing the "right" order additionally.
Can't we use the sorted value in the retrieval as well..?

So fix "the other side"..?

assume/reinforcement_learning/learning_role.py
```diff
@@ -251,6 +251,12 @@ async def store_to_buffer_and_update(self) -> None:
             # Remove last timestamp that has no reward yet
             timestamps_to_process = all_timestamps[:-1]
 
+            # Carry over the last timestamp's data (obs, actions, noises)
+            last_ts = all_timestamps[-1]
+            self.all_obs[last_ts] = current_obs[last_ts]
+            self.all_actions[last_ts] = current_actions[last_ts]
+            self.all_noises[last_ts] = current_noises[last_ts]
```
maurerle: This commit says "ensure consecutive buffer writing" - why is this needed? Can you describe this a bit more @mthede ?

pyproject.toml
maurerle: Weird - in the `pip install .[test]` you should only get pytest >=9

This was introduced with pytest 9 and landed assume here: https://github.com/assume-framework/assume/commit/c051d4522a6b2c6bdb94e56b13e8245d110d3346

maurerle: I wonder how you got pytest<9 in a fresh env

assume/reinforcement_learning/learning_role.py
kim-mskw: It is explained in the PR description

assume/reinforcement_learning/learning_utils.py
kim-mskw: I see, 
So the learning_role has this order from which it does the update calls. We could sort it alphabetically there as well, but then we have two sorting mechanisms instead of just using the original order stored in the `learning_role `on which also the extension of the NN relies when new agents are added for the critic. 

Why fix it on "the other side"

kim-mskw: @maurerle, could you clarify what you mean? I would like to close and merge this.

maurerle: Yes that's exactly what I meant.

> Why fix it on "the other side"

It would just preserve the order if you change something in the input file.
It does not matter that much - I'd just merge as is.

maurerle (approved):

assume/reinforcement_learning/learning_role.py
```diff
@@ -248,8 +248,20 @@ async def store_to_buffer_and_update(self) -> None:
         # Get timestamps from cache we took
         all_timestamps = sorted(current_obs.keys())
         if len(all_timestamps) > 1:
```
mthede, line 250: Not related, but just wondering: if there's only one observation, shouldn't we add that as well? so >= 1

mthede: Maybe that was the other error when I tried 24h bidding on a single day?

mthede: But maybe that would probably need an adjustment in the cache creation below, if the one timestamp is incomplete and timesteps_to_process is an empty list.

mthede (approved):

assume/reinforcement_learning/learning_role.py
kim-mskw: okay I think then the list operations do not work in general and we would need another way around this

PR #777, opened 2026-03-24
Building multiasset extension

Manish-Khanra:
## Related Issue
*If no issue exists, delete this line.*

## Description
This PR extends the building unit to support multi-asset configurations, with a particular focus on integrating multiple electric vehicles and charging stations into the building framework.

The main motivation is to move beyond the previous singleton-style setup in which only one asset of a given type could be connected to a building. With this change, buildings can now contain multiple EVs and multiple charging stations while preserving the existing building functionality for heat pumps, boilers, thermal storage, PV, and generic storage.

### Main changes
- Added support for prefix-based multi-asset handling in `Building`, allowing multiple components such as `electric_vehicle_1`, `electric_vehicle_2`, `charging_station_1`, etc.
- Added a dedicated `ChargingStation` DSM component in `dst_components.py` with:
  - unidirectional or bidirectional operation
  - optional availability profiles
  - ramp constraints
  - binary non-simultaneity for charging/discharging in bidirectional mode
- Extended the `ElectricVehicle` DSM component with:
  - availability-dependent operation
  - external range input
  - mobility-related energy usage
  - unidirectional or bidirectional power flow
  - binary non-simultaneity for charging/discharging
- Extended building connection logic:
  - if no charging station is present, EVs are connected directly to the building/grid balance
  - if charging stations are present, charging stations connect to the building/grid and EVs connect to charging stations through assignment variables
- Updated `DSMFlex.initialize_components()` to support prefixed DSM component names and to pass EV-specific external range data during model construction
- Updated `BuildingForecaster` and `loader_csv.py` to support:
  - aggregate building profiles such as `<building_id>_load_profile`
  - EV-specific availability and range profiles
  - charging-station-specific availability profiles
  - building-specific `electricity_price_flex` for use with `electricity_price_signal`
- Fixed switching between optimisation and flexibility modes in `DSMFlex.switch_to_opt()` so it no longer assumes that flexibility constraints exist for all flexibility measures
- Fixed variable-cost handling when flexible electricity-price signals are activated
- Added/updated release notes
- Expanded `tests/test_building.py` to cover both legacy building functionality and the new EV + charging-station integration logic

### Scope of testing
The updated building tests cover:
- existing building assets such as heat pumps, boilers, thermal storage, PV, and generic storage
- multiple EVs
- multiple charging stations
- EV availability and range-dependent usage
- EV state-of-charge balance
- charging-station availability and ramping
- EV-to-charging-station assignment consistency
- direct-grid fallback when no charging station is present
- total building power balance with and without charging stations

## Checklist

- Temporary CSV export utilities were used during development for internal validation of EV and charging-station operation, but they are not part of the formal test scope and are not intended as a user-facing feature.
- Existing example inputs may need updated forecast column names and component-specific profile columns to run with the new building multi-asset functionality.
- The release note entry for this change set was prepared as a minor feature release.


assume/units/building.py
```diff
@@ -103,12 +104,17 @@ def __init__(
 
         # check if the provided components are valid and do not contain any unknown components
         for component in components.keys():
-            if (
-                component not in self.required_technologies
-                and component not in self.optional_technologies
-            ):
+            base_component = next(
```
carl-wanninger: This snippet seems a little complicated.

If I understand the logic correctly, a formulation like the following would be more understandable:

```python
allowed = self.required_technologies + self.optional_technologies

for component in components:
    if component in allowed:
        continue

    if not any(component.startswith(tech) for tech in allowed):
        raise ValueError(
            f"Component '{component}' is not a valid component for the building unit."
        ) 
```

carl-wanninger: Furthermore, with the current formulation, a required technology will not be recognized as such if the component has a unit name attached. Is this intended?

assume/units/building.py
```diff
@@ -126,32 +132,99 @@ def __init__(
...
+
+        self.has_heatpump = len(self.heat_pumps) > 0
+        self.has_boiler = len(self.boilers) > 0
+        self.has_thermal_storage = len(self.thermal_storages) > 0
+        self.has_ev = len(self.evs) > 0
+        self.has_charging_station = len(self.charging_stations) > 0
+        self.has_battery_storage = len(self.battery_storages) > 0
+        self.has_pv = len(self.pv_plants) > 0
+
+        if self.has_ev and not self.has_charging_station:
```
carl-wanninger, line 153: What is the rationale behind allowing EVs without charging station? Would it not be simpler to render chargings stations mandatory for EVs?

assume/units/building.py
```diff
@@ -126,32 +132,99 @@ def __init__(
...
+            logger.warning(
+                "No charging station is included in the building. "
+                "Electric vehicles will be connected directly to the grid."
+            )
+            print(
```
carl-wanninger: I think to print a warning that has already logged is rather unusual.

assume/units/building.py
```diff
@@ -126,32 +132,99 @@ def __init__(
...
+            ev_config.pop("availability_profile", None)
+
+            profile_key_suffix = (
+                "charging_profile" if use_charging_profile else "availability_profile"
+            )
+            profile_key = f"{self.id}_{ev_key}_{profile_key_suffix}"
+
+            try:
+                ev_profile = self.forecaster[profile_key]
+            except Exception:
```
carl-wanninger: Wouldn't here a "KeyError" be more precise - or what kind of error is expected?

assume/units/building.py
```diff
@@ -126,32 +132,99 @@ def __init__(
...
+
+        # prepare EV-specific profiles
+        for ev_key in self.evs:
+            ev_config = self.components[ev_key]
+
+            use_charging_profile = str_to_bool(
+                ev_config.get("charging_profile", "false")
+            )
+            ev_config.pop("charging_profile", None)
+            ev_config.pop("availability_profile", None)
```
carl-wanninger: This confuses me a little: It is not decided yet whether to use charging profile or availability profile, yet both are removed from ev_config - why?

assume/units/building.py
```diff
@@ -126,32 +132,99 @@ def __init__(
...
+                if (
+                    hasattr(self.forecaster, "ev_load_profile")
+                    and self.forecaster.ev_load_profile is not None
+                ):
+                    ev_profile = self.forecaster.ev_load_profile
+                else:
+                    raise ValueError(
+                        f"Required profile '{profile_key}' not found in forecaster for EV '{ev_key}'."
+                    ) from None
+
```
carl-wanninger, line 190: I asked Gemini to simplify the code snippet above. This is what it came up with:

for ev_key in self.evs:
    ev_config = self.components[ev_key]
    
    # 1. Determine which profile type we need
    # We remove the keys from config while reading them to keep things tidy
    use_charging = str_to_bool(ev_config.pop("charging_profile", "false"))
    ev_config.pop("availability_profile", None) # Just discard this
    
    suffix = "charging_profile" if use_charging else "availability_profile"
    profile_key = f"{self.id}_{ev_key}_{suffix}"

    # 2. Try to get the specific profile, then the fallback, then fail
    ev_profile = self.forecaster.get(profile_key) or getattr(self.forecaster, "ev_load_profile", None)

    if ev_profile is None:
        raise ValueError(f"No profile found for {ev_key} (checked '{profile_key}' and generic fallback)")

Maybe this is helpful.

assume/units/building.py
```diff
@@ -126,32 +132,99 @@ def __init__(
...
+            ev_config[profile_key_suffix] = ev_profile
+
+            range_key = f"{self.id}_{ev_key}_range"
+            try:
+                ev_config["range"] = self.forecaster[range_key]
+            except Exception:
```
carl-wanninger: see above.

assume/common/forecaster.py
```diff
@@ -870,12 +860,31 @@ def __init__(
...
         self.load_profile = self._to_series(load_profile)
+        self.electricity_price_flex = self._to_series(electricity_price_flex)
+
+        # store arbitrary building-specific additional profiles
+        for key, value in kwargs.items():
+            if isinstance(value, pd.Series):
+                value = self._to_series(value)
+            self.__setattr__(key, value)
+
+    def __getitem__(self, key: str):
```
carl-wanninger, line 881: maybe add an optional default value?

assume/units/building.py
carl-wanninger: Note: This can be written most elegantly if self.forecaster has a getter that allows for default values.

assume/units/building.py
```diff
@@ -126,32 +132,99 @@ def __init__(
...
+                ):
+                    ev_profile = self.forecaster.ev_load_profile
+                else:
+                    raise ValueError(
+                        f"Required profile '{profile_key}' not found in forecaster for EV '{ev_key}'."
+                    ) from None
+
+            ev_config[profile_key_suffix] = ev_profile
+
+            range_key = f"{self.id}_{ev_key}_range"
```
carl-wanninger: What is the idea behind EV range? IMO this is not obvious.

assume/units/building.py
```diff
@@ -169,155 +242,223 @@ def define_parameters(self):
...
-    def define_variables(self):
-        """
-        Defines the decision variables for the Pyomo optimization model.
+        # EV external range parameters
+        for ev_key in self.evs:
+            ev_range = self.components[ev_key].get("range", None)
+            if ev_range is None:
+                continue
+
+            if hasattr(ev_range, "iloc"):
```
carl-wanninger: Why would ev_range have or not have "iloc"?

assume/units/building.py
```diff
@@ -169,155 +242,223 @@ def define_parameters(self):
...
+
+            self.model.add_component(
+                f"{ev_key}_availability",
+                pyo.Param(
+                    self.model.time_steps,
+                    initialize=profile_init,
+                    within=pyo.Binary,
+                    doc=f"Availability profile for {ev_key}",
+                ),
+            )
```
carl-wanninger, line 311: This code part is non-obvious to me and would profit from some explanation IMO.

The conversion of charging to availability profile could be more explicit.

assume/units/building.py
```diff
@@ -169,155 +242,223 @@ def define_parameters(self):
...
 
-        Both variables are defined over the `time_steps` set and are continuous real numbers.
-        """
+    def define_variables(self):
         self.model.total_power_input = pyo.Var(self.model.time_steps, within=pyo.Reals)
         self.model.variable_cost = pyo.Var(self.model.time_steps, within=pyo.Reals)
 
+        # only needed when charging stations are present
```
carl-wanninger: What is only needed?

assume/units/building.py
```diff
@@ -169,155 +242,223 @@ def define_parameters(self):
...
+                self.model.time_steps,
+                doc="Assignment only if EV is available",
+            )
+            def assignment_only_if_available(m, ev, cs, t):
+                ev_availability = getattr(m, f"{ev}_availability", None)
+                if ev_availability is not None:
+                    return m.is_assigned[ev, cs, t] <= ev_availability[t]
+                return pyo.Constraint.Skip
+
+        # ------------------------------------------------------------------
```
carl-wanninger, line 428: I am no expert in pyomo - but constraints seem logical and well described.

assume/units/building.py
```diff
@@ -169,155 +242,223 @@ def define_parameters(self):
...
-            if self.has_pv:
-                total_power -= m.dsm_blocks["pv_plant"].power[t]
-
-            # Assign the calculated total to the model's total_power_input for each time step
+            # electric loads from heat technologies
+            for hp in self.heat_pumps:
+                total_power += m.dsm_blocks[hp].power_in[t]
+
+            for boiler in self.boilers:
+                if hasattr(m.dsm_blocks[boiler], "power_in"):
```
carl-wanninger, line 440: why would a boiler have no "power_in"?

assume/units/building.py
```diff
@@ -169,155 +242,223 @@ def define_parameters(self):
...
+                total_power -= m.dsm_blocks[bat].discharge[t]
+
+            # PV
+            for pv in self.pv_plants:
+                total_power -= m.dsm_blocks[pv].power[t]
+
             return m.total_power_input[t] == total_power
 
     def define_constraints(self):
-        """
```
carl-wanninger, line 280: What is the rationale behind getting rid of these docstrings?

assume/units/dsm_load_shift.py
```diff
@@ -51,37 +51,49 @@ def initialize_components(self):
...
+        by resolving the base technology from the prefix.
         """
         components = self.components.copy()
         self.model.dsm_blocks = pyo.Block(list(components.keys()))
 
         for technology, component_data in components.items():
-            if technology in demand_side_technologies:
-                # Get the class from the dictionary mapping (adjust `demand_side_technologies` to hold classes)
-                component_class = demand_side_technologies[technology]
+            base_technology = next(
```
carl-wanninger, line 63: Gemini suggestion:

```python
for tech in demand_side_technologies:
    if technology.startswith(tech):
        base_technology = tech
        break
else:
    raise ValueError(f"Unknown DSM component technology: {technology}")
```

assume/units/dst_components.py
```diff
@@ -1390,50 +1390,73 @@ def operating_cost_constraint(b, t):
...
+        Max decrease in charge/discharge power between consecutive steps.
+    charging_profile : pd.Series | None
+        Optional fixed charging profile.
+    storage_loss_rate : float, default 0.0
+        Fractional storage loss per time step.
+    mileage : float | None
+        Specific electricity consumption for driving.
+        Interpreted here as energy per unit of external_range.
+        Example: if external_range is km, mileage should be MWh/km.
+    power_flow_directionality : str, default "unidirectional"
```
carl-wanninger, line 1436: I would find 
bidirectional: boolean | False 
simpler.

tests/test_building.py
```diff
@@ -6,763 +6,805 @@
 import pyomo.environ as pyo
```
carl-wanninger, line 6: To me it is a little confusing that there are ev tests in both test_building.py and test_ev.py .

carl-wanninger, line 6: I think test section would heavily profit from testing optimization.
That is: Here is a price signal now let us check that our building/EV actually finds the optimal solution. ( For various lengths).

assume/units/building.py
carl-wanninger: Is range the number of kilometers driven per timestep? 
If so: Would it not be simpler to model the power consumption directly?

carl-wanninger (changes requested):
This is a review draft. Ask for second opinion.

assume/scenario/loader_csv.py
```diff
@@ -646,6 +646,62 @@ def load_config_and_create_forecaster(
                     forecast_algorithms, unit
                 )
                 if type == "building":
+                    # Base aggregate building profiles
+                    building_load_profile = (
```
reinecfi: Maybe we could just use the following to be more concise:
`forecast_x = forecast_df.get(name, pd.Series(..., name))`

assume/units/dst_components.py
```diff
@@ -1450,76 +1473,315 @@ def __init__(
...
+
+            # Driving usage only when unavailable
+            if external_range is not None:
+                if self.mileage is None:
+                    raise ValueError(
+                        "`mileage` must be provided when `external_range` is used."
+                    )
 
-        # Apply predefined charging profile constraints if provided
+                @model_block.Constraint(self.time_steps)
```
reinecfi, line 1592: In our understanding, this constraint (no charging while driving and vice versa) should be handled in the input validation and not during runtime

assume/common/forecaster.py
Manish-Khanra: Thanks for the suggestion. I think it is better to keep __getitem__ strict here because the building logic already handles missing component profiles and with different semantics per component. A generic default would hide misconfigured or misspelled forecast keys and could silently change the optimization model.

assume/scenario/loader_csv.py
Manish-Khanra: Thanks, that makes sense. I have implemented it accordinly.

assume/units/building.py
Manish-Khanra: This is clearer. The logic is the same for our component naming scheme so I am happy to simplify it along these lines.

Thanks, that is a good catch. The current check is indeed too strict for our naming scheme since component keys may carry specific suffixes. I have talen care of it

assume/units/building.py
Manish-Khanra: We keep charging stations optional by design so the model can work at either level of detail. For simpler residential setups, it can represent a direct EV-to-building connection. For depots or fleet setup, it can instead assign EVs to specific charging stations when charger availability is essential. This keeps the simpler case free of unnecessary charger related parameters while also preserving backward compatibility.

assume/units/building.py
Manish-Khanra: The Boiler component can run on different fuels such as electricity, natural gas or hydrogen. Every boiler variant includes a power_in variable, but in practice only electric boilers use it. For gas and hydrogen fuel type, power_in is always fixed at zero, and the real energy input is represented through natural_gas_in or hydrogen_gas_in instead. So the current hasattr(..., "power_in") check is mainly there as a safeguard.

assume/units/building.py
Manish-Khanra: I have repased the old docstring with the new one becsue I thinnk the updated one can bring more elaboration to the users, in my perspective. What do you think about it?

assume/units/building.py
Manish-Khanra: Thanks for catching this. The redundant print statement has been removed

assume/units/building.py
Manish-Khanra: I have refactored the logic to move the decision point before the cleanup. Also Added a comment to explain the intent.

assume/units/building.py
Manish-Khanra: Thanks for the suggestion. The code now catches only KeyError for missing profile lookups. Alse a test have been added to verify.

assume/units/building.py
Manish-Khanra: I think with all the changes, the code is much simplified now. Please review it again.Thank you.

assume/units/building.py
Manish-Khanra: I have renamed "range" to "trip_distance" throughout the codebase to make the intent clearer. To give the user an option to simulate charging behavior demending on the distance covered and if not simply charging profile, is a design choice. So the model supports this, if a charging_profile is given, it is used as a fixed charging schedule and the trip distance calculation is skipped. The trip_distance + mileage approach is only needed when the user wants to model exogenous driving (e.g., for fleets with known routes) and let the optimizer decide charging.

assume/units/building.py
Manish-Khanra: goog point! I have now simplify the code by normalizing the input type.

assume/units/building.py
Manish-Khanra: I have included some comments and improved the readability

assume/units/building.py
Manish-Khanra: Thanks. I clarified the comment above this block to state what is only needed and why

carl-wanninger (changes requested):
Based on the provided tests and an additional test, the optimization seems to work as intended. Nice work!

We think two more things would be necessary for the code to be fully functional.

1. Right now I can have an EV with availability [1, 1] and trip_distances [80, 80]. This is physically not possible and should raise a validation error.

Either: 
Soft  validation constraint trip_distance > 0 -> avalability must be smaller than one. 
or:
Hard validation constraint: trip_distance > 0 -> availability must be 0.

2. We again strongly encourage that instead of trip_distance + mileage as input, the model should directly take used_energy as input. The reasoning is that if a dataset provides mileage and trip distance, it is very simple to calculate used energy. However, if a dataset only provides used energy (such as synpro from Fraunhofer ISE), one does have to fingate trip distance, which is very cumbersome.

In a future extension it might be nice to have soc pivots where the EV is requested to have a certain SoC at certain points of time - but this has not to be handled by this PR.

carl-wanninger:
Here is a test we constructed based on the provided tests to convince ourselves of the simultanoues arbitrage possiblity and trip distance handling.

def test_ev_able_to_make_arbitrage_and_meet_demand(time_index):
    # --- 1. Setup Data & Forecaster ---
    prices = [-10, 10, -10, 10, 100, -100, -100, 100]
    market_prices = {"EOM": _series(prices, time_index)}
    
    # EV 1: Available except hours 1 & 2. Trip in hours 1 & 2.
    avail_1 = [1, 0, 0, 1, 1, 1, 1, 1]
    dist_1  = [0, 40, 40, 0, 0, 0, 0, 0]
    
    # EV 4: Available except hours 2 & 3. Trip in hours 2 & 3.
    avail_2 = [1, 1, 0, 0, 1, 1, 1, 1]
    dist_2  = [0, 0, 40, 40, 0, 0, 0, 0]

    forecaster = BuildingForecaster(
        index=time_index,
        fuel_prices={"natural_gas": _series([20] * 8, time_index)},
        market_prices=market_prices,
        electricity_price_flex=_series(prices, time_index),
        A360_electric_vehicle_1_availability_profile=_series(avail_1, time_index),
        A360_electric_vehicle_1_trip_distance=_series(dist_1, time_index),
        A360_electric_vehicle_2_availability_profile=_series(avail_2, time_index),
        A360_electric_vehicle_2_trip_distance=_series(dist_2, time_index),
    )

    # --- 2. Define Components ---
    ev_config = {
        "capacity": 80.0,
        "min_soc": 0.0,
        "max_soc": 1.0,
        "max_power_charge": 11.0,
        "max_power_discharge": 11.0,
        "initial_soc": 0.7,
        "efficiency_charge": 0.95,
        "efficiency_discharge": 0.95,
        "ramp_up": 11.0,
        "ramp_down": 11.0,
        "storage_loss_rate": 0.0,
        "mileage": 0.5,
        "power_flow_directionality": "bidirectional",
    }

    components = {
        "electric_vehicle_1": ev_config.copy(),
        "electric_vehicle_2": ev_config.copy()
    }

    # --- 3. Solve ---
    building = _make_building(forecaster, components, prosumer="Yes")
    building, instance, _ = _solve_building_opt(building)

    # --- 4. Extractions & Assertions ---
    for ev_id in ["electric_vehicle_1", "electric_vehicle_2"]:
        ev_block = instance.dsm_blocks[ev_id]
        
        charge = [_val(ev_block.charge[t]) for t in range(8)]
        discharge = [_val(ev_block.discharge[t]) for t in range(8)]
        usage = [_val(ev_block.usage[t]) for t in range(8)]
        soc = [_val(ev_block.soc[t]) for t in range(8)]

    # 1. Arbitrage at Peak: Should discharge at t=4 (Price 100) 
    assert discharge[4] > 0, "EV should discharge at t=4 to take advantage of the 100 price"

    # 2. Maximum Charging at Bottom: Should charge heavily at t=5 or t=6 (Price -100)
    assert sum(charge[5:7]) > sum(charge[0:2]), \
        "EV should prioritize charging at t=5,6 over the milder negative prices at t=0,2"
    
    # 3. Final Peak: Should discharge again at t=7 (Price 100)
    assert discharge[7] > 0, "EV should discharge at final price peak (t=7)"

    # 4. Trip Energy: Meet trip energy needs at t=2
    assert usage[2] == 40 * 0.5, "EV should meet trip distance at t=2"

Manish-Khanra:
> Based on the provided tests and an additional test, the optimization seems to work as intended. Nice work!
> 
> We think two more things would be necessary for the code to be fully functional.
> 
>     1. Right now I can have an EV with availability [1, 1] and trip_distances [80, 80]. This is physically not possible and should raise a validation error.
> 
> 
> Either: Soft validation constraint trip_distance > 0 -> avalability must be smaller than one. or: Hard validation constraint: trip_distance > 0 -> availability must be 0.
> 
>     2. We again strongly encourage that instead of trip_distance + mileage as input, the model should directly take used_energy as input. The reasoning is that if a dataset provides mileage and trip distance, it is very simple to calculate used energy. However, if a dataset only provides used energy (such as synpro from Fraunhofer ISE), one does have to fingate trip distance, which is very cumbersome.
> 
> 
> In a future extension it might be nice to have soc pivots where the EV is requested to have a certain SoC at certain points of time - but this has not to be handled by this PR.

@carl-wanninger Included both the requests.

carl-wanninger (approved):
We approve this PR.
Thank you for your good work, Manish!

maurerle:
This PR added Release notes in a section of 0.6.1 which does not exist (yet). The latest release is still 0.6.0.
This is similar to #845 which added a wrong 0.6.2 instead of adding the changes to "upcoming changes".

PR #778, opened 2026-03-24
raise ValidationError when forecaster validation fails

jrasko:
use validation error to indicate wrong forecaster values. Make id and field values optional since they are not always provided

### DIFF
--- a/assume/common/exceptions.py
+++ b/assume/common/exceptions.py
@@ -10,7 +10,7 @@ class AssumeException(Exception):
 
 
 class ValidationError(ValueError):
-    def __init__(self, message: str, id: str, field: str):
+    def __init__(self, message: str, id: str = None, field: str = None):
         super().__init__(message)
         self.field = field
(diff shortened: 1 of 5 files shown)


assume/common/exceptions.py
```diff
@@ -10,7 +10,7 @@ class AssumeException(Exception):
 
 
 class ValidationError(ValueError):
-    def __init__(self, message: str, id: str, field: str):
+    def __init__(self, message: str, id: str = None, field: str = None):
```
maurerle, line 13: Which information does this give you, if it is None..?
What Information to highlight in the UI then? What is changed to the current behavior?

This is fine though - just add type hints as well :)

jrasko: Forecasters do not know the id of their units, so my idea is to catch and re-throw this exception with the missing knowledge. _field_ does not need to be None however since this should always be known

PR #781, opened 2026-03-31
Fix query failure on examples with DRL - if no run before

### LINKED ISSUE #780: Select queries on 'market_meta' and 'market_dispatch' fail in DRL examples (02a–02e) due to missing tables
Running the DRL examples (02a–02e) with the `timescale` data format causes select queries on the tables `market_meta` and `market_dispatch` to fail because the tables are not created.

**Steps to reproduce:**
Remove database directory
Start docker containers to initialize a fresh database
Set `'data_format' = "timescale"` and `example = "small_learning_1"` in [examples.py](https://github.com/assume-framework/assume/blob/main/examples/examples.py)
Run examples.py

**Expected behavior:**
The required tables (`market_meta`, `market_dispatch`) should be created automatically, and queries should execute successfully.

**Additional context:**
Running a non-DRL example beforehand resolves the issue, as those examples correctly create the required tables.

ChriKo97:
## Related Issue
Closes #780 

## Description
Sets agent output adress in [world.py](https://github.com/assume-framework/assume/blob/dda21992089655703426b973d84ad17aea4f5a7c/assume/world.py#L452) regardless of mode.

## Checklist

All checks pass

### DIFF
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -458,6 +458,11 @@ class WriteOutput(Role):
         for table, data_list in self.write_buffers.items():
             if len(data_list) == 0:
                 continue
+
+            # dont save market orders & dispatch for learning mode
+            if table in ["market_orders", "market_dispatch"] and (self.learning_mode or self.evaluation_mode):
+                continue
+
             df = None
             with self.locks[table]:
                 if table == "grid_topology":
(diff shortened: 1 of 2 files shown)


kim-mskw:
If I understand the issue correctly, it just appears if the very first simulation you run is a learning simulation because the tabels do not exist. I would suggest that the output agent creates empty tables instead of always changing the output agent address handling.  @ChriKo97, please correct me if I understood you wrong.

maurerle:
The idea of the existing code is to disable writing the bids of training runs to the database, to reduce the amount of data stored there, as we are typically not interested in the individual market results of a single training episode.

This PR would save all this data in the DB as well.

creating the tables beforehand is one option to fix this.
As we typically do not know all columns for sure, this is still not the best though..

Another option is to disable the KPI calculation if learning mode is enabled to not run into this error :/

ChriKo97:
The lastest commit [7a1df65 ](https://github.com/assume-framework/assume/pull/781/commits/7a1df6522f40e36720252980aca5502de4d00548) implements changes in query behaviour for the KPI definitions when learning mode is enabled, so that non-existent tables will not be queried. Also, on learning mode nothing will be stored in `market_meta` or `market_dispatch` tables :)

assume/common/outputs.py
```diff
@@ -458,6 +458,11 @@ async def store_dfs(self):
         for table, data_list in self.write_buffers.items():
             if len(data_list) == 0:
                 continue
+
+            # dont save market orders & dispatch for learning mode
+            if table in ["market_orders", "market_dispatch"] and (self.learning_mode or self.evaluation_mode):
```
kim-mskw, line 463: I fear this will lead both tables in the buffer to terribly overflow, as `["market_orders", "market_dispatch"]` are still present in the buffer but will never be cleared, correct? Please double-check

assume/world.py
```diff
@@ -449,7 +449,7 @@ def add_unit_operator(
         self.unit_operators[id] = units_operator
 
         # after creation of an agent - we set additional context params
-        if not self.learning_mode:
+        if self.output_agent_addr:
```
kim-mskw, line 452: I agree that "just not giving an output agent address" is a weird way of handling not writing the market results during training. However, I would like to avoid any unnecessary message sending, like for data that we do not want to write anyhow during the learning, which is everything besides the rl-params, which will be written by the `learning_role`. 

How about we use a different flag here and make this consistent for the market as well?

PR #792, opened 2026-04-15
Solver fixes

### LINKED ISSUE #785: Solver instance is created on every single call in complex_clearing
The problem is that a new solver is created via SolverFactory on every single call to market_clearing_opt in complex_clearing.py. For a year-long simulation with hourly clearing, thats too many solver instances being created and not being cleaned up.

Happens both with gurobi (file descriptor leak) and with highs (too many open files)
[insert exact gurobi error message]
[insert exact highs error message]

Possible solutions: create single solver instance in function define_solver(), store it as self.solver_instance, and reuse it by passing a new solver_instance parameter to market_clearing_opt()

This should be the same for redispatch and nodal_clearing.

gugrimm:
## Related Issue
Closes #785 and #795

## Description
This includes 4 changes:
- in forecast_algorithms, I fixed the bids generated without bid_id to contain bid_ids and cleaned the orderbook from empty bids (empty bids being transferred to complex clearing let the optimization fail, see #795 ). Also, I moved the inelastic_price_bid calculation, to avoid an error if no inelastic_demand exists.
- A check for available solvers in redispatch and nodal_clearing is added. It slightly differs from the one used in complex_clearing, as it uses linopy. I think grid_utils.py is the place for this "get_supported_solvers_linopy" check. I therefore had to rename "get_supported_solvers" to "get_supported_solvers_pyomo". 
- use "solver_name" for the string containing the name of the solver to be used, for example "gurobi" and distinguish it from "solver", which from now on is the actual solver object created by e.g. SolverFactory. This helps to better understand when we refer to what and is consistent with e.g. PyPSA. I added a __post_init__() method in the MarketConfig to check for "solver" in the config, rename it to "solver_name" and issue a deprecated warning.
- in complex_clearing: create solver instance (self.solver) once in __init__() by calling self.solver = SolverFactory(self.solver_name) and then reuse this solver instance in market_clearing_opt (see #785)

## Checklist


kim-mskw (commented):
Looking good! I quickly got rid of the double book keeping of the solvers, does this make sense @gugrimm?

assume/common/market_objects.py
```diff
@@ -153,6 +154,20 @@ def __hash__(self):
...
         )
 
+    def __post_init__(self):
+        """
+        Post-initialization checks for deprecated key "solver" in param_dict.
+        """
+
+        if "solver" in self.param_dict:
+            warn(
+                "The key 'solver' in 'param_dict' is deprecated and may be removed in future versions. "
```
maurerle, line 164: missing mention of "rename it to solver_name"

tests/test_grid_utils.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: ASSUME Developers
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import pytest
+
+try:
+    from assume.common.grid_utils import get_supported_solver_linopy
```
maurerle, line 8: `from pytest import importorskip`

kim-mskw:
@gugrimm, the comments of @maurerle are all addressed. If you think it is merge-ready as well, go ahead.

kim-mskw (approved):

PR #794, opened 2026-04-15
fix portfolio_learning_strategies

### LINKED ISSUE #784: missing function 'min_max_rescale' in portfolio_learning_strategies.py
Due to missing 'min_max_rescale' function in utils.py, an import error is thrown in portfolio_learning_strategies.py

https://github.com/assume-framework/assume/blob/d48b152b7374188bd864167beac572cc8750f2ba/assume/strategies/portfolio_learning_strategies.py#L14

@chiara-fb: 
- could you add the missing function to utils.py? 
- and maybe add a test for the RL portfolio strategy

Thanks!

maurerle:
## Related Issue
Closes #784 

## Description
**Improvements:**
   - **Add test for portfolio learning strategy and ``min_max_scale`` function**: Added a test for the portfolio learning strategy and ``min_max_rescale`` function to ensure its functionality and stability.

**Bug Fixes:**
  - **Fix errors in portfolio learning strategies**: The ``min_max_rescale`` function was missing from ``utils.py``, causing an ``ImportError`` in ``portfolio_learning_strategies.py``. Resolved by extending ``min_max_scale`` to cover the rescaling use case. And fix minor construction bug for observation space.

## Checklist


chiara-fb (commented):
Consider substituting [line 453](https://github.com/assume-framework/assume/blob/b3355675b898b41eeaa93fec2d9df3d2fc99e5b7/assume/strategies/learning_strategies.py#L453) of EnergyLearningStrategy from:
`bid_prices = actions * self.max_bid_price`
to:
`bid_prices = min_max_scale(actions, -1,1, self.min_bid_price, self.max_bid_price)`

to allow for more flexibility.

assume/strategies/learning_strategies.py
```diff
@@ -450,7 +450,9 @@ def calculate_bids(
         # =============================================================================
         # actions are in the range [-1,1], we need to transform them into actual bids
         # we can use our domain knowledge to guide the bid formulation
-        bid_prices = actions * self.max_bid_price
+        bid_prices = min_max_scale(
```
kim-mskw: If we want to change this, we need to adjust the intial exploration as well, because it now relies on the fact that it is symmetric around zero see: 

```
                marginal_cost = next_observation[
                    -1
                ].detach()  # ensure no gradients flow through
                # Add marginal cost to the action directly for initial random exploration
                curr_action += marginal_cost
```

kim-mskw: and because both are scaled with max_bid_price only beforehand

kim-mskw (dismissed):
This will break inital exploration mode

assume/strategies/learning_strategies.py
chiara-fb: correct. This is not the case in the portfolio strategy, which uses the current load status to proxy for the level of price markup that the operator can apply, [see](https://github.com/assume-framework/assume/blob/d65f614b6fca4dd9481b0158b6e5506d0d0da3c5/assume/strategies/portfolio_learning_strategies.py#L261): 

`curr_action = 2 * next_observation[2] - 1 + noise`

Possibly the initial experience could be shifted to be centered around `(max_bid_price - min_bid_price) / 2 + noise`?  That would make it independent from marginal costs though ...

mthede: Yeah, you're completely right @kim-mskw, I missed that. 

To achieve sensible exploration around the marginal costs, we could change the scaled mc in the observation space to the [min_bid_price, max_bid_price range]?

mthede: Changing the action space scaling for the other RL strategies needs further thinking. My proposition: revert it for normal RL Strategies and push the rest regarding the portfolio learning strategy. Ok @chiara-fb, @kim-mskw?

maurerle:
We will merge the initial PR to fix the missing function and later work on the improvements @mthede

mthede (approved):

assume/strategies/learning_strategies.py
kim-mskw: Sounds good and looks good

PR #801, opened 2026-04-24
added forecast saving option for config, adjusted forecast tests

### LINKED ISSUE #797: Include option to save forecast and reuse it
We once had a method to save forecasts to csv, but it was never used and removed in #724 
As forecasts may get more sophisticated and take longer to calculate in the near future, it would be good to have a saving option.
@reinecfi and my suggestion would be to include a "use_existing" or similar keyword in the forecast config. The default would be to not save and reuse, but can be activated. We would prefer not to have another user decision (user input), as this already happens when existing policies are overwritten and seems to confusing for users in our view.

### LINKED ISSUE #800: Price forecast with elastic demand has issue
Several things regarding the price forecast with elastic demands need improvement:

1. There is a bug due to missing bid_ids in the clearing
2. The forecaster fails if there is no inelastic demand present
3. The unit test for the forecast should be redone as it did not catch any of this!

reinecfi:
## Related Issue
Closes #797
Closes #800  (other parts fixed by PR #792)

## Description

This change adds the possibility to save forecasts and improves forecaster tests.

## Checklist


assume/common/forecast_algorithms.py
```diff
@@ -381,7 +381,10 @@ def calculate_naive_residual_load(
     config: MarketConfig,
     preprocess_information=None,
 ) -> dict[str, ForecastSeries]:
-    """Compute residual load as total demand minus renewable generation for each timestep."""
+    """Compute residual load as total demand minus renewable generation for each timestep.
+
+    NOTE: Elastic demands are not supported currently
+    """
```
mthede: What do you mean with "are not supported currently"? 
Are those demands not included at all or are those demands approximated as inelastic demands? When using it I'd want to know wheter the naive residual load forecast is rather under- or overestimated

assume/common/forecast_algorithms.py
```diff
@@ -441,6 +447,7 @@ def calculate_naive_congestion_signal(
             across all connected lines as the node's congestion signal.
 
     Returns an empty dict if grid data (buses/lines) is unavailable.
+    NOTE: Elastic demands are not supported currently
```
mthede: Same here, see above

assume/common/forecast_algorithms.py
```diff
@@ -461,6 +468,10 @@ def calculate_naive_congestion_signal(
         )
         return {}
 
+    # TODO: Add support for elastic demand (currently error of only elastic demands)
```
mthede: Is an error thrown? How would I figure out what's wrong and how to fix it?

docs/source/unit_forecasts.rst
```diff
@@ -10,6 +10,8 @@ Each unit in ASSUME has an associated **forecaster** that provides it with stati
 availability, fuel prices, or market price predictions. Forecast data is typically provided as time
 series, either loaded from a CSV file or calculated internally from simulation data.
 
+Forecasters can generally be safed via config parameter ``save_forecasts`` to the specified file given by: ``forecast_save_file`` (default: SCENARIO_PATH/saved_forecasts.csv)
```
mthede, line 13: Document usage of "use_forecasts_df" as well? Maybe clearify that provided forecasts will be used and missing, but expected, forecasts are calculated and then combined (if I understood correctly).

mthede: Ah, saw the other part of the documentation below. Never mind.

mthede (commented):
Hi Finn, thanks for the nice feature of saving the forecasts! That's a great way to check whether the intended values were provided to the units. 

I was at first confused by the naming and defaults of the config parameters. Then I realized that we don't really have full documentation of the top-level config options, other than in [Notebook 02](https://github.com/assume-framework/assume/blob/4887c111a7389d4508db47e991a2f349352d65ea/examples/notebooks/02_automated_run_example.ipynb).

Maybe it's worth documenting the behavior at one place and link to the specific forecaster details in the docs?

Other than those clarifications and documentations, it looks good :)

assume/common/forecast_algorithms.py
reinecfi: I just forgot to remove the TODO here. The if clause circumvents the problem. Currently you are not notified. Maybe one could/should add a warning here?

assume/common/forecast_algorithms.py
reinecfi: Adjusted the wording here. The forecast will underestimate the residual load if elastic demand is present.

assume/common/forecast_algorithms.py
reinecfi: More or less like above, but since some demand is ignored, the forecast should be smaller (which can lead to a smaller or bigger congestion in absolute terms, depending whether the net load is positive or negative).
Not sure if adding a comment on this will be more confusing than helpful. Especially as this is a naive forecast.

gugrimm (approved):

PR #810, opened 2026-05-25
fix(runtime): reduce memory growth in long-running multi-simulation workloads

isaac-gumbrell:
## Description
This PR addresses several likely contributors to memory growth observed when running many simulations concurrently over long periods.

During sustained multi-run workloads, memory usage increases significantly over time even after the replay buffer is fully populated. This PR patches a number of suspected retention points and cache / object-lifecycle issues to reduce unnecessary memory accumulation.

Changes in this PR:
- clean up several likely sources of retained objects and long-lived references
- reduce memory growth in long-running and parallel workflows
- preserve existing functionality and expected model behaviour

## Checklist

## Additional Notes
These combinations of changes have entirely eliminated memory creep without altering results.

### DIFF
--- a/assume/common/outputs.py
+++ b/assume/common/outputs.py
@@ -658,6 +658,13 @@ class WriteOutput(Role):
         # insert left records into db
         await self.store_dfs()
 
+        # Dispose the SQLAlchemy engine to release pooled connections
+        if self.db is not None:
+            try:
+                self.db.dispose()
+            except Exception:
+                pass
+
         if self.db is None:
             return
 
--- a/assume/common/units_operator.py
+++ b/assume/common/units_operator.py
@@ -234,7 +234,9 @@ class UnitsOperator(Role):
             order["market_id"] = content["market_id"]
 
         marketconfig = self.registered_markets[content["market_id"]]
-        self.valid_orders[marketconfig.product_type].extend(orderbook)
+        # Only keep accepted orders for dispatch tracking —
+        # rejected orders are not needed and cause unbounded growth.
+        self.valid_orders[marketconfig.product_type].extend(accepted_orders)
         self.set_unit_dispatch(orderbook, marketconfig)
         self.write_actual_dispatch(marketconfig.product_type)
 
--- a/assume/common/utils.py
+++ b/assume/common/utils.py
@@ -737,16 +737,52 @@ def parse_duration(duration_str):
 
 def calculate_content_size(content: list | dict) -> int:
     """
-    Calculate the size of a content in bytes.
+    Calculate the deep memory size of content in bytes.
+
+    Uses recursive traversal to account for nested structures,
+    DataFrames, numpy arrays, and torch tensors — unlike sys.getsizeof
+    which only measures the shallow container size.
     """
-    if isinstance(content, dict):  # For dictionaries
+    try:
+        import pandas as pd
+
+        if isinstance(content, pd.DataFrame):
+            return int(content.memory_usage(deep=True).sum())
+        if isinstance(content, pd.Series):
+            return int(content.memory_usage(deep=True))
+    except Exception:
+        pass
+
+    try:
+        import numpy as np
+
+        if isinstance(content, np.ndarray):
+            return content.nbytes
+    except Exception:
+        pass
+
+    try:
+        import torch
+
+        if isinstance(content, torch.Tensor):
+            return content.nelement() * content.element_size()
+    except Exception:
+        pass
+
+    if isinstance(content, dict):
         return sys.getsizeof(content) + sum(
-            sys.getsizeof(value) for value in content.values()
+            calculate_content_size(k) + calculate_content_size(v)
+            for k, v in content.items()
         )
-    elif isinstance(content, list):  # For lists, including lists of dicts
+    elif isinstance(content, (list, tuple)):
         return sys.getsizeof(content) + sum(
             calculate_content_size(item) for item in content
         )
+    elif isinstance(content, str):
+        return sys.getsizeof(content)
+    elif isinstance(content, (int, float, bool, type(None))):
+        return sys.getsizeof(content)
+    # Fallback for unknown types
     return sys.getsizeof(content)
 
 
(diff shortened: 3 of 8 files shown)


assume/common/outputs.py
```diff
@@ -658,6 +658,13 @@ async def on_stop(self):
         # insert left records into db
         await self.store_dfs()
 
+        # Dispose the SQLAlchemy engine to release pooled connections
+        if self.db is not None:
+            try:
+                self.db.dispose()
+            except Exception:
+                pass
```
maurerle, line 666: I would not fail silently here but still raise or log the error.

Besides, after on_stop, the `__del__` function should be called, which the reference count of self.db to 0, which immediately should call the objects `__del__` function, terminate the active db connections and not create memory leaks.
Did you actively encouter issues with this?

assume/common/units_operator.py
```diff
@@ -234,7 +234,9 @@ def handle_market_feedback(self, content: ClearingMessage, meta: MetaDict) -> No
             order["market_id"] = content["market_id"]
 
         marketconfig = self.registered_markets[content["market_id"]]
-        self.valid_orders[marketconfig.product_type].extend(orderbook)
+        # Only keep accepted orders for dispatch tracking —
+        # rejected orders are not needed and cause unbounded growth.
```
maurerle, line 238: I remember that there was something about this, but it looks more right now.

The valid_orders are not needed for dispatch tracking though, but are kept so that the agent has information about all past orders.

So please adjust/remove the comment.
It is better to describe this in the commit message instead.

assume/common/utils.py
```diff
@@ -737,16 +737,52 @@ def parse_duration(duration_str):
 
 def calculate_content_size(content: list | dict) -> int:
     """
-    Calculate the size of a content in bytes.
+    Calculate the deep memory size of content in bytes.
+
+    Uses recursive traversal to account for nested structures,
+    DataFrames, numpy arrays, and torch tensors — unlike sys.getsizeof
+    which only measures the shallow container size.
```
maurerle, line 744: Is the shallow size different? If so, when? Can we get a test for this?

In my short try-out there were no bigger differences between those two

assume/markets/base_market.py
```diff
@@ -670,7 +670,7 @@ async def clear_market(self, market_products: list[MarketProduct]):
                         0,
                     )
 
-        self.open_auctions - set(market_products)
+        self.open_auctions -= set(market_products)
```
maurerle, line 673: 👍

assume/markets/base_market.py
```diff
@@ -725,6 +725,12 @@ async def clear_market(self, market_products: list[MarketProduct]):
 
         await self.store_market_results(market_meta)
 
+        # Prevent unbounded growth of in-memory results.
+        # Keep only results that data-request handlers may still need
+        # (those whose product_start has not yet passed).
+        now = timestamp2datetime(self.context.current_timestamp)
+        self.results = [r for r in self.results if r.get("product_start", now) >= now]
```
maurerle, line 732: Currently, the `self.results` in the base_market are there to be available for future market mechanisms, and for inspection at runtime.
The results only contain the meta data of the market (cleared volume and so on).
I don't think that this has a large impact on simulation resources..?

Though this is a relevant point here.

assume/reinforcement_learning/learning_role.py
```diff
@@ -616,6 +616,14 @@ def init_logging(
             train_start (str): The start time of simulation.
         """
 
+        # Explicitly close the previous logger to release its DB engine
+        # and SummaryWriter file handles before creating a new one.
+        if (
+            hasattr(self, "tensor_board_logger")
```
maurerle, line 622: self.tensor_board_logger is initialized with None, so the hasattr can be removed and is just slop at this point

assume/reinforcement_learning/tensorboard_logger.py
```diff
@@ -363,10 +363,28 @@ def update_tensorboard(self):
...
+                self.writer.flush()
+                self.writer.close()
+            except Exception:
+                pass
+            self.writer = None
+        if hasattr(self, "db") and self.db is not None:
+            try:
+                self.db.dispose()
+            except Exception:
+                pass
```
maurerle, line 383: silently failing is unclean
I would still keep the close function content in `__del__`

assume/reinforcement_learning/learning_role.py
```diff
@@ -616,6 +616,14 @@ def init_logging(
...
+            and self.tensor_board_logger is not None
+        ):
+            self.tensor_board_logger.close()
```
maurerle, line 625: reassigning the tensor_board_logger, sets it reference count to 0, which immediately should call the objects `__del__` function and not create memory leaks.
Did you actively encouter issues with this?

```python
>>> class Tomate:
...     def __init__(self):
...         print('Tomate created')
... 
...     def __del__(self):
...         print('Tomate destroyed')
... 
... 
... a = Tomate()   # 1st instance
... a = Tomate()   # 2nd instance, overwriting the name 'a'
... 
Tomate created
Tomate created
Tomate destroyed
```

assume/world.py
```diff
@@ -874,6 +875,26 @@ def reset(self):
...
+        )
+
+        sort_units.cache_clear()
+        calculate_naive_price_inelastic.cache_clear()
+        calculate_naive_price_elastic.cache_clear()
+        calculate_naive_price.cache_clear()
+        calculate_naive_residual_load.cache_clear()
+        calculate_naive_congestion_signal.cache_clear()
+        calculate_naive_renewable_utilisation.cache_clear()
+
```
maurerle, line 897: The `lru_cache` only keeps up to `maxsize=128` elements. This does not increase in size. I would spare the complexity of keeping track of the low-level LRU_Cache in the world ourselves..?

maurerle (changes requested):
Hi @isaac-gumbrell - thanks for this PR, which includes a lot of well intended improvements.

I added some comments and remarks to the optimizations.
In general, I don't think that there is a lot to gain from doing memory management on top of pythons own memory management cycle.

kim-mskw:
Potentially reloading the world makes it slower; double-check that. Discussion with @isaac-gumbrell .

maurerle:
I'd like to close this as the valid findings were addressed in #827 and there is not much else.
Any obligations?

PR #812, opened 2026-05-25
feat(clearing): support directional transfer capacities in complex clearing

isaac-gumbrell:
## Description
This PR adds directional transfer-capacity support to `complex_clearing`.

Previously, transfer limits were treated symmetrically. This change allows users to specify different forward and reverse capacities and applies those asymmetric limits during complex market clearing. This is useful for zonal studies where commercial transfer limits differ by direction.

If directional capacity columns are not provided, the existing symmetric-capacity behaviour is retained.

Changes in this PR:
- add directional line-capacity aggregation
- integrate directional constraints into `complex_clearing`
- add an example scenario demonstrating directional capacities
- document the new behaviour in the market-mechanism docs

## Checklist


assume/common/utils.py
```diff
@@ -550,6 +550,130 @@ def create_incidence_matrix(lines, buses, zones_id=None):
     return incidence_matrix
 
 
+def aggregate_line_capacities(
```
gugrimm, line 553: I have a small problem with the name of this function. It sums the capacities. This addition is only valid if we consider a transport model, not if we consider a powerflow. The function is intended to be used within the complex clearing, which uses a transport model and which is therefore ok. But I would appreciate a way to rename this function in a way that makes clear that it is not intended to be used for aggregation of lines for an PF. What about "sum_line_capacities"?

assume/markets/clearing_algorithms/complex_clearing.py
```diff
@@ -184,17 +189,27 @@ def energy_balance_rule(model, node, t):
...
-                capacity = lines.at[line, "s_nom"] * s_max_pu
-                # Limit the flow on each line
-                model.transmission_constr.add(model.flows[t, line] <= capacity)
-                model.transmission_constr.add(model.flows[t, line] >= -capacity)
+                # If precomputed directional capacities are provided, use them
+                if (
+                    directional_capacities is not None
+                    and line in directional_capacities.index
+                ):
+                    cap_forward = directional_capacities.at[line, "cap_forward"]
```
gugrimm, line 197: for consistency I would multiply by s_max_pu here too. s_max_pu defaults to 1. But if people explicitly use it (e.g. setting to 0.7 to approximate n-1 security or setting to high values to mimic unconstrained networks without needing to adjust s_nom), it should not only apply to non-directional capacities but also to directional ones.

assume/markets/clearing_algorithms/complex_clearing.py
```diff
@@ -184,17 +189,27 @@ def energy_balance_rule(model, node, t):
...
+                    cap_reverse = directional_capacities.at[line, "cap_reverse"]
```
gugrimm, line 198: same here: multiply by `s_max_pu`

docs/source/market_mechanism.rst
```diff
@@ -79,7 +79,13 @@ is given by: :math:`\mathbf{a}_{c, p} \: u_c \leq u_{p} \quad \forall \: c, p \i
...
 with the incidence matrix :math:`\mathbf{a}_{c, p}` defining the links between bids as 1, if c is linked as child to p, 0 else.
 
-Flows in the network are limited by the Net Transfer Capacity ('s_nom' * 's_max_pu') of each line l: :math:`\quad -NTC_{l} \leq F_{l, t} \leq NTC_{l} \quad \forall \: l \in \mathcal{L}, t \in \mathcal{T}`,
+Flows in the network are limited by the Net Transfer Capacity of each line :math:`l`. By default ASSUME derives a symmetric limit from ``s_nom * s_max_pu`` and applies it in both directions:
+
+.. math::
+
+  -NTC_{l} \leq F_{l, t} \leq NTC_{l} \quad \forall \: l \in \mathcal{L}, t \in \mathcal{T}
+
+If ``lines.csv`` additionally provides ``s_nom_forward`` and/or ``s_nom_reverse``, the complex clearing uses these directional limits instead. This allows transport constraints to differ by flow direction, which is useful for zonal representations with asymmetric commercial transfer capacities.
```
gugrimm, line 88: add sth like: `s_max_pu` is still applied.

docs/source/release_notes.rst
```diff
@@ -15,6 +15,7 @@ Upcoming Release
 
 **New Features:**
   - **Generic Forecasting Interface**: This interface enables to specify different forecast algorithms for preprocess, initialization and update during runtime. They can be specified in the config.yaml or unit csv files. For more information about currently implemented algorithms and how to specify them please read the documentation on Unit forecasts.
+  - **Directional transfer capacities in complex clearing**: ``complex_clearing`` can now use asymmetric line limits from ``s_nom_forward`` and ``s_nom_reverse`` instead of assuming the same transfer capacity in both directions.
```
gugrimm, line 18: add sth like: s_max_pu is still applied.

gugrimm (commented):
Thank you for this useful initiative! I don't have any fundamental objection. The development also does not block the time dependent s_max_pu (dynamic line rating) that we would potentially like to introduce in future. But I have three comments:
- I think `s_max_pu` should also apply to the directional limits
- the aggregation function (sum) is only correct in a transport model, not in a PF. So I would like to be sure users do not mis-use or mis-interpret the function as a way of clustering /simplyfying  networks. Which is what I could think it is, if I hear "aggregation".
- tests are missing

gugrimm:
I added some tests with support by LLM. Could you have a look?
The one with s_max_pu still fails because we do not consider it yet, but as mentioned, I think we should.

PR #815, opened 2026-06-03
Avoid torch seeding in non-RL simulations (+ minor general typos + default example fixed)

mthede:
## Description
- Skip torch seeding when torch is installed but not used: Irrelevant seeding was performed and a warning was thrown about deterministic PyTorch behavior, even though simulation does not use RL. This is fixed by only setting the PyTorch seeds when learning is active.
- While at it, fixed the default example, some typos and minor rephrasing for clarity

## Checklist


reinecfi (approved):
Looks reasonable for me. Not sure why the tests fail.

PR #834, opened 2026-07-08
ci: update config to be less verbose

maurerle:
### **User description**
Trying to make the pr agent less verbose from the feedback of the dev meeting.

### DIFF
--- a/.github/workflows/pr-agent.yml
+++ b/.github/workflows/pr-agent.yml
@@ -17,12 +17,17 @@ jobs:
       contents: write
     steps:
       - name: PR Agent action step
-        uses: the-pr-agent/pr-agent@v0.36.1
+        uses: the-pr-agent/pr-agent@v0.39.0
         env:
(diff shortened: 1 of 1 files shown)


reinecfi (approved):
A little less verbosity sound reasonable.

maurerle:
somehow the label creation is lost now - hm. 
For now the PR-Agent does not allow all choices what to have and what not :(

`config.output_relevant_configurations` shows the existing config, but not which model **was** used.
i created two issues for pr-agent: 

* https://github.com/The-PR-Agent/pr-agent/issues/2514
* https://github.com/The-PR-Agent/pr-agent/issues/2516

maurerle:
For now, I think that it is rather fine to have both blocks (code suggestions and reviewer guide).
In here: https://github.com/assume-framework/assume/pull/839#issuecomment-4983689396 it found important fixes.

So I would just like to fix the two open issues (not overwriting the issue desc that much and providing more information about used models/tokens)

PR #842, opened 2026-07-27
scenario: replace GPL-licensed pyyaml-include with in-house !include support

jvgkr:
## Description

`assume/scenario/loader_amiris.py` depended on `pyyaml-include` to resolve the
`!include` tags used by AMIRIS scenario YAML files. That package is licensed
under GPLv3+, which is incompatible with distributing ASSUME under a
permissive license.

This PR removes the dependency and replaces it with a small, from-scratch
implementation (`assume/scenario/yaml_include.py`) covering exactly the
subset of behavior AMIRIS scenario files actually rely on:

- `!include "path.yaml"` — load and parse a single file
- `!include ["glob/*.yaml", "Label"]` — glob-expand and return an unflattened
  list of each matched file's parsed content (the label is accepted for
  compatibility but has no effect, matching the original library's behavior)
- nested `!include`s always resolve relative to the original `base_dir`, not
  the including file's own directory

The exact semantics were reverse-engineered from `pyyaml-include` 2.2's
source and cross-checked against real AMIRIS scenario files
(gitlab.com/dlr-ve/esy/amiris/examples) before implementation, so this is a
behavioral drop-in replacement, not a reinterpretation of the format.

`pyyaml-include >=2.2a` has been removed from `pyproject.toml`.

## Checklist
- [x] Documentation updated (docstrings, READMEs, user guides, inline comments, `docs` folder updates, etc.)
- [x] Consent to release this PR's code under the GNU Affero General Public License v3.0 (and MIT license)

## Additional Notes

Added `tests/test_yaml_include.py` (unit tests directly against the new
constructor) and `tests/test_loader_amiris.py` (integration test against a
small AMIRIS-shaped fixture at `tests/fixtures/amiris_scenario/`, including a
check that `get_send_receive_msgs_per_id` still consumes the parsed output
correctly). Full existing test suite (363 tests) still passes.

This only removes the GPL *dependency*.

### DIFF
--- a/assume/scenario/yaml_include.py
+++ b/assume/scenario/yaml_include.py
@@ -72,7 +72,7 @@ class Constructor:
 
         base = Path(self.base_dir)
         loader_type = type(loader)
-        
+
         if any(char in urlpath for char in WILDCARD_CHARS):
             matches = sorted(base.glob(urlpath))
             return [self._load(match, loader_type) for match in matches]
--- a/assume/scenario/yaml_include.py
+++ b/assume/scenario/yaml_include.py
@@ -71,13 +71,13 @@ class Constructor:
         else:
             raise TypeError(f"!include does not support {type(node).__name__} nodes")
 
-        path = Path(self.base_dir) / urlpath
+        base = Path(self.base_dir)
         loader_type = type(loader)
-
+        
         if any(char in urlpath for char in WILDCARD_CHARS):
-            matches = sorted(glob.glob(str(path), recursive=True))
+            matches = sorted(base.glob(urlpath))
             return [self._load(match, loader_type) for match in matches]
-        return self._load(path, loader_type)
+        return self._load(base / urlpath, loader_type)
 
     @staticmethod
     def _load(path: Path | str, loader_type: type) -> Any:
--- a/assume/scenario/yaml_include.py
+++ b/assume/scenario/yaml_include.py
@@ -36,8 +36,8 @@ they appear in.
 from __future__ import annotations
 
 import glob
-import os
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Any
 
 import yaml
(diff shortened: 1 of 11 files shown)


assume/scenario/yaml_include.py
```diff
@@ -0,0 +1,85 @@
...
+            urlpath = loader.construct_scalar(node)
+        elif isinstance(node, yaml.SequenceNode):
+            # Additional entries are accepted for compatibility with AMIRIS
+            # scenario files (e.g. `!include ["contracts/*.yaml", "Contracts"]`)
+            # but carry no meaning here and are ignored.
+            urlpath = loader.construct_sequence(node)[0]
+        else:
+            raise TypeError(f"!include does not support {type(node).__name__} nodes")
+
+        path = os.path.join(self.base_dir, urlpath)
```
maurerle: please use pathlib instead

maurerle (approved):
Looks fine in general. 85 lines of code instead of an additional dependency - worth it.

assume/scenario/yaml_include.py
jvgkr: fixed with new commit

assume/scenario/yaml_include.py
```diff
@@ -0,0 +1,85 @@
...
+        else:
+            raise TypeError(f"!include does not support {type(node).__name__} nodes")
+
+        path = Path(self.base_dir) / urlpath
+        loader_type = type(loader)
+
+        if any(char in urlpath for char in WILDCARD_CHARS):
+            matches = sorted(glob.glob(str(path), recursive=True))
+            return [self._load(match, loader_type) for match in matches]
+        return self._load(path, loader_type)
```
maurerle: What about something like this and fully ditch glob and str path :

```suggestion
        base = Path(self.base_dir)
        loader_type = type(loader)
        
        if any(char in urlpath for char in WILDCARD_CHARS):
            matches = sorted(base.glob(urlpath))
            return [self._load(match, loader_type) for match in matches]
        return self._load(base / urlpath, loader_type)
```

maurerle:
Looks fine to me - I would squash merge this PR as it contains various iterations :)

PR #852, opened 2026-09-07
fix loss of the last ``train_freq`` window during learning

### LINKED ISSUE #851: During training the last "train_freq" window is not recorded properly
### Description

Tasks that fire at exactly `world.end` run, but any message they schedule is never
delivered — the run loop exits and the container shuts down first. For the RL learning
role this silently drops the final `train_freq` window from the `rl_params` table on
every learning run.
`rl_params` is short by exactly one `train_freq` window on every run (e.g. 93 of 743
  eligible hours for a 744 h horizon with `train_freq: 96h` → auto-adjusted to `93h`;
  24 h for the default `train_freq: 24h`).
As a result, TensorBoard curves, Grafana training dashboards and absolute episode reward totals are all short
  by one window. Relative comparisons between episodes stay matched, since every episode
  is truncated identically.

The bug results from the schedule to the last hour here [
`learning_role:142-142`](https://github.com/assume-framework/assume/blob/f349e1f35c4709216ee83bd846ff95cc377bc819/assume/reinforcement_learning/learning_role.py#L142) and the loop exiting early here: https://github.com/assume-framework/assume/blob/f349e1f35c4709216ee83bd846ff95cc377bc819/assume/world.py#L839

mthede:
## Related Issue
Closes #851 

## Description
At the final timestep during a learning episode, premature shutdown of the container led to loss of training data which was not stored in the outputs or the buffer. Now, all remaining tasks are completed before shutdown. 

## Checklist
- [x] Documentation updated (docstrings, READMEs, user guides, inline comments, `docs` folder updates, etc.)

## Additional Notes
@maurerle do you know if anything changes with newer mango-agents versions that could have introduced a new default behavior? I was really surprised to observe missing data for the final``train_freq`` period.

### DIFF
--- a/assume/world.py
+++ b/assume/world.py
@@ -848,6 +848,10 @@ class World:
                     else:
                         self.clock.set_time(end_ts)
                     prev_delta = delta
+
+                # tasks which fire at exactly end_ts need to be awaited before shutdown
+                await asyncio.sleep(0)
+                await tasks_complete_or_sleeping(c, except_sources=[])
             else:
                 # real-time mode
                 while self.clock.time < end_ts:
--- a/assume/world.py
+++ b/assume/world.py
@@ -848,10 +848,6 @@ class World:
                     else:
                         self.clock.set_time(end_ts)
                     prev_delta = delta
-
-                # tasks which fire at exactly end_ts need to be awaited before shutdown
-                await asyncio.sleep(0)
-                await tasks_complete_or_sleeping(c, except_sources=[])
             else:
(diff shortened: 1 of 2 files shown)


reinecfi (approved):
From my perspective this looks fine. 02a now runs with correct training horizon.
Thanks for the digging.

reinecfi:
@maurerle is it a bug on my end or does the PR code suggestions bot seem to be a little enthusiastic here. On my page every comment was done 3 times.

PR #853, opened 2026-09-07
Split cashflow and reward

mthede:
## Related Issue
Closes #<issue‑number>
Also fixes missing final time step of "market_dispatch" outputs.

## Description

Splits the cashflow from the reward calculation and decouples the export of the *market*
dispatch from the execution of the *actual* dispatch. This enables reward calculation based on final feasibility and supports the implementation of multi-market RL strategies. 

Until now, `write_actual_dispatch` was triggered by every market clearing and did three things at
once: aggregate the accepted orders per market, execute the units' dispatch and calculate cashflow
and reward. That coupling meant the executed dispatch (and with it the reward) depended on expectations about the future delivery, e.g. if starting-costs need to be attributed to the reward of the market. 

**Cashflow vs. reward**

- `BaseUnit.calculate_cashflow_and_reward` is split into `calculate_cashflow` and
  `calculate_reward`. The cashflow follows directly from the accepted prices and volumes, so it is
  booked immediately at clearing. The reward depends on the dispatch that is actually realized, so
  it is deferred.
- `UnitsOperator` keeps the orders of every market in `pending_orders` until their delivery period
  has ended; `calculate_unit_reward` then computes the unit and portfolio rewards on the basis of
  the dispatch actually realized across *all* markets that cleared for that delivery period.
- Because profits and costs can now receive contributions from several markets for the same time
  step, the heuristic strategies (`flexable`, `flexable_storage`) accumulate into
  `unit.outputs["profit"]` / `["total_costs"]` instead of overwriting them.

**Market dispatch vs. actual dispatch**

- `execute_dispatch` is scheduled as a recurrent task on the simulation index, so the units'
  dispatch is executed exactly once per time step, independent of any market. It writes the
  `unit_dispatch` output and triggers the reward calculation for all delivery periods that have
  passed. 
- `write_market_dispatch` stays at the clearing and is now tracked **per market** rather than per
  product type (the aggregation groups by `market_id` anyway). At the last clearing of a market it
  aggregates up to the market end, so the results of the final clearing are exported as well. It therefore fixes the missing final timestep in the `market_dispatch` output.

**Supporting changes**

- The last possible market opening moves from `MarketMechanism.last_market_opening` to a
  `MarketConfig.last_opening` property, so the units operator can use it to detect the final
  clearing.
- `create_rrule` accepts a `timedelta` in addition to a frequency string;
  `convert_to_rrule_freq` is renamed to `convert_str_to_rrule_freq` and complemented by
  `convert_timedelta_to_rrule_freq`, which picks the coarsest matching unit (`HOURLY` /
  `MINUTELY` / `SECONDLY`) so that iterating the rule stays cheap.
- `Learning.store_to_buffer_and_update` yields once before reading the caches, since the rewards
  are now written after the delivery period.

## Checklist
- [ ] Documentation updated (docstrings, READMEs, user guides, inline comments, `docs` folder updates, etc.)
- [ ] New unit/integration tests added -> but need to be revised

Several `TODO`s are added where the current cost/reward attribution is questionable now that a unit
can be active on several markets at once (generation costs as the product instead of the integral
of the marginal cost, start-up cost allocation across markets, costs based on dispatch instead of
accepted volume, missing profit outputs in the portfolio learning strategy). 

Tests still need to be reviewed and potentially added.

There are also multiple other `TODO`s that need to be investigated before converting this into a review-ready PR .

### DIFF
--- a/assume/strategies/flexable.py
+++ b/assume/strategies/flexable.py
@@ -226,7 +226,9 @@ class EnergyHeuristicFlexableStrategy(MinMaxStrategy):
 
             output = unit.outputs[product_type].at[start]
             marginal_cost = unit.calculate_marginal_cost(start, output)
-            costs[i] += marginal_cost * output
+            costs[i] += (
+                marginal_cost * output
+            )  # TODO: only base the costs on the actual market contribution (accepted volume)
 
             if output != 0 and op_time < 0:
                 start_up_cost = unit.get_starting_costs(op_time)
@@ -235,11 +237,13 @@ class EnergyHeuristicFlexableStrategy(MinMaxStrategy):
         profit -= costs
 
         # store results in unit outputs which are written to database by unit operator
-        unit.outputs["profit"].loc[products_index] = profit
-        unit.outputs["total_costs"].loc[products_index] = costs
+        unit.outputs["profit"].loc[products_index] += profit
+        unit.outputs["total_costs"].loc[products_index] += costs
 
         # update average operation time
-        update_avg_op_time(unit, product_type, products_index[0], products_index[-1])
+        update_avg_op_time(
+            unit, product_type, products_index[0], products_index[-1]
+        )  # TODO: check if this needs to be moved to the cashflow calculation?
 
 
 class CapacityHeuristicBalancingPosStrategy(MinMaxStrategy):
--- a/assume/strategies/flexable_storage.py
+++ b/assume/strategies/flexable_storage.py
@@ -203,18 +203,21 @@ class StorageEnergyHeuristicFlexableStrategy(MinMaxChargeStrategy):
             end_excl = order["end_time"] - unit.index.freq
 
             # Extract outputs and costs in one step
-            outputs = unit.outputs[product_type].loc[start:end_excl]
+            outputs = unit.outputs[product_type].loc[
+                start:end_excl
+            ]  # TODO: use accepted volume (and price?) from orderbook
             costs = np.where(
                 outputs != 0,
                 np.abs(outputs)
                 * np.array([unit.calculate_marginal_cost(start, x) for x in outputs]),
-                0,
+                0,  # TODO: check if this is correct for step-wise marginal costs, shouldn't it be the area under the curve?
             )
 
-            unit.outputs["profit"].loc[start:end_excl] = (
-                unit.outputs[f"{product_type}_cashflow"].loc[start:end_excl] - costs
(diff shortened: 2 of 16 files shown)


assume/strategies/flexable_storage.py
```diff
@@ -203,18 +203,21 @@ def calculate_reward(
...
-            outputs = unit.outputs[product_type].loc[start:end_excl]
+            outputs = unit.outputs[product_type].loc[
+                start:end_excl
+            ]  # TODO: use accepted volume (and price?) from orderbook
             costs = np.where(
                 outputs != 0,
                 np.abs(outputs)
                 * np.array([unit.calculate_marginal_cost(start, x) for x in outputs]),
-                0,
+                0,  # TODO: check if this is correct for step-wise marginal costs, shouldn't it be the area under the curve?
```
reinecfi, line 213: For the current storage implementation this should be correct, as it is not time dependent. If it was time dependent, one should probably swap `start` with the actual timestep.

assume/strategies/flexable.py
```diff
@@ -235,11 +237,13 @@ def calculate_reward(
...
-        unit.outputs["profit"].loc[products_index] = profit
-        unit.outputs["total_costs"].loc[products_index] = costs
+        unit.outputs["profit"].loc[products_index] += profit
+        unit.outputs["total_costs"].loc[products_index] += costs
 
         # update average operation time
-        update_avg_op_time(unit, product_type, products_index[0], products_index[-1])
+        update_avg_op_time(
+            unit, product_type, products_index[0], products_index[-1]
+        )  # TODO: check if this needs to be moved to the cashflow calculation?
```
reinecfi, line 246: If I understand correctly, this will be double counted both in `calculate_reward` AND in `calculate_cashflow`, as both are called once per market. 
I think the right place might be `execute_dispatch`

reinecfi:
I like the idea really much, as I was also thinking about multimarket strategies recently. Had some trouble wrapping my head around what needs to be accounted for where and when =)

As the code review bot already mentioned: there seems to be a bug with the last_executed_dispatch and some double costs and double counting problems (although they latter ones are already marked with TODOs)

assume/strategies/flexable.py
mthede: moved it to be called after `unit.execute_current_dispatch` (<-- `get_actual_dispatch` <-- `execute_dispatch`)

PR #854, opened 2026-09-08
Balancing markets query wrong forecaster

reinecfi:
## Description
Removes a bug where the balancing market strategies: 
- StorageCapacityHeuristicBalancingPosStrategy
- CapacityHeuristicBalancingPosStrategy
- CapacityHeuristicBalancingNegStrategy

try to calculate potential revenues on the EOM market, but query their own balancing market instead.

## Checklist
- [ ] Documentation updated (docstrings, READMEs, user guides, inline comments, `docs` folder updates, etc.)


maurerle (approved):
I'll look into why the pr-agent adds new comments on each added commit.

Looks good.
We might change the `reference_market` property in the future, but I think this is a good way to move away from the hard coded "EOM" values we still have laying around.

